PR #61, opened 2023-12-26
bnb consensus client implementation

dharjeezy:
This PR implements an ISMP consensus client for the BNB Blockchain network.


parachain/modules/consensus/bnb-parlia/verifier/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+        &aggregate_public_key_byte_vector,
+        bls_public_keys_as_byte_vectors.as_slice(),
+        msg.0.to_vec(),
+        signature.to_vec().as_ref(),
+    )
+    .map_err(|_| anyhow!("Could not verify aggregate signature"))?;
+
+    let hash = rlp_header.hash::<I>()?;
+
+    let validator_addresses: Option<BTreeSet<H160>> = Some(
```
Wizdave97, line 65: This should only be Some if the validators array in the extra data is non empty

Wizdave97 (changes requested):
Rlp encoding is only required for the header and VoteAttestation and VoteData

parachain/modules/consensus/bnb-parlia/verifier/src/primitives.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub extra_vanity: Vec<u8>,
+    pub validator_size: u8,
+    pub validators: Vec<CodecValidatorInfo>,
+    pub extra_seal: Vec<u8>,
+    pub agg_signature: [u8; 96],
+    pub vote_data_hash: H256,
+}
+
+#[derive(Debug, Clone)]
+pub struct CodecValidatorInfo {
```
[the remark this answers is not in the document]
dharjeezy: We'd need this since we are have to know the validator address and also the Bls public key associated with that validator address

parachain/modules/consensus/bnb-parlia/verifier/src/primitives.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub vote_address_set: u64,
+    pub agg_signature: [u8; 96],
+    pub data: VoteData,
+    pub extra: alloy_primitives::Bytes,
+    pub validator_size: Option<u8>,
+    pub validators: Option<Vec<ValidatorInfo>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct CodecExtraData {
```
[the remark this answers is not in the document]
dharjeezy: Maybe some data in the the `CodecExtraData` is not needed. But then we need the `agg_signature` and `validators` in it as that will be the next_validators to be stored in the consensus state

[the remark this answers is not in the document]
Wizdave97: Yeah we only need the aggregate signature, vote attestation and validators from the extra data.

parachain/modules/consensus/bnb-parlia/verifier/src/lib.rs
Wizdave97, line 65: validators list should be a vector not a set

parachain/modules/consensus/bnb-parlia/verifier/src/primitives.rs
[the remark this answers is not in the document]
dharjeezy: OK. Need to also fix the naming too

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    let parse_extra_data = parse_extra::<I>(&rlp_header.extra_data.0.as_ref())
+        .map_err(|_| anyhow!("could not parse extra data from header"))?;
+
+    let bls_public_keys: Vec<[u8; 48]> = validators
+        .iter()
+        .map(|validator| validator.bls_public_key)
+        .collect();
+
+    let aggregate_public_key = aggregate_public_keys(bls_public_keys.clone())?;
```
Wizdave97, line 55: You are meant to agggregate only the keys at the corresponding positions that have 1 in the validator bit set

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+        bls_public_keys_as_byte_vectors.as_slice(),
+        msg.0.to_vec(),
+        signature.to_vec().as_ref(),
+    )
+    .map_err(|_| anyhow!("Could not verify aggregate signature"))?;
+
+    let hash = rlp_header.hash::<I>()?;
+
+    let next_validator_addresses: Option<Vec<H160>> = {
+        let mut iter = parse_extra_data.validators.iter().map(|data| H160::from_slice(&data.address));
```
Wizdave97, line 80: Collect the items once and check if the vector is empty

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+        if let Some(first) = iter.next() {
```
Wizdave97, line 81:
```suggestion
        if !validators.is_empty() {
```

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+            Some(iter.collect())
```
Wizdave97, line 82:
```suggestion
            Some(validators)
```

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
Wizdave97, line 80:
```suggestion
        let validators = parse_extra_data.validators.iter().map(|data| H160::from_slice(&data.address)).collect();
```

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    };
+
+    Ok(VerificationResult { hash, header, next_validators: next_validator_addresses})
+}
+
+fn aggregate_public_keys(keys: Vec<[u8; 48]>) -> Result<Vec<u8>, anyhow::Error> {
+    let mut aggregate = pubkey_to_point(keys[0].to_vec().as_ref())
+        .map_err(|_| anyhow!("could not convert index 0 public key to point"))?;
+
+    for key in keys.iter().skip(1) {
```
Wizdave97, line 95: why are you skipping?

Wizdave97 (changes requested):

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
dharjeezy: because i already initialized the aggregate with the first public key

parachain/modules/consensus/bnb-pos/verifier/src/lib.rs
dharjeezy: So, the idea is that, the validators being passed into the method are those that have their corresponding positions in the validator bit set

parachain/modules/ismp/bnb-pos/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        let bnb_client_update = BnbClientUpdate::decode(&mut &proof[..]).map_err(|_| {
+            Error::ImplementationSpecific("Cannot decode bnb client update".to_string())
+        })?;
+
+        let mut consensus_state = ConsensusState::decode(&mut &trusted_consensus_state[..])
+            .map_err(|_| {
+                Error::ImplementationSpecific("Cannot decode trusted consensus state".to_string())
+            })?;
+
+        let current_epoch_block_number =
```
Wizdave97, line 55: I don't think this rotation block logic is correct.
You don't need to compute the epoch, 
You are meant to compare the next validator rotation block with the attested header block number, if the latter is greater you rotate the validator set then move on to verify the update

Wizdave97 (changes requested):

parachain/modules/ismp/bnb-pos/src/lib.rs
```diff
@@ -0,0 +1,115 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#[warn(unused_imports)]
+#[warn(unused_variables)]
+extern crate alloc;
+
+use core::marker::PhantomData;
+
+use alloc::{boxed::Box, collections::BTreeMap, string::ToString, vec, vec::Vec};
+use bnb_pos_verifier::{
+    primitives::{compute_epoch, BnbClientUpdate, ConsensusState, EPOCH_LENGTH},
```
Wizdave97, line 10: Define the consensus state in this crate

parachain/modules/consensus/bnb-pos/prover/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+        let target_header = self.fetch_header(target_hash).await?;
+
+        let bnb_client_update = BnbClientUpdate { source_header, target_header, attested_header };
+
+        Ok(bnb_client_update)
+    }
+
+    pub async fn fetch_finalized_state<I: Keccak256>(
+        &self,
+    ) -> Result<(CodecHeader, Vec<BlsPublicKey>, Option<NextValidators>), anyhow::Error> {
```
Wizdave97, line 89: We can't know the next validators ahead of time, there's no need for this here.

Wizdave97 (approved):

PR #66, opened 2024-01-04
Fraud proofs for polygon pos ismp client

Wizdave97:
Introduces some fraud proof conditions which would allow the Polygon POS light client to be frozen if polygon validators become byzantine.


scripts/zombienet/local-testnet.toml
```diff
@@ -35,7 +35,7 @@ args = []
 
 [[parachains]]
 id = 2000
-chain = "dev-2000"
+chain = "messier-2000"
```
seunlanlege, line 38: lets revert, the integration tests is using an older node release

PR #114, opened 2024-03-06
Relayer configuration update

Wizdave97:
Updates the relayer configuration documentation with the latest changes


docs/pages/network/relayer.mdx
```diff
@@ -174,6 +180,8 @@ etherscan_api_key = ""
 handler = ""
 # Contract address of the IsmpHost contract
 ismp_host = ""
+# Maximum block range for eth_getLogs in a single rpc call
+query_batch_size = 1000
```
seunlanlege, line 184: maybe add that if it isn't supplied, it doesn't break down the range

PR #116, opened 2024-03-08
implement reconnect client.

developeruche:
Various fixes to hyperclient

 - [x] Switch to using [niklasad1/reconnecting-jsonrpsee-ws-client](https://github.com/niklasad1/reconnecting-jsonrpsee-ws-client)
 - [x] Add integration tests for wasm
 - [x] remove loops in `stream::unfold`


modules/client/pkg/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@polytope-labs/hyperclient",
+  "name": "hyperclient",
```
seunlanlege, line 2: why?

modules/client/pkg/package.json
```diff
@@ -1,5 +1,5 @@
...
   "version": "0.1.0",
```
seunlanlege, line 3: bump version

modules/client/src/interfaces.rs
```diff
@@ -20,12 +20,20 @@ pub struct JsChainConfig {
     pub hash_algo: i32,
```
seunlanlege, line 20: please remove

seunlanlege (approved):

modules/client/pkg/package.json
seunlanlege: please bump version

Wizdave97:
There's a loop in the `timeout_stream` function also, think we should remove it.

Wizdave97 (approved):

PR #117, opened 2024-03-08
Add retry middleware to sync committee prover

Wizdave97:
Enable retries when fetching sync committee data


modules/consensus/sync-committee/prover/src/middleware.rs
```diff
@@ -0,0 +1,62 @@
...
+        };
+        match result {
+            Ok(response) => {
+                if response.status() == StatusCode::OK {
+                    return Ok(Some(response));
+                };
+                let next_index = next_state()?;
+                _state.active_url_index = next_index;
+                let url = self.providers[next_index].clone();
+                request.headers_mut().insert(HOST, url.parse().map_err(|e| anyhow!("{e:?}"))?);
```
seunlanlege, line 48: would be good to get tests for this, unsure if this is correct. I think you want to modify `request.url`

Wizdave97: I'll add tests some tests.

Wizdave97:
Should we integrate in this middleware instead of the regular retry?

modules/consensus/sync-committee/prover/src/test.rs
```diff
@@ -263,6 +267,28 @@ async fn test_prover() {
     }
 }
 
+#[tokio::test]
```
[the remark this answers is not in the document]
Wizdave97: Yeah

Wizdave97:
> > Should we integrate in this middleware instead of the regular retry?
> 
> Yeah i don't think we need `reqwest-retry` we just need the provider switching middleware. Although now that i think about it, why not both? I'm assuming if all urls fails, `reqwest-retry` would kick in?

Yeah but it would only retry on the last provider url.

Wizdave97:
Or we use retry first, then switch as the second strategy.

seunlanlege:
> Or we use retry first, then switch as the second strategy.

I think instead, we cycle back to the first url?

Wizdave97:
> I think instead, we cycle back to the first url?

We can only retry a maximum of the number of urls + the number of retries configured in the `reqwest-retry` middleware.
we can't switch the urls used by the `reqwest-retry` middleware.

 Do you mean we should just modify the request to use the first url before exiting the provider switcher?

seunlanlege (approved):

PR #129, opened 2024-03-20
Fix check for minimum number of sync-committee participants in verifier

vmarkushin:
According to the [code](https://github.com/polytope-labs/hyperbridge/blob/bdbc115bce984afae0800b6dda82d93680180822/modules/consensus/sync-committee/prover/src/lib.rs#L269) in the prover in prover, the number of sync-committee participants should be 2/3 + 1 (not just 2/3)


seunlanlege (approved):
Good catch, thanks 🫡

seunlanlege:
You'll need to rebase with master so CI workflows can run on your branch

PR #132, opened 2024-03-22
Make ismp-parachain accessible

Szegoo:
I was trying to test reading from other parachains locally when I realized that I couldn't access the `ismp-parachain` crate. I would assume it was left out by accident in the `Cargo.toml` file.

Also, there is another unrelated modification in this PR. It was for some reason failing to compile since it couldn't find the `parachain-system` package, even though in the `Cargo.toml` it was explicitly noted that it should look for `cumulus-pallet-parachain-system`.


seunlanlege (approved):
LGTM

Szegoo:
Yeah, will need that as well. While at it, I also added the runtime-api as well

Szegoo:
Weird :/ Some of the CI checks managed to run. I formatted the code, not sure if that could've been causing issues.

seunlanlege:
> Weird :/ Some of the CI checks managed to run. I formatted the code, not sure if that could've been causing issues.

Has to do with GitHub secrets sharing with forks. I've found a few solutions. Will try them out tomorrow

modules/ismp/clients/parachain/inherent/src/lib.rs
```diff
@@ -68,6 +68,7 @@ impl ConsensusInherentProvider {
         let message = ConsensusMessage {
             consensus_state_id: ismp_parachain::consensus::PARACHAIN_CONSENSUS_ID,
             consensus_proof: consensus_proof.encode(),
+            signer: Default::default(),
```
Szegoo, line 71: I would assume there is no signer since this is coming from the relay chain.

seunlanlege: Yes this is correct

Szegoo: That could make sense. However, I am not sure from where I could get the block author's public key. From the digest?

seunlanlege: Yeah I'd like to believe that there's a runtime or keystore API for this.

seunlanlege:
I've merged #135 which should allow CI run on contributor PRs. Not sure if you need to rebase, but if everything fails try to rebase

PR #133, opened 2024-03-22
Optimizations to evm proof sizes

Wizdave97:
Expect a single storage trie proof for all membership and non membership proofs for a contract.

In the `EvmStateProof` struct the `storage_proof` field is now a map of contract address to MPT proof.


Wizdave97:
Tested changes with relayer on local devnet.

modules/ismp/clients/sync-committee/src/utils.rs
```diff
@@ -210,17 +210,12 @@ pub(super) fn derive_array_item_key<H: IsmpHost>(slot: u64, index: u64, offset:
 pub(super) fn get_values_from_proof<H: IsmpHost + Send + Sync>(
     keys: Vec<Vec<u8>>,
     root: H256,
-    mut proof: BTreeMap<Vec<u8>, Vec<Vec<u8>>>,
+    proof: Vec<Vec<u8>>,
 ) -> Result<Vec<Option<DBValue>>, Error> {
     let mut values = vec![];
+    let proof_db = StorageProof::new(proof).into_memory_db::<KeccakHasher<H>>();
+    let trie = TrieDBBuilder::<EIP1186Layout<KeccakHasher<H>>>::new(&proof_db, &root).build();
```
seunlanlege, line 217: Why not return the key, value as a tuple here?

Wizdave97: The calling function can't really use the keys passed to this function to map to the values, these are the hashed versions of the actual keys.

seunlanlege (approved):

PR #135, opened 2024-03-25
Allow CI on contributor PRs

seunlanlege:
Try to share repo secrets safely


seunlanlege:
Going to sudo merge here because `pull_request_target` only runs from the context of the target branch. Which in this case would be the main branch. But the workflow doesn't yet exist on the main branch.

PR #141, opened 2024-03-30
Fix zero value bug in accumulate fees

Wizdave97:
Skip none values when accumulating fees


seunlanlege (approved):
Yeah looks good, the size of these proofs makes me think we'll need a call compression pallet to save on POV costs

PR #142, opened 2024-03-31
Fix try runtime

Szegoo:
Building with `features try-runtime` was failing since the ismp pallets did not have the try-runtime feature.


seunlanlege (approved):
Interesting, what's `try-runtime` useful for?

Szegoo:
> Interesting, what's `try-runtime` useful for?

https://paritytech.github.io/try-runtime-cli/try_runtime/

seunlanlege:
> https://paritytech.github.io/try-runtime-cli/try_runtime/

I see, we just use [`polytope-labs/sc-simnode`](https://github.com/polytope-labs/sc-simnode) instead.

Szegoo:
> I see, we just use [`polytope-labs/sc-simnode`](https://github.com/polytope-labs/sc-simnode) instead.

Will look into it. I wanted to use `try-runtime` in the RegionX parachain for testing runtime upgrades, but this could also be an interesting alternative.

PR #145, opened 2024-04-01
pallet call decompressor

dharjeezy:
Introduces a new Pallet which decompresses Runtime calls that are unsigned and also executes them.


Wizdave97 (approved):

Wizdave97:
Using wasm time in release mode to decompresses 209kb to 1mb:

image

Wizdave97:
Comparing to release mode in native which uses approximately 3ms, wasm has a relatively good performance
image

PR #151, opened 2024-04-03
Handle asset with call and teleport asset with call

developeruche:
1. Introduced function to teleport tokens on gateway and attach calldata to be made to the destination contract.
2. Handle Asset with call from ISMP host via `onAccept` function.


evm/src/modules/CallDispatcher.sol
```diff
@@ -0,0 +1,29 @@
...
+import {ICallDispatcher} from "../interfaces/ICallDispatcher.sol";
+
+
+/// @notice This contract is used to dispatch calls to other contracts.
+contract CallDispatcher is ICallDispatcher {
+    /// @dev funtion returns `success = false` if the target is not a contract and reverts if the call to the target contract fails.
+    function dispatch(address target, bytes calldata data) external returns (bytes memory result, bool success) {
+        if (isContract(target)) {
+            return _dispatch(target, data);
+        }
```
seunlanlege, line 13: I think we should consolidate all this stuff into this `dispatch` call. They're not re-used anywhere

evm/src/modules/TokenGateway.sol
```diff
@@ -30,6 +31,43 @@ struct TeleportParams {
...
+    // recipient address
+    address to;
+    // The Erc20 token to be used to swap for a fee
+    address feeToken;
+    // recipient state machine
+    bytes dest;
+    // timeout in seconds
+    uint64 timeout;
+    // destination contract call data
+    bytes data;
```
seunlanlege, line 52: rather than needing a new struct, just add it to the existing teleport params. If the `calldata.length > 0` then use `BodyWithCall`

evm/test/TokenGatewayTest.sol
```diff
@@ -35,6 +36,35 @@ contract TokenGatewayTest is BaseTest {
...
+                timeout: 0,
+                to: address(miniStaking),
+                assetId: keccak256("USD.h"),
+                data: stakeCalldata
+            })
+        );
+
+        assert(feeToken.balanceOf(address(this)) == 0);
+
+        assert(feeToken.balanceOf(address(host)) == messagingFee);
```
seunlanlege, line 65: Add an assertion for the call to ministaking

seunlanlege: ah yes you're right

evm/src/modules/TokenGateway.sol
```diff
@@ -161,33 +200,86 @@ contract TokenGateway is BaseIsmpModule {
...
+                erc6160,
+                params.dest,
+                data,
+                params.timeout,
+                params.to
+            )
+        );
+    }
+
+    function teleportWithCall(TeleportParams memory params) public {
```
seunlanlege, line 225: Let's remove

seunlanlege (approved):

PR #155, opened 2024-04-05
Relay chain native asset XCM Support

Wizdave97:
Allow receiving the native asset from the relay chain through xcm and dispatch an ismp request to the destination chain.
Multilocation variant accepted for beneficiary is of the form described below

The nework id for the `AccountKey20` should be the chain id for a destination evm network supported by hyperbridge.
The value in the GeneralIndex is going to be used as the request timeout for the ismp post request that would be dispatched.
```
   MultiLocation {
        parents: 0,
        interior:
              Junctions::X3(
                   Junction::AccountId32 { id , .. },
                   Junction::AccountKey20 { network: Some(network), key },
                   GeneralIndex(timeout)
               ),
     }
```


modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+        /// The asset tranfer's pallet id, used for deriving its sovereign account ID.
+        /// All escrowed funds will be custodied by this account
+        #[pallet::constant]
+        type PalletId: Get<PalletId>;
+
+        /// Protocol fees will be custodied by this account
+        #[pallet::constant]
+        type ProtocolAccount: Get<PalletId>;
```
seunlanlege, line 71: why not just use the `PalletId` for this?

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+        /// The 32 bytes Asset Id used to identify the DOT token on Token Gateway deployments
+        #[pallet::constant]
+        type DotAssetId: Get<H256>;
+
+        /// Percentage to be taken as protocol fees
+        #[pallet::constant]
+        type ProtocolFees: Get<Percent>;
+
+        /// Evm account id type
+        type EvmAccountId: Parameter;
```
seunlanlege, line 86: Whats this for?

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+    ) -> Result<(), Error<T>> {
+        let dispatcher = Dispatcher::<T>::default();
+
+        let mut to = [0u8; 32];
+        let _temp: [u8; 20] = multi_account.evm_account.clone().into();
+        to.iter_mut().enumerate().for_each(|(index, item)| {
+            if index < 20 {
+                *item = _temp[index];
+            }
+        });
```
seunlanlege, line 154: i think you can just use `to[..20].copy_from_slice(&multi_account.evm_account)`

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+            from: from.into(),
+            to: to.into(),
+        };
+
+        let dispatch_post = DispatchPost {
+            dest: multi_account.dest_state_machine,
+            from: T::TokenGateWay::get().0.to_vec(),
+            to: T::TokenGateWay::get().0.to_vec(),
+            // 1 hour timeout
+            timeout_timestamp: 60 * 60,
```
seunlanlege, line 175: Is there any way to make this configurable from the xcm call?

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
Wizdave97: Wanted to have a separate account that custodies protocol fees, just for ease of tracking

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+    u128: From<<T::Assets as fungibles::Inspect<T::AccountId>>::Balance>,
+    <T::Assets as fungibles::Inspect<T::AccountId>>::AssetId: From<MultiLocation>,
+    T::AccountId: Into<[u8; 32]> + From<[u8; 32]>,
+    T::EvmAccountId: Into<[u8; 20]>,
+{
+    fn on_accept(&self, post: ismp::router::Post) -> Result<(), ismp::error::Error> {
+        let request = Request::Post(post.clone());
+        // Check that destination module is equal to the known module id
+        ensure!(
+            request.destination_module() == T::TokenGateWay::get().0.to_vec(),
```
seunlanlege, line 233: meh, this check isn't neccessary i believe the router already does this, i think we instead want to check that the request originates from the `TokenGateway` contract address

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    fn on_timeout(&self, request: Timeout) -> Result<(), ismp::error::Error> {
+        // We don't custody user funds, we send the dot back to the relaychain using xcm
+
+        match request {
+            Timeout::Request(Request::Post(post)) => {
+                let request = Request::Post(post.clone());
+                ensure!(
+                    request.source_module() == T::TokenGateWay::get().0.to_vec(),
```
seunlanlege, line 329: Router should already perform this check no?

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+                        dest: request.dest_chain(),
+                        nonce: request.nonce(),
+                    },
+                })?;
+
+                Pallet::<T>::deposit_event(Event::<T>::AssetReceived {
+                    beneficiary,
+                    amount: amount.into(),
+                    source: request.dest_chain(),
+                });
```
seunlanlege, line 400: Maybe should be `AssetRefunded` ?

modules/ismp/pallet/pallet-asset-transfer/src/xcm_utilities.rs
```diff
@@ -0,0 +1,248 @@
...
+const ARBITRUM_SEPOLIA_CHAIN_ID: u64 = 421614;
+const OPTIMISM_SEPOLIA_CHAIN_ID: u64 = 11155420;
+const BASE_SEPOLIA_CHAIN_ID: u64 = 84532;
+const SEPOLIA_CHAIN_ID: u64 = 11155111;
+const BSC_TESTNET_CHAIN_ID: u64 = 97;
+
+pub struct WrappedNetworkId(pub NetworkId);
+
+impl WrappedNetworkId {
+    pub fn transform_to_state_machine(self) -> Option<StateMachine> {
```
seunlanlege, line 35: why not a `TryFrom`?

modules/ismp/pallet/pallet-asset-transfer/src/xcm_utilities.rs
```diff
@@ -0,0 +1,248 @@
...
+pub struct MultiAccount<A, B> {
+    /// Origin substrate account
+    pub substrate_account: A,
+    /// Destination evm account
+    pub evm_account: B,
+    /// Destination state machine
+    pub dest_state_machine: StateMachine,
+}
+
+impl<A: From<[u8; 32]> + Into<[u8; 32]> + Clone, B: From<[u8; 20]> + Into<[u8; 20]> + Clone>
```
seunlanlege, line 68: nit: prefer a where bound instead of this.

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
Wizdave97: Identifier for ethereum accounts, not really necessary in retrospect

modules/ismp/pallet/testsuite/src/xcm.rs
seunlanlege, line 1: Why do we need this?

parachain/runtimes/gargantua/src/ismp.rs
```diff
@@ -86,6 +95,58 @@ impl pallet_call_decompressor::Config for Runtime {
     type MaxCallSize = ConstU32<2>;
 }
 
+// todo: set corrrect parameters
+parameter_types! {
+    pub const AssetPalletId: PalletId = PalletId(*b"asset-tx");
+    pub const ProtocolAccount: PalletId = PalletId(*b"protocol");
+    pub const TokenGateWay: H160 = H160::zero();
+    pub const DotAssetId: H256 = H256::zero();
+    pub const ProtocolFees: Percent = Percent::from_percent(1);
```
seunlanlege, line 104: So actually i think these should be configurable at runtime and adjustable by root

parachain/runtimes/gargantua/src/xcm.rs
```diff
@@ -213,7 +212,7 @@ impl staging_xcm_executor::Config for XcmConfig {
     type MessageExporter = ();
     type UniversalAliases = Nothing;
     type CallDispatcher = RuntimeCall;
-    type SafeCallFilter = Everything;
+    type SafeCallFilter = Nothing;
```
seunlanlege, line 215: hmmm, so people can dispatch any xcm call to our chain? Would be good to know the associated risks. Otherwise would prefer to limit this.

modules/ismp/pallet/testsuite/src/xcm.rs
Wizdave97: Simulating the xcm flow from the relay chain to the parachain

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
seunlanlege: > Wanted to have a separate account that custodies protocol fees, just for ease of tracking

But the `PalletId` will not be used for any other fees?

parachain/runtimes/gargantua/src/xcm.rs
Wizdave97: We have to add a filter for the calls we want to support, but we there's none at the moment that's why I put nothing, to block all calls

modules/ismp/pallet/testsuite/src/xcm.rs
seunlanlege: > Simulating the xcm flow from the relay chain to the parachain

I'm woried that this maybe be misrepresentative of real world usage, why are we using a mock instead of the pallet directly

parachain/runtimes/gargantua/src/ismp.rs
Wizdave97: Yeah, was thinking of putting them in storage.

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
Wizdave97: Yeah, I can probably use the generalIndex variant to accept the timeout

parachain/runtimes/gargantua/src/xcm.rs
seunlanlege: ok so `Nothing` means block all calls? I thought it meant filter `Nothing` out. sigh, these naming schemes

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
Wizdave97: If we use the same account, it might be more tricky to seperate  protocol fees from TVL if we change the protocol fee percentage a couple times.

modules/ismp/pallet/testsuite/src/xcm.rs
Wizdave97: I've added the XcmpQueue pallet, but we still have to manually implement the trait that executes the messages coming from the relay chain in the unit tests, the `xcm-simulator` still relies on `DmpHandler` which is deprecated.

Wizdave97:
Working on an integration test.

modules/ismp/pallet/testsuite/src/xcm.rs
seunlanlege: Yeah i think this is much better

seunlanlege (approved):

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+            from: from.into(),
+            to: to.into(),
+        };
+
+        let dispatch_post = DispatchPost {
+            dest: multi_account.dest_state_machine,
+            from: Self::token_gateway_address().0.to_vec(),
+            to: Self::token_gateway_address().0.to_vec(),
+            timeout_timestamp: multi_account.timeout,
+            data: alloy_rlp::encode(body),
```
seunlanlege, line 246: This should be prefixed with a zero-byte for the handle incoming asset variant.

https://github.com/polytope-labs/hyperbridge/blob/main/evm/src/modules/TokenGateway.sol#L226

modules/ismp/pallet/pallet-asset-transfer/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+}
+
+impl<T: Config> IsmpModule for Module<T>
+where
+    <T::Assets as fungibles::Inspect<T::AccountId>>::Balance: From<u128>,
+    u128: From<<T::Assets as fungibles::Inspect<T::AccountId>>::Balance>,
+    <T::Assets as fungibles::Inspect<T::AccountId>>::AssetId: From<MultiLocation>,
+    T::AccountId: Into<[u8; 32]> + From<[u8; 32]>,
+{
+    fn on_accept(&self, post: ismp::router::Post) -> Result<(), ismp::error::Error> {
```
seunlanlege, line 298: We should also charge protocol fees here

Wizdave97: I thought gateway will do that when initializing the transaction on the evm that's why I removed the charge there.

seunlanlege: > I thought gateway will do that when initializing the transaction on the evm that's why I removed the charge there.

It doesn't charge fees for `ERC6160` assets. It only charges when redeeming `ERC20` assets

PR #161, opened 2024-04-09
Chore: SDK implemented

Adebara123:


Wizdave97:
Why do we have all those type definitions not directly connected to the teleport function in gateway.

Adebara123:
> Why do we have all those type definitions not related to GateWay, are they all needed to make it work?

I have removed all the types not needed and cleaned up the code.

PR #162, opened 2024-04-09
Simplify fee model

### LINKED ISSUE #106: Support Account Abstraction(?)
In the account abstraction paradigm, user accounts will no longer be `tx.origin`. This is as a result of moving their assets from an EOA to a smart contract account. Consequently, the `EvmHost` should be reworked to support smart contract accounts which hold assets in `msg.sender` instead of `tx.origin`.

Some ideas

- [x] Don't charge `tx.origin` by default when dispatching requests, instead allow the application to choose who should pay for fees.
- [x] Allow relayers use account abstraction.

casweeney:
Rather than making the user approve the host infitely in a way that can be exploited by malicious applications. We instead simply charge the application the neccessary fees, the application in turn should charge it's users. With EIP-3074 confirmed we now allow relayers also be managed by a smart contract account

Closes #106


evm/src/hosts/EvmHost.sol
```diff
@@ -13,6 +13,8 @@ import {StateCommitment, StateMachineHeight} from "ismp/IConsensusClient.sol";
 import {IHandler} from "ismp/IHandler.sol";
 import {PostRequest, PostResponse, GetRequest, GetResponse, PostTimeout, Message} from "ismp/Message.sol";
 
+import {IAllowanceTransfer} from "../interfaces/IAllowanceTransfer.sol";
```
seunlanlege, line 16: prefer we import it directly from a library, or submodule

evm/src/hosts/EvmHost.sol
```diff
@@ -42,6 +44,8 @@ struct HostParams {
     uint256 lastUpdated;
     // latest state machine height
     uint256 latestStateMachineHeight;
+    // Contract Instance of Permit2 contract
+    IAllowanceTransfer permit2;
```
seunlanlege, line 48: Prefer we use address

```suggestion
    address permit2;
```

evm/src/hosts/EvmHost.sol
```diff
@@ -712,7 +721,8 @@ abstract contract EvmHost is IIsmpHost, IHostManager, Context {
 
         require(metadata.sender != address(0), "Unknown request");
         require(metadata.sender != _msgSender(), "User can only fund own requests");
-        require(IERC20(feeToken()).transferFrom(_msgSender(), address(this), amount), "Payer has insufficient funds");
+
+        _hostParams.permit2.transferFrom(_msgSender(), address(this), uint160(amount), feeToken()); // Payer has insufficient funds
```
seunlanlege, line 725: I think for this we should use `permitTransferFrom`, otherwise the user will have to make 2 txs, one to approve and one to fund their request.

https://docs.uniswap.org/contracts/permit2/reference/signature-transfer#single-permittransferfrom

evm/src/interfaces/IAllowanceTransfer.sol
seunlanlege, line 1: Yeah lets just bring in the full library

evm/src/modules/TokenGateway.sol
```diff
@@ -35,6 +43,13 @@ struct TeleportParams {
     // calculated amountInMax: 
     // used if selected fee token is not expected fee token
     uint256 amountInMax;
+    // Permit Details & Signature for host to spend feeToken
+    TeleportPermit hostFeeTokenPermit;
+    // Permit Details & Signature for gateway to spend asset
+    TeleportPermit gatewayAssetPermit;
+    // Permit Details & Signature for gateway to user selected fee token
+    // This is useful when the selected fee token is not the required fee token
+    TeleportPermit gatewayInputFeeTokenPermit;
```
seunlanlege, line 52: uhm i think if this is the case, might be better to just do a batch permit?

https://docs.uniswap.org/contracts/permit2/reference/allowance-transfer#batched-permit

evm/src/hosts/EvmHost.sol
```diff
@@ -42,6 +45,8 @@ struct HostParams {
     uint256 lastUpdated;
     // latest state machine height
     uint256 latestStateMachineHeight;
+    // Permit2 contract address
+    address permit2Address;
```
seunlanlege, line 49:
```suggestion
    address permit2;
```

evm/src/modules/TokenGateway.sol
```diff
@@ -205,13 +222,16 @@ contract TokenGateway is BaseIsmpModule {
             );
 
         if (erc20 != address(0) && !params.redeem) {
-            require(IERC20(erc20).transferFrom(from, address(this), params.amount), "Insufficient user balance");
+
+            IAllowanceTransfer(permit2Address).permit(from, params.gatewayAssetPermit.permitSingle, params.gatewayAssetPermit.signature);
+            IAllowanceTransfer(permit2Address).transferFrom(from, address(this), uint160(params.amount), erc20);
```
seunlanlege, line 227: lets not use permit for gateway

evm/src/modules/TokenGateway.sol
```diff
@@ -169,92 +157,112 @@ contract TokenGateway is BaseIsmpModule {
...
-            dest: params.dest,
+            dest: teleportParams.dest,
             to: abi.encodePacked(address(this)),
-            body: bytes.concat(hex"00", data), // add enum variant for body
-            timeout: params.timeout,
-            fee: params.fee,
+            body: data,
+            timeout: teleportParams.timeout,
+            fee: teleportParams.fee,
             payer: msg.sender
```
developeruche, line 266: Was thinking the payer for this dispatch call to the host would be the application in this case `Gateway`, [address(this)]

seunlanlege: The payer is the account to be refunded in the case of a timeout

seunlanlege (approved):

PR #181, opened 2024-04-21
initial setup

MrishoLukamba:
resolves#72

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,22 @@
+{
+    "workbench.colorCustomizations": {
+        "activityBar.activeBackground": "#3399ff",
+        "activityBar.background": "#3399ff",
+        "activityBar.foreground": "#15202b",
+        "activityBar.inactiveForeground": "#15202b99",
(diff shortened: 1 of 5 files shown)


Wizdave97:
This would use the naive Beefy Consensus verifier, no zk involved here, it's running on substrate chains so there's sufficient compute capacity.

MrishoLukamba:
> This would use the naive Beefy Consensus verifier, no zk involved here, it's running on substrate chains so there's sufficient compute capacity.

Okey so the interactive beefy verifier. 
Thanks for clarification

Wizdave97:
> > This would use the naive Beefy Consensus verifier, no zk involved here, it's running on substrate chains so there's sufficient compute capacity.
> 
> Okey so the interactive beefy verifier. Thanks for clarification

Not interactive, just a rust version of what we have for solidity https://github.com/polytope-labs/hyperbridge/blob/main/evm/src/consensus/BeefyV1.sol

MrishoLukamba:
Okey

PR #186, opened 2024-04-26
Pallet ISMP Doc

dharjeezy:
Introduce Doc to pallet ismp
fix `cargo doc` issues


Wizdave97:
Add docs for the Call Compressor Pallet and how it can be used with pallet-ismp

PR #188, opened 2024-04-30
Use child trie root as unique offchain identifier.

Wizdave97:
Using the parent block hash as a unique idntifier for offchain keys is not sufficient for fork awareness because 1 block deep reorgs could still overwrite canonical keys.

This PR allows us to use the child trie root of `pallet-ismp` as the unique indentifier for offchain keys, this works because the child trie root is unique for any sequence of requests or responses dispatched in a block, if the sequences of requests or responses are identical in different forks we are not worried about the forks overwriting each other because both forks will yield the same result.


seunlanlege (approved):

Wizdave97:
Hyperclient tests have passed multiple times, for some reason, it keeps restarting after every run. cancelling and merging.

PR #191, opened 2024-05-03
Prevent pruning potentially unfinalized mmr nodes

Wizdave97:
We could have fork blocks that have the same child trie root if they were built from the same sequence of transactions. 
In some cases the stale branch is pruned before the finalized branch which causes the nodes to be deleted before they can be moved to canonical storage.

This PR prevents pruning a node if we cannot find it in the canonical storage yet.

An example from live tests :
Nodes with `0x70a62bc49d50f915d6cb98f48f25119859de95fe6b03ff78998449ff8af9a2` child trie root were pruned earlier:  
image

The same nodes with the same child trie root from a different block could not be canonicalized later on:
image


seunlanlege (approved):
Very subtle bug

PR #201, opened 2024-05-07
Update events.rs: Addition of event metadata to PostRequestTimeout

ShunkyBoskilitu:
This pull request addresses the need to add event metadata to the PostRequestTimeout event, as requested. The changes include the introduction of a new struct TimeoutMetadata to hold additional metadata specific to the timeout event. Additionally, a new variant PostRequestTimeoutHandled is added to the Event enum to represent the event when a post request timeout is handled. This variant includes the commitment field to identify the request or response and the metadata field to hold the timeout-specific metadata.


seunlanlege:
Superceded by #202

PR #209, opened 2024-05-11
Removed the folder "target" from .dockerignore

yunusemrecatalcam:
local docker build for tesseract failing as we got the "target" folder in .dockerignore while the executable placed in there

Ref from tesseract.Dockerfile
`COPY ./target/release/tesseract ./`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-target
 .github
 .cargo
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 target
 !target/release/tesseract
 !target/release/telemetry-server
+!target/release/hyperbridge
 
 .github
 .cargo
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -1,4 +1,3 @@
-target
```
seunlanlege, line 1: I'm not sure if we should remove this though, otherwise, we'd have to copy the target folder into the image. This slows down the build process because this folder would be 10s of GBs

yunusemrecatalcam: Gotcha, didn't know that. Patched it in order to exclude the target folder excluding the executable files for tesseract and telemetry if that helps

seunlanlege: Nice, also add one for the hyperbridge binary and should be good to merge

seunlanlege (approved):

PR #242, opened 2024-06-24
[TokenGateway] Decouple relayer from liquidity providers

seunlanlege:
Previously in the `TokenGateway` protocol, relayers had to double as liquidity providers when users bridge ERC20 assets. This PR removes this requirement and instead introduces a new party called "fillers" who bid to fulfil the incoming ERC20 assets. Through this bidding mechanism, we effectively create a race to the bottom for liquidity fees, which should make `TokenGateway` cost-competitive with existing token bridges.


evm/src/modules/TokenGateway.sol
```diff
@@ -199,9 +204,33 @@ contract TokenGateway is BaseIsmpModule {
         address indexed beneficiary, uint256 amount, bytes32 indexed assetId, bytes dest, uint256 indexed nonce
     );
 
+    // A filler has just placed a bid to fulfil some request
+    event BidPlaced(bytes32 commitment, bytes32 indexed assetId, uint256 bid, address indexed bidder);
+
+    // A filler has just placed a bid to fulfil some request
```
seunlanlege, line 210:
```suggestion
    // A filler's bid has now been refuneded
```

Wizdave97 (approved):

PR #258, opened 2024-07-15
UX Improvements

seunlanlege:
This PR makes a lot of UX improvements, including changes to the EvmHost events, making them human-readable. It also modifies the token gateway contract so that it can receive assets from its instances on other chains that might have a different contract address, either due to evm incompatibility or upgraded versions of the token gateway.


modules/subxt/utils/src/lib.rs
```diff
@@ -149,16 +149,16 @@ mod gargantua_conversion {
...
+		fn from(post: ismp::router::PostRequest) -> Self {
 			Self {
 				source: post.source.into(),
 				dest: post.dest.into(),
 				nonce: post.nonce,
 				from: post.from,
 				to: post.to,
 				timeout_timestamp: post.timeout_timestamp,
-				data: post.data,
+				data: post.body,
```
Wizdave97, line 161: reminder to regenerate runtime types before we merge, so we can fix this conversion.

Wizdave97 (approved):

PR #251, opened 2024-06-29
parachain ismp relayer messaging integration tests

MrishoLukamba:
tesseract messaging relayer integration tests between 2 parachains, Testing ismp get and post requests.

PR #265, opened 2024-07-23
tesseract messaging integration test fix

MrishoLukamba:
refer to PR #251

### DIFF
--- a/tesseract/integration-test/src/lib.rs
+++ b/tesseract/integration-test/src/lib.rs
@@ -5,13 +5,12 @@ use anyhow::anyhow;
 use futures::StreamExt;
 use ismp::{
 	consensus::StateMachineHeight,
-	events::{Event, StateMachineUpdated},
 	host::StateMachine,
 	messaging::{Message, Proof, ResponseMessage},
 	router::{Request, RequestResponse},
 };
-
 use pallet_hyperbridge::VersionedHostParams;
+use pallet_ismp_demo as IsmpPalletDemo;
 use pallet_ismp_host_executive::HostParam;
 use sc_service::TaskManager;
 use std::{
@@ -21,10 +20,8 @@ use std::{
 use substrate_state_machine::{HashAlgorithm, StateMachineProof, SubstrateStateProof};
 use subxt::{
 	ext::{
-		codec::{Decode, Encode},
-		sp_core::{sr25519::Pair, Pair as PairT},
+		codec::{Decode, Encode}
 	},
-	tx::PairSigner,
 	utils::AccountId32,
 };
 use subxt_signer::sr25519::dev::{self};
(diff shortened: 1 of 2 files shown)


seunlanlege:
Get the request from the events in this call,

https://github.com/polytope-labs/hyperbridge/blob/02c2fab657dbf928ff07ea20f02297f971e0b637/tesseract/integration-test/src/lib.rs#L407

and pass it to this function

https://github.com/polytope-labs/hyperbridge/blob/02c2fab657dbf928ff07ea20f02297f971e0b637/tesseract/integration-test/src/lib.rs#L51-L56

so we can avoid this unnecessary search, which is the cause of the flakiness

https://github.com/polytope-labs/hyperbridge/blob/02c2fab657dbf928ff07ea20f02297f971e0b637/tesseract/integration-test/src/lib.rs#L67-L94

MrishoLukamba:
> Get the request from the events in this call,
> 
> https://github.com/polytope-labs/hyperbridge/blob/02c2fab657dbf928ff07ea20f02297f971e0b637/tesseract/integration-test/src/lib.rs#L407
> 
> and pass it to this function
> 
> https://github.com/polytope-labs/hyperbridge/blob/02c2fab657dbf928ff07ea20f02297f971e0b637/tesseract/integration-test/src/lib.rs#L51-L56
> 
> so we can avoid this unnecessary search, which is the cause of the flakiness
> 
> https://github.com/polytope-labs/hyperbridge/blob/02c2fab657dbf928ff07ea20f02297f971e0b637/tesseract/integration-test/src/lib.rs#L67-L94

The issue of getting the event from this function is I have to wait for tx to be executed ( wait for success) The issue for that is while tx is in block it sometimes yield invalid tx but at the same time the tx is valid and is executed ( hence leading to flakiness), Consider subxt docs belowScreenshot 2024-07-22 at 15 25 13

Wizdave97:
I suggest you look into why `wait_for_success` is flaky, there's also the option to use `wait_for_in_block`.
 It's a dev environment, transaction submission should be deterministic.

seunlanlege:
> I will try to construct a Getrequest type and pass it, the issue is it is of different type ( subxt codegen type) and this yield error. But pushing commit soon

You can write a `From` implementation

MrishoLukamba:
> I suggest you look into why `wait_for_success` is flaky, there's also the option to use `wait_for_finalized`. It's a dev environment, transaction submission should be deterministic.

should be deterministic yes, the issue is the tx do not fail but when you wait for finalization or sucess it returns tx status and among one status is Invalid, which when returned ( as it is probabilistic) based on their docs. the stream returns Err and breaks the test, meanwhile the tx is actually executed perfectly, So thats why i avoid using waiting for anything

tesseract/integration-test/src/lib.rs
```diff
@@ -126,33 +86,13 @@ async fn relay_get_response_message(
 	};
 	// =================== send to the source chain ================================
 	let _res = chain_a_client.submit(vec![response]).await?;
-	//==================== after approx 7-9 blocks the response event is emitted ===
 	// =================== fetch the returned value ================================
+	let encoded_event = chain_a_sub_client
```
Wizdave97, line 90: This can't be a reliable way to get the events; the event could have been emitted before this function is executed; instead of using events, you can also modify the demo pallet to store the result of the get response so you can just read that and use it to compare.

tesseract/integration-test/src/lib.rs
```diff
@@ -476,6 +396,8 @@ async fn run_integration_tests() -> Result<(), anyhow::Error> {
 		set_host_params(chain_b_sub_client.clone()).await?;
 	}
 
+	tokio::time::sleep(Duration::from_secs(6)).await;
```
Wizdave97, line 399: Use subscribe block headers to know when the parachain has started block production instead of sleeping.

You should do that before calling `set_host_params`

tesseract/integration-test/src/lib.rs
```diff
@@ -392,57 +318,51 @@ async fn get_request_works() -> Result<(), anyhow::Error> {
...
-		Pair::from_string("//Dave", None).expect("Unable to create Dave account"),
-	);
 	// parachain info pallet fetching para id
 	let encoded_chain_b_id_storage_key =
 		"0x0d715f2646c8f85767b5d2764bb2782604a74d81251e398fd8a0a4d55023bb3f";
+	let key = hex::decode(encoded_chain_b_id_storage_key.strip_prefix("0x").unwrap()).unwrap();
 
 	let latest_height_b =
-		chain_a_client.query_latest_height(chain_b_client.state_machine_id()).await? - 5;
+		chain_a_client.query_latest_height(chain_b_client.state_machine_id()).await? - 10;
```
Wizdave97, line 333: Use a finalized height, and it should be the exact height in the state machine update event.

PR #266, opened 2024-07-23
EvmHost: Support native tokens

seunlanlege:
Allow the `EvmHost` to accept the native token as payment for outgoing messages, so long as it can be swapped for the `feeToken` required to dispatch messages. This functionality has also been extended to `fundRequest` and `fundResponse`. Relayer fees are still refunded using `feeToken`.

This should make Hyperbridge's UX/DX more seamless, and it looks something like:

```solidity
IDispatcher(_hostAddr).dispatch{value: cost}(DispatchPost({}));
```

This is achieved by swapping the native token for the feeToken using a local `IUniswapV2Router02` instance. The address for this instance is tracked in the `HostParams`


evm/src/hosts/EvmHost.sol
```diff
@@ -316,18 +313,18 @@ abstract contract EvmHost is IIsmpHost, IHostManager, Context {
 
     // An application has accessed the Hyperbridge state commitment
     event StateCommitmentRead(
-    	// the application responsible
-    	address indexed caller,
-     	// The fee that was paid
-     	uint256 fee
+        // the application responsible
```
Wizdave97, line 316: Is this event still used?

seunlanlege: nahh it won't good catch

Wizdave97:
PR looks good to me.

Wizdave97 (approved):
Looks good.

Issue #187, opened 2024-04-26
OffchainDB MMR pruning


seunlanlege:
With #184 merged, we now store mmr nodes fully offchain. With the peaks stored on-chain. This is great but as the mmr grows, so does the storage requirements for node runners (See #185). 

It's clear that long-term we will need to introduce pruning for the mmr stored in the offchaindb. Luckily, MMRs are very amenable to pruning. More so, we will be focusing on pruning ancient subtrees whose leaves (requests) have already timed out. Meaning knowledge of their intermediate nodes are infact useless to relayers. Relayers & Users will only care about more recent subtrees in the MMR. This should keep the storage requirements for hyperbridge low enough that it can remain a radically inclusive & decentralized network.

PR #269, opened 2024-07-26
offchain DB MMR prunning functionality and tests

MrishoLukamba:
per #187 , 
The pruning is based on depth acting as a storage size cap for the request leaves in the offchain DB


seunlanlege:
Closing this for now because pruning isn't an immediate problem and is restricted entirely to the client-side of things, not the runtime.

PR #273, opened 2024-07-27
Migrated StateMachine Ethereum enum variant to address opaque bytes instead of EVM enum type

MrishoLukamba:


seunlanlege:
https://github.com/polytope-labs/hyperbridge/pull/270

PR #274, opened 2024-07-28
Updated contract addresses and abis used for indexing

omadoyeabraham:
- Updated contract addresses used in subquery indexers
- Updated indexers to use new field names from updated contracts


seunlanlege (approved):

indexers/src/handlers/events/evmHost/postRequest.event.handler.ts
```diff
@@ -32,14 +35,14 @@ export async function handlePostRequestEvent(
...
-    data,
+    body
   );
 
   // Create the request entity
   await RequestService.findOrCreate({
     chain,
     commitment: request_commitment,
-    data,
+    data: body,
```
seunlanlege, line 45: can we change this field name to `body`

seunlanlege:
Also i think we should change all height fields from BigInt to normal Ints so they're not strings and we can actually make queries based on filters

indexers/schema.graphql
```diff
@@ -6,189 +6,189 @@
...
 """
 enum SupportedChain {
-    ETHE
-    ARBI
-    OPTI
-    BASE
-    BSC
-    POLY
-    HYPERBRIDGE
+  ETHE
```
Wizdave97, line 9: We've changed the state machine ids for ethereum chains, you need to update the`extractStateMachineIdFromSubstrateEventData`

Or is there a way to remove this enum and just use a string to represent state machines all through

indexers/src/addresses/state-machine.addresses.ts
```diff
@@ -0,0 +1,58 @@
...
+      startBlock: 8906802,
+      isEvm: true,
+      isEvmL2: true,
+      ethereumHostAddress: "0x27D689e361ab92aCab04Ea21c1B3F507A94a9DAd",
+    }, // Optimism Sepolia
+    "EVM-421614": {
+      chainId: "421614",
+      startBlock: 20034995,
+      isEvm: true,
+      isEvmL2: false,
```
Wizdave97, line 36:
```suggestion
      isEvmL2: true,
```

indexers/src/addresses/state-machine.addresses.ts
```diff
@@ -0,0 +1,58 @@
...
+      ethereumHostAddress: "0x15Ba7e42BC2c3e8FeDEb30D13CEE611D97315E7F",
+    }, // Arbitrum Sepolia
+    BSC: {
```
Wizdave97, line 39:
```suggestion
   "EVM-97": {
```

Wizdave97 (approved):

PR #277, opened 2024-07-30
tesseract messaging integration test fix

MrishoLukamba:
refer to PR #251


tesseract/integration-test/src/lib.rs
```diff
@@ -392,57 +320,52 @@ async fn get_request_works() -> Result<(), anyhow::Error> {
...
-		.all_events_in_block()
-		.clone();
+	chain_a_sub_client.get_request(get_request_param).await?;
 
-	let tx_block_hash = tx_result.block_hash();
-	let tx_block_height = client_a.blocks().at(tx_block_hash).await?.number() as u64;
-	let events = client_a.events().at(tx_block_hash).await?;
-	let event = events.find_last::<RequestEvent>()?.unwrap();
-	log::info!("Ismp Events: {:?} \n", event);
+	let event = chain_a_sub_client.pallet_ismp_demo_events_stream(1, "Ismp", "Request").await?;
```
Wizdave97, line 344:
```suggestion
	let event = chain_a_sub_client.pallet_ismp_demo_events_stream(1, "Ismp", "Request").await?;
```
I prefer you don't use this function, but rather modify the function that submits the tx to return the block hash where the Tx was included, then fetch events from that block.

tesseract/integration-test/src/lib.rs
```diff
@@ -392,57 +320,52 @@ async fn get_request_works() -> Result<(), anyhow::Error> {
...
-		Pair::from_string("//Dave", None).expect("Unable to create Dave account"),
-	);
 	// parachain info pallet fetching para id
 	let encoded_chain_b_id_storage_key =
 		"0x0d715f2646c8f85767b5d2764bb2782604a74d81251e398fd8a0a4d55023bb3f";
+	let key = hex::decode(encoded_chain_b_id_storage_key.strip_prefix("0x").unwrap()).unwrap();
 
-	let latest_height_b =
-		chain_a_client.query_latest_height(chain_b_client.state_machine_id()).await? - 5;
+	let latest_height_b = chain_b_client.query_finalized_height().await?;
```
Wizdave97, line 333: This is not quite it, fetch the finalized height of `chain_a`, then query the latest state machine height of `chain_b` on `chain_a` at the finalized height.

Wizdave97 (changes requested):

tesseract/integration-test/src/lib.rs
```diff
@@ -392,57 +320,51 @@ async fn get_request_works() -> Result<(), anyhow::Error> {
...
-		Pair::from_string("//Dave", None).expect("Unable to create Dave account"),
-	);
 	// parachain info pallet fetching para id
 	let encoded_chain_b_id_storage_key =
 		"0x0d715f2646c8f85767b5d2764bb2782604a74d81251e398fd8a0a4d55023bb3f";
+	let key = hex::decode(encoded_chain_b_id_storage_key.strip_prefix("0x").unwrap()).unwrap();
 
-	let latest_height_b =
-		chain_a_client.query_latest_height(chain_b_client.state_machine_id()).await? - 5;
+	// as we are fetching data that is being set since genesis para id we just set the fetching height to be 1
```
Wizdave97, line 333: Use the recommended method, you can't be sure height 1 will always be available.

Wizdave97 (changes requested):
Also revert the changes to yarn.lock

MrishoLukamba:
> Also revert the changes to yarn.lock

didnt understand, my branch is at the tip of main So where should i revert to?

Wizdave97:
> > Also revert the changes to yarn.lock
> 
> didnt understand, my branch is at the tip of main So where should i revert to?

Just discard the changes to the file using git reset or git checkout.

tesseract/integration-test/src/lib.rs
```diff
@@ -392,57 +320,63 @@ async fn get_request_works() -> Result<(), anyhow::Error> {
...
-	let dave_signer = PairSigner::<Hyperbridge, _>::new(
-		Pair::from_string("//Dave", None).expect("Unable to create Dave account"),
-	);
 	// parachain info pallet fetching para id
 	let encoded_chain_b_id_storage_key =
 		"0x0d715f2646c8f85767b5d2764bb2782604a74d81251e398fd8a0a4d55023bb3f";
+	let key = hex::decode(encoded_chain_b_id_storage_key.strip_prefix("0x").unwrap()).unwrap();
+
+	let mut latest_fetch_height = 0;
+	let mut state_machine_update_b = chain_a_sub_client.state_machine_update_notification(chain_b_sub_client.state_machine_id()).await?;
```
Wizdave97, line 334: You can take the first yielded value from the stream directly instead of using the loop, which is much simpler.
```suggestion
	let mut state_machine_update_b = chain_a_sub_client.state_machine_update_notification(chain_b_sub_client.state_machine_id()).take(1).await?;
```

PR #290, opened 2024-08-12
Support world state queries for EVM get requests

Wizdave97:
Allow querying world state from EVM chains in get requests.
Fixes how state machine Id is extracted from hyperbridge events.


seunlanlege (approved):
beautiful

PR #298, opened 2024-08-21
Fix bug where pruned relay chain block breaks ismp parachain inherent creation

### LINKED ISSUE #296: Reading state from a pruned relay chain block
Recently, our parachain on Rococo was down for about 40 minutes due to an unrelated issue. Once it was back online, block production was failing because the ISMP inherent was attempting to read a relay chain block that had been pruned by the time the chain resumed. We resolved the issue by manually disabling the inherent on the client side. Once block production resumed, we were able to re-enable it.

This issue could have been prevented if, instead of stalling block production, the inherent skipped the pruned block.

The error message:
```
UnknownBlock: State already discarded for 0x066781b71140ad16b12cc4c0feffd8a639f201cd8483f70d91cc50144fb927d4
```

Wizdave97:
Closes #296 

Skips inherent creation if relaychain block has been pruned.
Handles error in inherent execution, so block production is not broken if inherent becomes invalid during a reorg.

Update version number for ismp crates.

Kill orphaned background tasks in messaging relayer.

### DIFF
--- a/modules/ismp/clients/parachain/client/src/lib.rs
+++ b/modules/ismp/clients/parachain/client/src/lib.rs
@@ -116,8 +116,16 @@ pub mod pallet {
 				data.consensus_state_id, PARACHAIN_CONSENSUS_ID,
 				"Only parachain consensus updates should be passed in the inherents!"
 			);
-			pallet_ismp::Pallet::<T>::handle_messages(vec![Message::Consensus(data)])?;
-			ConsensusUpdated::<T>::put(true);
+
+			// Handling error will prevent this inherent from breaking block production is there's a
+			// reorg and it's no longer valid
+			if let Err(err) =
+				pallet_ismp::Pallet::<T>::handle_messages(vec![Message::Consensus(data)])
+			{
+				log::trace!(target: "ismp", "Parachain inherent consensus update failed {err:?}");
+			} else {
+				ConsensusUpdated::<T>::put(true);
+			}
 
 			Ok(Pays::No.into())
 		}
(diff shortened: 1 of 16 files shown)


modules/ismp/clients/parachain/client/src/lib.rs
```diff
@@ -116,8 +116,16 @@ pub mod pallet {
...
+
+			// Handling error will prevent this inherent from breaking block production if there's a
+			// reorg and it's no longer valid
+			if let Err(err) =
+				pallet_ismp::Pallet::<T>::handle_messages(vec![Message::Consensus(data)])
+			{
+				log::trace!(target: "ismp", "Parachain inherent consensus update failed {err:?}");
+			} else {
+				ConsensusUpdated::<T>::put(true);
+			}
```
seunlanlege, line 128: not sure i understand the motivation here

Wizdave97: When I run the inherent locally and there's a reorg, the collator misses block production for that slot because it reinserts the inherent from the reorged block into the new block.
The inherent is no longer valid at that point.

Wizdave97: I'll paste an example here in a bit.

Wizdave97: image

seunlanlege: Ahh ok

seunlanlege (approved):

PR #341, opened 2024-11-12
Fully disable validate_unsigned when feature not active

blakebyrnes:
The validate_unsigned attribute was still active when the unsigned feature isn't active, which means you can't actually disable unsigned messages.


seunlanlege (approved):
Good catch!

PR #343, opened 2024-11-14
Control unsigned ismp with a storage item

blakebyrnes:
This PR adds a storage item, genesis setting and runtime api to control the activation of unsigned transactions for the ISMP pallet.

The unsigned feature flag is removed, and the development testnet configs have the enableUnsignedTransactions flag turned on by default.

NOTE: I did not update any chain specs.

### DIFF
--- a/modules/ismp/pallets/pallet/README.md
+++ b/modules/ismp/pallets/pallet/README.md
@@ -32,7 +32,7 @@ To use it in your runtime, you need to implement the ismp
 ### Dispatchable Functions
 
 - `handle` - Handles incoming ISMP messages.
-- `handle_unsigned` Unsigned variant for handling incoming messages, enabled by `feature = ["unsigned"]`
+- `handle_unsigned` Unsigned variant for handling incoming messages
 - `create_consensus_client` - Handles creation of various properties for a particular consensus client. Can only be called by the `AdminOrigin`.
 - `update_consensus_state` - Updates consensus client properties in storage. Can only be called by the `AdminOrigin`.
 - `fund_message` - In cases where the initially provided relayer fees have now become insufficient, due to a transaction fee spike on the destination chain. Allows a user to add more funds to the request to be used for delivery and execution. Should never be called on a completed request.
--- a/modules/ismp/pallets/pallet/src/lib.rs
+++ b/modules/ismp/pallets/pallet/src/lib.rs
@@ -62,8 +62,7 @@
 //! ### Dispatchable Functions
 //!
 //! * `handle` - Handles incoming ISMP messages.
-//! * `handle_unsigned` Unsigned variant for handling incoming messages, enabled by `feature =
-//!   ["unsigned"]`
+//! * `handle_unsigned` Unsigned variant for handling incoming messages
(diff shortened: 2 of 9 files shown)


modules/ismp/pallets/pallet/src/lib.rs
```diff
@@ -394,6 +405,24 @@ pub mod pallet {
...
 
+	/// Genesis settings
+	#[pallet::genesis_config]
+	#[derive(frame_support::DefaultNoBound)]
+	pub struct GenesisConfig<T: Config> {
+		/// Should unsigned transactions be enabled. Preferred mode for testnet to simplify relayer
+		pub enable_unsigned_transactions: bool,
+		/// Phantom data
+		#[serde(skip)]
+		pub _phantom: PhantomData<T>,
```
seunlanlege, line 416: Should be private

```suggestion
	 _phantom: PhantomData<T>,
```

modules/ismp/pallets/pallet/README.md
```diff
@@ -32,7 +32,7 @@ To use it in your runtime, you need to implement the ismp
 ### Dispatchable Functions
 
 - `handle` - Handles incoming ISMP messages.
-- `handle_unsigned` Unsigned variant for handling incoming messages, enabled by `feature = ["unsigned"]`
+- `handle_unsigned` Unsigned variant for handling incoming messages
```
seunlanlege, line 35: We would need to also consolidate these calls

modules/ismp/pallets/pallet/src/lib.rs
blakebyrnes: Don't think it can be since you build these from outside the module.

seunlanlege: Oh yeah GenesisConfig is no longer strongly typed

modules/ismp/pallets/pallet/README.md
blakebyrnes: For instance, depending on how much backward compatibility you want to preserve, we could optionally take out index 0 and start them `handle`, but if you think that will break people, we should leave handle at 1
```
/// Execute the provided batch of ISMP messages. This call will short-circuit and revert if
		/// any of the provided messages are invalid.
		///
		/// The dispatch origin for this call must be an unsigned one.
		///
		/// - `messages`: A set of ISMP [`Message`]s to handle or process.
		///
		/// Emits different message events based on the Message received if successful.
		#[pallet::weight(get_weight::<T>(&messages))]
		#[pallet::call_index(1)]
		#[frame_support::transactional]
		pub fn handle(origin: OriginFor<T>, messages: Vec<Message>) -> DispatchResultWithPostInfo {
			if EnableUnsigned::<T>::get() {
				ensure_none(origin)?;
			} else {
				ensure_signed(origin)?;
			}

			Self::handle_messages(messages)
		}```

blakebyrnes:
We're putting this on pause for now. Need to merge handle and handle_message, which means anywhere that is currently looking up whether handle_message is available as a call will need a storage lookup to see if the api is enabled.

PR #362, opened 2025-01-02
Upgrade alloy&mmr dependencies

kku9706:


seunlanlege:
Why should we upgrade the MMR?

gavofyork:
image

^^^ Eeek

kku9706:
> Why should we upgrade the MMR?

Somehow be consistent with https://github.com/paritytech/polkadot-sdk/blob/f3ab3854e1df9e0498599f01ba4f9f152426432a/substrate/frame/merkle-mountain-range/src/mmr/storage.rs#L141

Though It may not be necessary, is it possible to reuse the MMR library from Parity directly?

kku9706:
About rust toolchain,  does https://github.com/paritytech/dockerfiles/blob/2ab2cf3a4efb4a4f7539f42b064c8a42d308d450/ci-unified/Dockerfile#L91
 mean Parity already updated to use 1.81?

seunlanlege:
Unfortunately, given how sensitive the mmr is to the safety of Hyperbridge, the fact that it is already deployed on mainnet, alongside solidity verifiers, we will not be upgrading the mmr. 

The current version in use (v0.5.2) underwent significant audits and testing to ensure safety and compatibility with the [solidity-merkle-trees library](https://github.com/polytope-labs/solidity-merkle-trees). Which is used for mmr verification on evm chains. 

Now `polkadot-sdk` seems to be using a [fork built on the v0.6.0 version](https://github.com/paritytech/merkle-mountain-range) which is quite concerning given they include the v0.5.2 in their root [Cargo.toml](https://github.com/paritytech/polkadot-sdk/blob/master/Cargo.toml#L866) but actually use a different version in [`sp-mmr-primitives`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/merkle-mountain-range/Cargo.toml#L20). 

We will however, take some time to investigate in detail the [diff](https://github.com/nervosnetwork/merkle-mountain-range/compare/v0.5.0...v0.6.0) between `0.5.2` and `0.6.0` once we have the bandwidth to do so.

@kku9706 Do not be dismayed, you are welcome to contribute by tackling our other open issues 
https://github.com/polytope-labs/hyperbridge/issues

PR #366, opened 2025-01-15
Allow updates to asset precision

Wizdave97:


seunlanlege (approved):

seunlanlege:
Think we need to update the docs as well

PR #385, opened 2025-02-13
introduce IntentGateway

seunlanlege:
As alluded to in https://github.com/polytope-labs/hyperbridge/pull/306, This PR introduces a new protocol: `IntentGateway` which will enable the exchange of ERC20 assets using Hyperbridge for settlement and asset recovery. It does not implement the ERC-7683 standard. Which is currently still in draft phase and in flux.

 - [x] Introduce the IntentGateway contracts
   - [x] Add tests
 - [x] Introduce `UniV3UniswapV2Wrapper` which wraps UniswapV3 deployments with a v2 interface.
 - [x] Introduce `IntentGateway` governance to `pallet-token-governor`


evm/src/modules/IntentGateway.sol
```diff
@@ -0,0 +1,609 @@
...
+        // escrow tokens
+        uint256 msgValue = msg.value;
+        uint256 inputsLen = order.inputs.length;
+        for (uint256 i = 0; i < inputsLen; i++) {
+            if (order.inputs[i].amount == 0) revert InvalidInput();
+            address token = address(uint160(uint256(order.inputs[i].token)));
+            if (token == address(0)) {
+                // native token
+                if (msgValue < order.inputs[i].amount) revert InsufficientNativeToken();
+                msgValue -= order.inputs[i].amount;
```
Wizdave97, line 336: Does this line automatically escrow  the native asset?
```suggestion
                msgValue -= order.inputs[i].amount;
```

seunlanlege: Yes the amount in msg.value has been deposited into the contract

PR #386, opened 2025-02-20
pallet-token-gateway: UX improvement

seunlanlege:
This patch allows the sender to execute runtime calls without any signatures, by using the sender information directly as the dispatch origin.

### DIFF
--- a/modules/pallets/token-gateway/src/lib.rs
+++ b/modules/pallets/token-gateway/src/lib.rs
@@ -618,58 +618,73 @@ where
 		if let Some(call_data) = body.data {
 			let substrate_data = SubstrateCalldata::decode(&mut &call_data.0[..])
 				.map_err(|err| anyhow!("Calldata decode error: {err:?}"))?;
-			// Verify signature against encoded runtime call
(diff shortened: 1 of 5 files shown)


modules/pallets/token-gateway/src/lib.rs
```diff
@@ -618,58 +618,73 @@ where
...
+						}
+					},
+				};
+
+				beneficiary.clone().into()
+			} else {
+				// check if the first 12 bytes of body.from is all zeroes
+				if body.from[0..12] == [0u8; 12] {
+					// sender is evm account
+					T::EvmToSubstrate::convert(H160::from_slice(&body.from[12..]))
```
Wizdave97, line 673: Is it worth it to enforce that the source chain must be an evm chain before doing this?

Wizdave97 (approved):

modules/pallets/token-gateway/src/lib.rs
seunlanlege: I also want to substrate chains

Issue #239, opened 2024-06-17
Separate L2 Updates from sync committee updates


Wizdave97:
It would be much more maintainable to allow Eth L2 State machine updates to happen independent of the beacon consensus client update.

PR #390, opened 2025-02-28
separate L2 update as a standalone function

varun-doshi:
Ref #239 

Note: this creates an additional trait function, hence currently is a breaking change. Will update as adviced

### DIFF
--- a/modules/ismp/clients/sync-committee/src/beacon_client.rs
+++ b/modules/ismp/clients/sync-committee/src/beacon_client.rs
@@ -13,10 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use core::marker::PhantomData;
+
 use alloc::{collections::BTreeMap, format, string::ToString};
-use arbitrum_verifier::{verify_arbitrum_bold, verify_arbitrum_payload};
+use arbitrum_verifier::{
+	verify_arbitrum_bold, verify_arbitrum_payload, ArbitrumBoldProof, ArbitrumPayloadProof,
+};
 use codec::{Decode, Encode};
 use evm_state_machine::construct_intermediate_state;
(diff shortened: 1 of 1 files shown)


modules/ismp/core/src/consensus.rs
```diff
@@ -139,6 +139,15 @@ pub trait ConsensusClient {
...
 
+	/// Process the L2 state update
+	fn verify_l2_state_update(
+		&self,
+		trusted_consensus_state: Vec<u8>,
+		consensus_state_id: ConsensusStateId,
+		state_machine_map: &mut BTreeMap<StateMachine, Vec<StateCommitmentHeight>>,
+		consensus_proof: Vec<u8>
+		)-> Result<Vec<u8>, Error>;
+
```
seunlanlege, line 150: So this isn't really the intended direction. Rather you should construct new implementations of the `ConsensusClient` trait for the various L2 proof types. This trait implementation will be configured with the state machine id of the ethereum chain, and accepts storage proofs of ethereum, proving new finalized L2 state commitments

modules/ismp/clients/sync-committee/src/beacon_client.rs
```diff
@@ -25,211 +30,385 @@ use crate::{
...
+					Err(_) => {
+						return Err(Error::Custom("Cannot decode update".to_string()));
+					},
+				}
+			},
+		};
+
+		// Process full beacon update with optional L2 updates
+		self.process_beacon_update(host, consensus_state_id, consensus_state, update)
+	}
```
seunlanlege, line 323: The goal is that the verification of L2 state assertions to happen independent of the sync-committee client so it can be updated independently, potentially replacing it entirely with our future zk-casper light client. So you'd have to put the verifiers into new crates

seunlanlege:
stale

PR #391, opened 2025-02-28
Try to prove block ranges for GRANDPA

seunlanlege:
The `grandpa_proveFinality` seems to return proofs only for the latest finalized height instead of for blocks within the epoch. This updates the prover to only prove blocks within the requested range


Wizdave97 (approved):
LGTM

PR #413, opened 2025-04-19
Grandpa prover fix

Wizdave97:
Set max height to finalized height when it's greater than the latter.


seunlanlege (approved):

modules/consensus/grandpa/prover/src/lib.rs
```diff
@@ -224,6 +224,8 @@ where
 				return Ok(finality_proof);
 			}
 			log::trace!("grandpa_proveFinality proof for {} could not be verified, switching to linear search", self.options.state_machine);
+			// Set max the height as the finalized block number
```
royvardhan, line 227: I think should be 'set max height'

royvardhan (approved):

PR #431, opened 2025-06-09
fix: use estimated gas with buffer

royvardhan:


tesseract/evm/src/tx.rs
```diff
@@ -460,6 +470,14 @@ pub fn get_chain_gas_limit(state_machine: StateMachine) -> u64 {
 	}
 }
 
+pub fn apply_gas_limit_buffer(estimated_gas: u64, buffer_percentage: Option<u32>) -> u64 {
+	if let Some(buffer) = buffer_percentage {
+		estimated_gas + ((estimated_gas * buffer as u64) / 100)
+	} else {
+		estimated_gas
```
Wizdave97: Let's use a default buffer of 5%

royvardhan: Added.

Wizdave97 (approved):

PR #432, opened 2025-06-09
Implementation for Pallet Relayer Incentives

dharjeezy:
This PR introduces a new pallet called Relayer incentives for incentivizing relayers that relays consensus messages.

It introduces a new storage item  in pallet ISMP called `PreviousStateMachineHeight` for storing the previous state machine height for a State machine.

It also modifies the `FeeHandler` trait by including `Events` to the `on_executed` trait function.

- [x] implementation
- [ ] tests


modules/pallets/ismp/src/host.rs
```diff
@@ -314,6 +316,10 @@ impl<T: Config> IsmpHost for Pallet<T> {
 		Responded::<T>::insert(req_commitment, true);
 		Ok(())
 	}
+
+	fn previous_commitment_height(&self, id: StateMachineId) -> Result<u64, Error> {
+		Ok(PreviousStateMachineHeight::<T>::get(id).unwrap_or_default())
```
seunlanlege, line 321: Maybe lets make this function return an option instead since the underlying storage is option query

PR #434, opened 2025-06-11
improve relayer incentives

dharjeezy:
- [x] Removal of inconsequential storage items
- [x] Improve `RelayerRewarded` event data by adding StateMachineHeight


modules/pallets/relayer-incentives/src/lib.rs
```diff
@@ -109,8 +92,8 @@ pub mod pallet {
 			relayer: T::AccountId,
 			/// Amount of the reward
 			amount: <T as pallet_ismp::Config>::Balance,
-			/// Message identifier that was processed
-			message_id: H256,
+			/// The state machine height at which the reward happens
```
seunlanlege:
```suggestion
			///  Metadata about the state machine and height that was rewarded
```

PR #438, opened 2025-06-17
Add tesseract-consensus to Hyperbridge

dharjeezy:


tesseract-consensus/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
```
Wizdave97: Delete this file

tesseract-consensus/rustfmt.toml
```diff
@@ -0,0 +1,23 @@
+# Basic
```
Wizdave97: Also this

Issue #439, opened 2025-06-17
Ismp Parachain Consensus Verification Bug


Wizdave97:
When more than one parachain is added to the ismp parachain pallet, the proofs will not always contain a header for each parachain at the same time; inherent only fetches proofs for parachains whose headers have changed. 

The consensus client currently does not consider this and fails for proofs that don't contain headers for all supported parachains.
We get errors that the key for a certain parachain is missing, and the entire proof fails

![Image](https://github.com/user-attachments/assets/ea024d9d-be93-4232-a4a7-625fc94b9be3)

We should skip parachains whose keys are not in the proof.

PR #441, opened 2025-06-20
do not decode parachain headers whose keys are not in the proof

dharjeezy:
closes https://github.com/polytope-labs/hyperbridge/issues/439

### DIFF
--- a/modules/ismp/state-machines/substrate/src/lib.rs
+++ b/modules/ismp/state-machines/substrate/src/lib.rs
@@ -300,13 +300,23 @@ where
 	let trie = TrieDBBuilder::<LayoutV0<H>>::new(&db, root).build();
 	let mut result = BTreeMap::new();
 
-	for key in keys.into_iter() {
-		let value = trie
+	for key in keys {
+		let raw_val = trie
 			.get(key.as_ref())
-			.map_err(|e| Error::Custom(format!("Error reading from trie: {e:?}")))?
-			.and_then(|val| Decode::decode(&mut &val[..]).ok());
-		result.insert(key.as_ref().to_vec(), value);
+			.map_err(|e| Error::Custom(format!("Error reading from trie: {e:?}")))?;
+
+		let maybe_decoded = match raw_val {
+			Some(val) => {
+				Some(Decode::decode(&mut &val[..])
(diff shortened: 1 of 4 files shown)


modules/ismp/state-machines/substrate/src/lib.rs
seunlanlege, line 1: I don't think this is where the problem is

modules/ismp/state-machines/substrate/src/lib.rs
```diff
@@ -300,14 +300,24 @@ where
...
 
-	for key in keys.into_iter() {
-		let value = trie
+	for key in keys {
+		let raw_val = trie
 			.get(key.as_ref())
-			.map_err(|e| Error::Custom(format!("Error reading from trie: {e:?}")))?
-			.and_then(|val| Decode::decode(&mut &val[..]).ok());
-		result.insert(key.as_ref().to_vec(), value);
+			.map_err(|e| Error::Custom(format!("Error reading from trie: {e:?}")))?;
```
seunlanlege: If you try to fetch a key that's not in the proof. You'll get this error anyways.

Wizdave97: Yeah, skip keys that return `IncompleteDatabase` errors.

modules/ismp/state-machines/substrate/src/lib.rs
```diff
@@ -300,12 +300,22 @@ where
...
+		match trie.get(raw_key) {
+			Ok(Some(val)) => {
+				let decoded = Decode::decode(&mut &val[..])
+					.map_err(|e| Error::Custom(format!("Decode error: {e:?}")))?;
+				result.insert(raw_key.to_vec(), Some(decoded));
+			},
+			Ok(None) => {
+				result.insert(raw_key.to_vec(), None);
+			},
+			Err(_e) => {
```
seunlanlege: This isn't correct still, you have to match over the exact error

dharjeezy: Ok. What should happen if it is an error other than the Incomplete Database Error? Fail outrightly right?

seunlanlege: Propagate the error

seunlanlege (approved):

PR #446, opened 2025-06-24
feat: add tendermint verifier and prover

royvardhan:
#442 

- [x] Tendermint Verifier
- [x] Tendermint Prover

Issue #442, opened 2025-06-23
Tracking issue for Ismp Tendermint/ComeBFT integration


Wizdave97:
An Ismp Consensus client for tendermint/cometBFT

- [ ] Tendermint/CometBFT verifier
- [ ] Tendermint/CometBFT Ismp consensus client
- [ ] Ics23 state machine client
- [ ] Tendermint consensus relayer

PR #448, opened 2025-07-02
add tendermint verifer with prover

royvardhan:
#442 

- [x] Tendermint Verifier
- [x] Tendermint Prover

previous conversation: https://github.com/polytope-labs/hyperbridge/pull/446


modules/ismp/clients/tendermint/prover/src/peppermint_rpc.rs
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	pub async fn signed_header(&self, height: u64) -> Result<SignedHeader, ProverError> {
+		let response = self.raw_request("commit", json!({"height": height.to_string()})).await?;
+		let signed_header = response
+			.get("signed_header")
+			.ok_or_else(|| ProverError::RpcError("No signed_header in response".to_string()))?;
+
+		// Detect if this is a Polygon/peppermint endpoint (by chain_id or network name)
+		let is_polygon = self.is_polygon_chain().await.unwrap_or(false);
```
seunlanlege: why not have this client be generic over the header?

modules/ismp/clients/tendermint/prover/src/peppermint_rpc.rs
```diff
@@ -0,0 +1,218 @@
...
+							if let Some(type_val) = precommit_obj.get("type") {
+								signature.insert("block_id_flag".to_string(), type_val.clone());
+							}
+							// Handle secp256k1 signature format (65 bytes -> 64 bytes)
+							if let Some(sig) = precommit_obj.get("signature") {
+								if let Some(sig_str) = sig.as_str() {
+									if let Ok(sig_bytes) =
+										base64::engine::general_purpose::STANDARD.decode(sig_str)
+									{
+										if sig_bytes.len() == 65 {
```
seunlanlege: nit: this is too deeply nested, can we flatten?

royvardhan: Yeah agreed. Cleanup is left still. One polygon test is failing. Debugging.

modules/ismp/clients/tendermint/prover/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,115 @@
...
+		println!("Testing Polygon's Peppermint Fork (Full Verification)");
+		run_integration_test(POLYGON_RPC_URL).await.unwrap();
+	}
+
+	/// Full integration test: prover and verifier
+	async fn run_integration_test(rpc_url: &str) -> Result<(), Box<dyn std::error::Error>> {
+		let client = TendermintRpcClient::new(rpc_url).await?;
+		ensure_healthy(&client).await?;
+		let chain_id = client.chain_id().await?;
+		let latest_height = client.latest_height().await?;
```
Wizdave97: We should allow these tests run over 3-4 validator set transitions

modules/ismp/clients/tendermint/prover/src/peppermint_rpc.rs
```diff
@@ -57,6 +59,7 @@ impl PeppermintRpcClient {
 	}
 
 	pub async fn validators(&self, height: u64) -> Result<Vec<Validator>, ProverError> {
+		println!("Getting validators from Polygon/peppermint endpoint");
```
seunlanlege: prefer we using `tracing::trace!`

modules/ismp/clients/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,273 @@
...
+
+	let tendermint_trusted_state = TrustedBlockState {
+		chain_id: &chain_id,
+		header_time: time,
+		height,
+		next_validators: &next_validators,
+		next_validators_hash,
+	};
+
+	let validators = ValidatorSet::new(consensus_proof.validators.clone(), None);
```
Wizdave97, line 44: We should use the validators from the trusted state, the  consensus proof should only provide the next_validators optionally

modules/ismp/clients/tendermint/prover/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,259 @@
...
+
+		trace!(
+			"Starting verification with {} validator set transitions",
+			VALIDATOR_SET_TRANSITIONS
+		);
+		trace!("Initial trusted height: {}", trusted_state.height);
+
+		// Perform multiple validator set transitions
+		for transition in 1..=VALIDATOR_SET_TRANSITIONS {
+			let target_height = trusted_state.height + 5;
```
Wizdave97: For accurate validator set transitions, I think you need to use the epoch length configured for Heimdall.

Could you find out the epoch length of the heimdall chain and refactor the test to verify one consensus proof at the boundary of each epoch so we can properly simulate validator set transitions?

Wizdave97: They use something called ackCount to track epochs https://github.com/0xPolygon/heimdall-v2/blob/3a8f5abc77a3bca55da8e00f3b56aec737c518d9/x/stake/types/validator.go#L54

MIght need to add the epoch information to the trusted state

Wizdave97: On second thoughts, you can reserve the transitions through epochs for the consensus relayer implementation, that's where it would be needed.

modules/ismp/clients/tendermint/prover/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+version = "0.1.0"
+edition = "2021"
+description = "CometBFT light client prover for ISMP"
+authors = ["Polytope Labs <hello@polytope.technology>"]
+publish = false
+
+[dependencies]
+# Core dependencies
+cometbft-rpc = { git = "https://github.com/cometbft/cometbft-rs", package = "cometbft-rpc", features = ["http-client", "secp256k1"] }
+cometbft = { git = "https://github.com/cometbft/cometbft-rs", features = ["rust-crypto", "secp256k1"] }
```
seunlanlege: we should pin these git dependencies to a specific tag/branch etc

modules/ismp/clients/tendermint/prover/src/account.rs
```diff
@@ -0,0 +1,28 @@
...
+			let pubkey_bytes = pk.to_encoded_point(false).as_bytes().to_vec();
+			// Remove the 0x04 prefix (first byte) as done in Go implementation
+			let keccak_hash = Keccak256::digest(&pubkey_bytes[1..]);
+			// Take last 20 bytes (bytes 12-31)
+			CometbftAccountId::new(keccak_hash[12..32].try_into().unwrap())
+		},
+		#[allow(unreachable_patterns)]
+		_ => {
+			// Catch-all for non_exhaustive enum
+			panic!("Unsupported public key type for Polygon/Heimdall verification")
```
seunlanlege: no panicking in the runtime

modules/ismp/clients/tendermint/prover/src/client.rs
```diff
@@ -0,0 +1,222 @@
...
+	async fn validators(&self, height: u64) -> Result<Vec<Validator>, ProverError>;
+	async fn next_validators(&self, height: u64) -> Result<Vec<Validator>, ProverError>;
+	async fn signed_headers_range(
+		&self,
+		start_height: u64,
+		end_height: u64,
+	) -> Result<Vec<SignedHeader>, ProverError>;
+	async fn chain_id(&self) -> Result<String, ProverError>;
+	async fn is_healthy(&self) -> Result<bool, ProverError>;
+}
```
seunlanlege: Missing docs. Lets add a `#![deny(missing_docs)]` to lib.rs

modules/ismp/clients/tendermint/prover/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+pub use cometbft::{
```
seunlanlege:
```suggestion
#![deny(missing_docs)]
pub use cometbft::{
```

modules/ismp/clients/tendermint/verifier/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+[dependencies]
+# Core dependencies
+cometbft-light-client-verifier = { git = "https://github.com/cometbft/cometbft-rs", package = "cometbft-light-client-verifier" }
+cometbft = { git = "https://github.com/cometbft/cometbft-rs", features = ["rust-crypto", "secp256k1"] }
+cometbft-proto = { git = "https://github.com/cometbft/cometbft-rs", package = "cometbft-proto" }
+thiserror = "2.0.12"
+serde = { version = "1.0.219", features = ["derive"], optional = true }
+prost = "0.13"
+sha3 = "0.10"
+k256 = { version = "0.13", features = ["ecdsa"] }
```
seunlanlege: this library should compile for the no-std target. You'll need to turn off default features for the dependencies and add the #![no_std] tag to the library

modules/ismp/clients/tendermint/verifier/src/hashing.rs
```diff
@@ -0,0 +1,90 @@
...
+				let pub_key = polkadot_sdk::sp_core::ed25519::Public::try_from(pk.as_bytes())
+					.map_err(|_| signature::Error::MalformedPublicKey)?;
+
+				if sp_io::crypto::ed25519_verify(&sig, msg, &pub_key) {
+					Ok(())
+				} else {
+					Err(signature::Error::VerificationFailed)
+				}
+			},
+			#[cfg(feature = "secp256k1")]
```
seunlanlege: why is this feature flagged?

modules/ismp/clients/tendermint/prover/Cargo.toml
seunlanlege, line 1: Move this tendermint folder to `modules/consensus/`. Thats where verifiers live. Only ismp consensus client implementations live in `modules/ismp/clients`.

modules/ismp/clients/tendermint/verifier/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+pub use cometbft::{
```
seunlanlege:
```suggestion
#![cfg_attr(not(feature = "std"), no_std)]
pub use cometbft::{
```

modules/ismp/clients/tendermint/verifier/src/sp_io_verifier.rs
```diff
@@ -0,0 +1,176 @@
...
+		&self,
+		signed_header: &SignedHeader,
+		validators: &ValidatorSet,
+		commit_validator: &dyn CommitValidator,
+	) -> Result<(), VerificationError> {
+		commit_validator.validate(signed_header, validators)?;
+		commit_validator.validate_full(signed_header, validators)?;
+
+		Ok(())
+	}
```
seunlanlege: This changes nothing from the default implementation, best to remove

modules/ismp/clients/tendermint/verifier/src/types.rs
```diff
@@ -0,0 +1,337 @@
...
+	/// Next validator set
+	pub next_validators: Vec<Validator>,
+	/// Hash of the next validator set
+	pub next_validators_hash: [u8; 32],
+	/// Trusting period in seconds
+	pub trusting_period: u64,
+	/// Verification options for this consensus state
+	pub verification_options: VerificationOptions,
+	/// Frozen height (if any)
+	pub frozen_height: Option<u64>,
```
seunlanlege: Unneeded, ismp handles freezing outside the consensus state

modules/ismp/clients/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,273 @@
...
```
seunlanlege: I think david already asked this and i somehow missed it, But the validators shouldn't come from the consensus proof. They should be known to the verifier in the trusted state

modules/ismp/clients/tendermint/verifier/src/hashing.rs
```diff
@@ -0,0 +1,90 @@
...
+					result = sp_io::crypto::ecdsa_verify_prehashed(&sig, &msg_hash, &pub_key);
+				}
+
+				if result {
+					Ok(())
+				} else {
+					Err(signature::Error::VerificationFailed)
+				}
+			},
+			_ => Err(signature::Error::UnsupportedKeyType),
```
seunlanlege: This pattern is unreachable

modules/ismp/clients/tendermint/prover/src/account.rs
seunlanlege: Also this is unreachable

modules/ismp/clients/tendermint/verifier/src/types.rs
seunlanlege: a lot of unused fields on the `TrustedState` we should trim them

modules/ismp/clients/tendermint/verifier/src/verifier.rs
Wizdave97: Yeah, I've told him to find out how the tendermint verifier does membership check for the validator set internally so we know if it needs us to provide the subset of validators that signed the header.

seunlanlege: I've taken a look at the cometbft code and it performs no checks relating to the supplied validator set. I think we should supply the correct validator set from the trusted state depending on the current validators hash in the signed header.

royvardhan: Here is my thought process on why I am using validator on the consensus proof:

royvardhan: Here is my thought process on why I am using `validator` on the consensus proof, with my assumptions:

 - The `UntrustedBlockState` should hold values which are not yet verified. So if trusted height is 100, `UntrustedBlockState` will have values for target block.
 - The `verify_update_header` calls [verify_validator_sets](https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/light-client-verifier/src/verifier.rs#L119) with only the `UntrustedBlockState` block struct which does the following:
  -- Checks if `untrusted.validators` and `untrusted.signed_header.header.validators_hash` match. These validators are `target_block` values.
 -- Checks if `untrusted_next_validators` and `untrusted.signed_header.header.next_validators_hash` match. These validators are `target_block` + 1 values

 - And then it checks [validate_against_trusted](https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/light-client-verifier/src/verifier.rs#L152), where If the untrusted block is the very next block after the trusted block, it should match with their (next) validator sets hash.

modules/ismp/clients/tendermint/prover/src/tests/integration_tests.rs
royvardhan: Alright! Noted for the upcoming pr.

modules/ismp/clients/tendermint/prover/src/account.rs
royvardhan: Had to add a catch-all arm because PublickKey has [#[non_exhaustive]](https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/cometbft/src/public_key.rs#L39) defined.

I am removing the panic and adding a default zero address return for non-exhaustive enum.

modules/ismp/clients/tendermint/verifier/src/hashing.rs
royvardhan: Following the cometbft pattern [here](https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/cometbft/src/crypto/default/signature.rs#L22)

seunlanlege: I think no need to feature flag it honestly. Enable the secp256k1 feature flag on cometbft by default

modules/ismp/clients/tendermint/verifier/src/verifier.rs
seunlanlege: > I've taken a look at the cometbft code and it performs no checks relating to the supplied validator set. I think we should supply the correct validator set from the trusted state depending on the current validators hash in the signed header.

Like i said, the validator sets (both current & next) that come with the signed header **cannot be trusted**. Nothing stops literally anyone from **constructing their own fake validator set and submitting it to our verifier**. The verifier needs to ensure that only the validators it knows (either current or next) have signed the provided header.

So check the currrent validator's hash in the signed header to know which set is currently signing the header, and pass the correct validator set from the trusted state. Otherwise this verifier isn't very useful if anyone can tell it what validator set to use.

modules/ismp/clients/tendermint/verifier/src/hashing.rs
royvardhan: This is the same non-exhaustive pub key enum. So a catch-all arm.

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,272 @@
...
+		chain_id: &chain_id,
+		header_time: time,
+		height,
+		next_validators: &next_validators,
+		next_validators_hash,
+	};
+
+	let validators = ValidatorSet::new(consensus_proof.validators.clone(), None);
+	let next_validators_proof = ValidatorSet::new(consensus_proof.next_validators.clone(), None);
+
```
Wizdave97, line 47: After verifying the sitgned header correctly, It's our responsibliity to check that `untrusted_signed_header.next_validator_hash == consensus_proof.next_validators.hash()`
https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/light-client-verifier/src/verifier.rs#L278

modules/ismp/clients/tendermint/verifier/src/verifier.rs
Wizdave97: Looking into cometbft code here https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/light-client-verifier/src/verifier.rs#L230, I don't really understand why they need to verify signatures from an untrusted set, but if you follow the code to the implementation here  https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/light-client-verifier/src/predicates.rs#L213  and here  https://github.com/cometbft/cometbft-rs/blob/ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d/light-client-verifier/src/operations/voting_power.rs#L97 they check that 2/3 of `trusted.next_validators` signed the header and also that 2/3 of `untrusted.validators` also signed the header, that second check is not really useful imo. @royvardhan I believe you should use `trusted.next_validators` for both trusted state and untrusted state.

Wizdave97: If you look into the second  paragraph of the abstract in the specs here https://github.com/cometbft/cometbft-rs/blob/main/docs/spec/lightclient/verification/verification.md, 
It reads:
```In the Tendermint blockchain, the validator set may change with every new block. The staking and unbonding mechanism induces a [security model](https://github.com/cometbft/cometbft-rs/blob/main/docs/spec/lightclient/verification/verification.md#tmbc-fm-2thirds1): starting at time Time of the [header](https://github.com/cometbft/cometbft-rs/blob/main/docs/spec/lightclient/verification/verification.md#tmbc-header1), more than two-thirds of the next validators of a new block are correct for the duration of TrustedPeriod. The fault-tolerant read operation is designed for this security model.```

So `trusted_state.next_validators` is what we need to use to verify any future headers within the trusting period configured for the tendermint chain until we reach a header where `next_validators` has changed, then we rotate our view of the validator set.

File: modules/consensus/tendermint/verifier/Cargo.toml [hunk 19]
```diff
@@ -0,0 +1,34 @@
...
+[dependencies]
+# Core dependencies
+cometbft-light-client-verifier = { git = "https://github.com/cometbft/cometbft-rs", package = "cometbft-light-client-verifier", rev = "ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d"}
+cometbft = { git = "https://github.com/cometbft/cometbft-rs", features = ["rust-crypto", "secp256k1"], rev = "ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d" }
+cometbft-proto = { git = "https://github.com/cometbft/cometbft-rs", package = "cometbft-proto", rev = "ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d" }
+thiserror = "2.0.12"
+serde = { version = "1.0.219", features = ["derive"], optional = true }
+prost = "0.13"
+sha3 = "0.10"
+k256 = { version = "0.13", features = ["ecdsa"] }
```
seunlanlege: These dependencies are unused since we do cryptography stuff with `sp-io`

modules/consensus/tendermint/prover/src/prover.rs
```diff
@@ -51,18 +45,12 @@ pub async fn prove_misbehaviour_header(
 		Vec::new()
 	};
 
-	let validators = rpc_client.validators(target_height).await?;
-	if validators.is_empty() {
-		return Err(ProverError::NoValidators(target_height));
-	}
-
 	let next_validators = rpc_client.next_validators(target_height).await?;
 	if next_validators.is_empty() {
```
Wizdave97, line 49: If next_validators is empty I don't think it should be an error.

Wizdave97: Is next_validators not an optional field in the tendermint header?

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,321 @@
...
+	old_trusted_state: &TrustedState,
+	consensus_proof: &ConsensusProof,
+) -> Result<UpdatedTrustedState, VerificationError> {
+	// Check if validator set rotation is needed
+	let (validators, next_validators, new_next_validators_hash) = {
+		let signed_header_hash = &consensus_proof.signed_header.header.next_validators_hash;
+
+		// Only rotate if signed header has next_validators_hash and consensus proof has
+		// next_validators
+		if !signed_header_hash.is_empty() && consensus_proof.next_validators.is_some() {
```
Wizdave97: If `signed_header_hash` is not empty and `consensus_proof.next_validators` is None, this function should return an error. We don't want the light client to be bricked.

royvardhan: When `signed_header_hash` is empty, the `consensus_proof.next_validators` will be None. I had added this logic in the prover.

modules/consensus/tendermint/prover/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+base64 = { version = "0.21", optional = true }
+tokio = { version = "1.0", features = ["rt-multi-thread", "macros"], optional = true }
+sha2 = "0.10.7"
+cometbft-proto = { git = "https://github.com/cometbft/cometbft-rs", package = "cometbft-proto", rev = "ac3db79c5807cd1d0b8b14a7b8199b8d79d6408d"}
+prost = "0.13"
+sha3 = "0.10"
+tracing = "0.1.40"
+
+# Internal dependencies
+tendermint-verifier = { path = "../verifier" }
```
seunlanlege, line 30: Should be a dev-dependency

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,321 @@
...
+
+	let tendermint_trusted_state = TrustedBlockState {
+		chain_id: &chain_id,
+		header_time: time,
+		height,
+		next_validators: &next_validators,
+		next_validators_hash,
+	};
+
+	let validators = ValidatorSet::new(trusted_state.next_validators.clone(), None);
```
seunlanlege: This is still wrong. Your logic should look something like:

```rust
let validators = if consensus_proof.signed_header.validators_hash == trusted_state.current_validators_hash {
   ValidatorSet::new(trusted_state.validators.clone(), None)
} else if if consensus_proof.signed_header.validators_hash == trusted_state.next_validators_hash {
   ValidatorSet::new(trusted_state.next_validators.clone(), None)
} else {
    Err(Error::UnknownValidatorSet(consensus_proof.signed_header.validators_hash))?
}
```

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,321 @@
...
+	let next_validators_proof = consensus_proof
```
seunlanlege: technically not a proof

```suggestion
	let next_validators = consensus_proof
```

modules/consensus/tendermint/verifier/src/verifier.rs
seunlanlege: I don't see where we make this check

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,321 @@
...
+
+	let verifier = SpIoVerifier::default();
+	let result = verifier.verify_update_header(
+		untrusted_block_state,
+		tendermint_trusted_state,
+		&verifier_options,
+		now,
+	);
+
+	// Validate ancestry chain before cryptographic verification
```
seunlanlege: this comment is wrong, best to remove

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,321 @@
...
+use crate::{
+	hashing::SpIoSha256, ConsensusProof, SpIoVerifier, TrustedState, UpdatedTrustedState,
+	VerificationError, VerificationOptions,
+};
+
+/// Main verification function for header updates
+pub fn verify_header_update(
+	trusted_state: TrustedState,
+	consensus_proof: ConsensusProof,
+	options: VerificationOptions,
```
seunlanlege: I think these optoins should be part of trusted state

modules/consensus/tendermint/verifier/src/verifier.rs
seunlanlege: Anyone can run a prover @royvardhan

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,321 @@
...
+
+		// Only rotate if signed header has next_validators_hash and consensus proof has
+		// next_validators
+		if !signed_header_hash.is_empty() && consensus_proof.next_validators.is_some() {
+			let consensus_proof_validators = consensus_proof.next_validators.as_ref().unwrap();
+			let validator_set = ValidatorSet::new(consensus_proof_validators.clone(), None);
+			let consensus_proof_hash = validator_set.hash_with::<SpIoSha256>();
+
+			// Only rotate if hashes match
+			if consensus_proof_hash.as_bytes() == signed_header_hash.as_bytes() {
```
seunlanlege: We should perform this check before header verification

modules/consensus/tendermint/verifier/src/verifier.rs
seunlanlege: basically look at which validator set the signed header references. They should either be the current ot next set known to the verifier. Otherwise its an invalid header. Also if `signed_header.next_validators_hash != trusted_state.next_validators_hash` we should validate that the provided `consensus_proof.next_validators.hash() == signed_header.next_validators_hash` before this block. And use that as a signal to rotate the authorities view

modules/consensus/tendermint/verifier/src/verifier.rs
royvardhan: Yeah, that makes sense. I have added the checks now.

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,355 @@
...
+/// Validates which validator set the signed header references and check next_validators_hash
+/// rotation. Returns the correct ValidatorSet for the header and validates next_validators if
+/// rotation is signaled.
+fn extract_validators<'a>(
+	trusted_state: &'a TrustedState,
+	consensus_proof: &'a ConsensusProof,
+) -> Result<ValidatorSet, VerificationError> {
+	let signed_header_validators_hash = &consensus_proof.signed_header.header.validators_hash;
+	let current_validators_hash =
+		ValidatorSet::new(trusted_state.validators.clone(), None).hash_with::<SpIoSha256>();
```
Wizdave97: Don't we still need  to employ the same hashing strategy we used in `validate_validator_set_hash` for polygon?

royvardhan: Yes you are correct.

modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -0,0 +1,355 @@
...
+		return Err(VerificationError::Invalid(format!(
+			"Unknown validator set hash: {:?}",
+			signed_header_validators_hash
+		)));
+	};
+
+	// Validate next_validators_hash rotation if signaled
+	let signed_header_next_validators_hash =
+		&consensus_proof.signed_header.header.next_validators_hash;
+	if signed_header_next_validators_hash != &next_validators_hash {
```
Wizdave97: This check should only run when `signed_header_next_validators_hash` is not empty, From the cometbft go code,  next_validator_hash is allowed to be empty.
https://github.com/cometbft/cometbft/blob/bd1bc3bb6392a973d7b60fd0b19da6e71ca09b47/types/block.go#L426

https://github.com/cometbft/cometbft/blob/bd1bc3bb6392a973d7b60fd0b19da6e71ca09b47/types/validation.go#L247

modules/consensus/tendermint/verifier/src/verifier.rs
Wizdave97: You should remove this check here since it's already implemented in `extract_validators`, rotate the authority set if it's provided without any checks.

modules/consensus/tendermint/prover/src/prover.rs
```diff
@@ -0,0 +1,59 @@
+use crate::{client::Client, error::ProverError};
+use tendermint_verifier::{ConsensusProof, TrustedState};
```
seunlanlege: Maybe introduce a primitives crate

modules/consensus/tendermint/prover/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+default = ["std"]
+std = [
+    "serde",
+    "serde_json",
+    "reqwest",
+    "base64",
+    "tokio",
+    "polkadot-sdk/std",
+]
+no-std = []
```
Wizdave97:
```suggestion
```

modules/consensus/tendermint/prover/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+secp256k1 = []
```
Wizdave97:
```suggestion
```

modules/consensus/tendermint/prover/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+prost = "0.13"
+sha3 = "0.10"
+tracing = "0.1.40"
+tendermint-verifier = { path = "../verifier" }
+
+[dependencies.polkadot-sdk]
+workspace = true
+features = ["sp-runtime", "sp-core", "sp-io"]
+
+[features]
```
Wizdave97: Prover does not need these feature flags, enable default features on all crates

modules/consensus/tendermint/prover/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,261 @@
+#[cfg(test)]
+mod tests {
+	use std::time::{SystemTime, UNIX_EPOCH};
+
+	use crate::{prove_header_update, Client, CometBFTClient, HeimdallClient};
+	use tendermint_verifier::{TrustedState, VerificationError, VerificationOptions};
+	use tracing::trace;
+
+	const STANDARD_RPC_URL: &str =
+		"https://rpc.ankr.com/sei/8c2c2b98554b32f7dd5fc9bc44a6b1da42f117c19ee97f5a55c9b74d4d33c7cd";
```
Wizdave97: Don't commit the rpc addresses, fetch them from environment variables

Wizdave97: I've added them to secrets with the following names
`STANDARD_TENDERMINT_URL`
`POLYGON_HEIMDALL`

royvardhan: Roger that!

modules/consensus/tendermint/prover/src/prover.rs
royvardhan: Done

.github/workflows/ci.yml
```diff
@@ -370,6 +378,10 @@ jobs:
               run: |
                   cargo test -p bsc-prover -- --nocapture --ignored
 
+            - name: Tendermint Polygon Integration Test
+              run: |
+                  cargo test test_polygon_heimdall_full_verification -- --nocapture --ignored
```
Wizdave97:
```suggestion
                  cargo test -p tendermint-prover -- --nocapture --ignored
```

.github/workflows/ci.yml
```diff
@@ -24,6 +29,9 @@ env:
     GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
     MAINNET_FORK_URL: ${{ secrets.MAINNET_FORK_URL }}
     TELEMETRY_SECRET_KEY: ${{ secrets.TELEMETRY_SECRET_KEY }}
+    STANDARD_RPC_URL: ${{secrets.STANDARD_TENDERMINT_URL}}
```
Wizdave97:
```suggestion
    STANDARD_TENDERMINT_URL: ${{secrets.STANDARD_TENDERMINT_URL}}
```

.github/workflows/ci.yml
```diff
@@ -24,6 +29,9 @@ env:
...
+    POLYGON_RPC_URL: ${{secrets.POLYGON_HEIMDALL}}
```
Wizdave97:
```suggestion
    POLYGON_HEIMDALL: ${{secrets.POLYGON_HEIMDALL}}
```

modules/consensus/tendermint/prover/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,267 @@
...
+		trusted_state.validate()?;
+
+		trace!(
+			"Starting verification with {} validator set transitions",
+			VALIDATOR_SET_TRANSITIONS
+		);
+		trace!("Initial trusted height: {}", trusted_state.height);
+
+		// Perform multiple validator set transitions
+		for transition in 1..=VALIDATOR_SET_TRANSITIONS {
```
Wizdave97: To test validator set transitions only count a transition when trusted_state.next_validator_hash is equal to consensus_proof.signed_header.validator_hash

modules/consensus/tendermint/prover/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	#[tokio::test]
+	#[ignore]
+	async fn test_polygon_heimdall_basic_rpc() {
+		trace!("Testing Polygon's Heimdall Fork (Basic RPC)");
+		test_polygon_basic_rpc(&get_polygon_rpc_url()).await.unwrap();
+	}
+
+	// Fails with Invalid Proof, debugging
```
seunlanlege: This still true?

royvardhan: nope, forgot to remove it.

.github/workflows/ci.yml
```diff
@@ -1,12 +1,17 @@
...
+
+    pull_request:
         branches:
             - main
 
-    pull_request_target:
-        types: [opened, synchronize]
+
+    # pull_request_target:
+    #     types: [opened, synchronize]
```
seunlanlege: we have to revert this

royvardhan: Yes yes once ci is green.

seunlanlege: we're green

royvardhan: 🟢

Wizdave97 (approved):

PR #458, opened 2025-07-09
Timestamp digest introduction

### LINKED ISSUE #389: Notes on changing `SLOT_DURATION`
Hyperbridge nexus currently uses a slot duration of 12 seconds on mainnet. If for whatever reason, this needs to change, whether to support elastic scaling. It's important that connected chains do not learn of an incorrect state commitment timestamp, more critically a timestamp that is incorrectly greater than hyperbridge's timestamp as this enables timeouts for arbitrary requests, even delivered ones.

So updating the `SLOT_DURATION` requires some coordinated effort. Solochains and parachains must disconnect from hyperbridge before it performs it's slot duration change. Only reconnecting with the updated slot duration after the change is completed. Evm chains must be configured with a new consensus client ahead of the change, so they never calculate the incorrect timestamp

dharjeezy:
This PR introduces timestamp directly into a `PreRuntime` digest item in every block header for Hyperbridge.
 closes #389

### DIFF
--- a/modules/pallets/ismp/src/utils.rs
+++ b/modules/pallets/ismp/src/utils.rs
@@ -151,6 +151,6 @@ pub const ISMP_TIMESTAMP_ID: sp_runtime::ConsensusEngineId = *b"ISTM";
 /// Timestamp log digest for pallet ismp
 #[derive(Encode, Decode, Clone, scale_info::TypeInfo, Default)]
 pub struct TimestampDigest {
-	/// Timestamp digest	Timestamp value in seconds
+	/// Timestamp value in seconds
 	pub timestamp: u64,
 }


modules/pallets/ismp/src/utils.rs
```diff
@@ -144,3 +144,13 @@ pub struct ConsensusDigest {
...
 	pub child_trie_root: H256,
 }
+
+/// The `ConsensusEngineId` of Ismp `TimestampDigest` in the parachain header.
+pub const ISMP_TIMESTAMP_ID: sp_runtime::ConsensusEngineId = *b"ISTM";
+
+/// Timestamp log digest for pallet ismp
+#[derive(Encode, Decode, Clone, scale_info::TypeInfo, Default)]
+pub struct TimestampDigest {
+	/// Timestamp digest
```
seunlanlege:
```suggestion
	/// Timestamp value in seconds
```

modules/ismp/clients/grandpa/src/lib.rs
```diff
@@ -101,7 +101,7 @@ pub mod pallet {
 	impl<T: Config> Pallet<T> {
 		/// Add some a state machine to the list of supported state machines
 		#[pallet::call_index(0)]
-		#[pallet::weight(T::WeightInfo::add_state_machines(new_state_machines.len() as u32))]
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::add_state_machines(new_state_machines.len() as u32))]
```
seunlanlege: lets revert all these fully qualified trait method calls

modules/pallets/ismp/src/lib.rs
```diff
@@ -280,9 +280,19 @@ pub mod pallet {
...
 
 			let log = ConsensusDigest { child_trie_root, mmr_root: root.into() };
-
 			let digest = sp_runtime::generic::DigestItem::Consensus(ISMP_ID, log.encode());
 			<frame_system::Pallet<T>>::deposit_log(digest);
+
+			let timestamp_secs = T::TimestampProvider::now()
+				.as_millis()
+				.saturated_into::<u64>()
+				.saturating_div(1000);
```
seunlanlege:
```suggestion
				.as_secs();
```

modules/pallets/ismp/src/utils.rs
```diff
@@ -144,3 +144,13 @@ pub struct ConsensusDigest {
...
 	pub child_trie_root: H256,
 }
+
+/// The `ConsensusEngineId` of Ismp `TimestampDigest` in the parachain header.
+pub const ISMP_TIMESTAMP_ID: sp_runtime::ConsensusEngineId = *b"ISTM";
+
+/// Timestamp log digest for pallet ismp
+#[derive(Encode, Decode, Clone, scale_info::TypeInfo, Default)]
+pub struct TimestampDigest {
+	/// Timestamp digest	Timestamp value in seconds
```
seunlanlege:
```suggestion
	/// Timestamp value in seconds
```

PR #483, opened 2025-08-18
Help to fix XCM testing

bkontur:
Fixes:
- the messaging from ParaA to ParaB was not working - ParaA just executed `TransferReserveAsset` but nothing was executed on ParaB (fix around `MsgQueue` and  `ParachainXcmRouter<crate::runtime::MsgQueue>;`).
- invalid `TestReseve` filter, AssetHub/ParaA has `Location::new(1, Parachain(1000))` and not `Location::new(1, Here)`

### DIFF
--- a/modules/pallets/testsuite/src/xcm.rs
+++ b/modules/pallets/testsuite/src/xcm.rs
@@ -293,7 +293,7 @@ pub struct TestReserve;
 impl ContainsPair<Asset, Location> for TestReserve {
 	fn contains(asset: &Asset, origin: &Location) -> bool {
 		println!("TestReserve::contains asset: {asset:?}, origin:{origin:?}");
-		let assethub_location = Location::new(1, Here);
+		let assethub_location = Location::new(1, Parachain(1000));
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -101,8 +101,8 @@ members = [
     "tesseract/consensus/sync-committee",
     "tesseract/consensus/grandpa",
     "tesseract/consensus/integration-tests",
-    "tesseract/consensus/beefy",
-    "tesseract/consensus/beefy/zk",
+#    "tesseract/consensus/beefy",
+#    "tesseract/consensus/beefy/zk",
```
bkontur, line 105: @seunlanlege @dharjeezy Yes, sorry - this was temporary, just to pass test compilation without your private repo. I shouldn’t have pushed it. Also, I didn’t expect it to be merged - more like just to show the diff with the fix, but no problem :)

dharjeezy: i will revert on this on my branch. Thank you

PR #498, opened 2025-08-27
register asset locally without dispatching any request

### LINKED ISSUE #495: Allow registering assets locally in tokengateway
Add a call to token gateway that registers an asset locally without dispatching a request to hyperbridge.

dharjeezy:
closes #495

### DIFF
--- a/modules/ismp/core/src/handlers/response.rs
+++ b/modules/ismp/core/src/handlers/response.rs
@@ -24,7 +24,7 @@ use crate::{
 	router::{GetResponse, Request, RequestResponse, Response, StorageValue},
 };
 use alloc::{vec, vec::Vec};
-use codec::{Decode, Encode};
+use codec::Decode;
 use crypto_utils::verification::Signature;
 use sp_weights::Weight;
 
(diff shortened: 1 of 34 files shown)


modules/pallets/token-gateway/src/lib.rs
```diff
@@ -503,6 +510,33 @@ pub mod pallet {
 			}
 			Ok(())
 		}
+
+		/// Registers a multi-chain ERC6160 asset without sending any dispatch request.
+		/// The asset should not already exist.
```
Wizdave97:
```suggestion
```

modules/ismp/core/src/handlers/response.rs
```diff
@@ -33,6 +35,12 @@ where
 {
 	let signer = msg.signer.clone();
 
+	let data = sp_io::hashing::keccak_256(&msg.datagram.encode());
+	let relayer_public_key = Signature::decode(&mut signer.as_slice())
```
Wizdave97: Don't verify the signature, implement a method on the Signature object that just returns public key or address from the enum and use that as the signer everywhere msg.signer is  used.

PR #501, opened 2025-08-29
get latest block hash instead of using subxt's at_latest

dharjeezy:


modules/utils/subxt/src/lib.rs
```diff
@@ -215,3 +216,11 @@ pub fn fisherman_storage_key(address: Vec<u8>) -> Vec<u8> {
...
 	[pallet_prefix, storage_prefix, key_1, address.encode()].concat()
 }
+
+pub async fn get_latest_block_hash<C: subxt::Config>(
+	rpc: &LegacyRpcMethods<C>,
+) -> Result<HashFor<C>, anyhow::Error> {
+	rpc.chain_get_block_hash(None)
+		.await?
+		.ok_or_else(|| anyhow!("Failed to query latest block hash"))
+}
```
seunlanlege: I don't know if we need this tbh

Wizdave97: Are you talking about the function just to this?

Wizdave97: @dharjeezy, just use that line directly where it's needed instead of creating this function.
Also in this function https://github.com/polytope-labs/hyperbridge/blob/0f37db9c89cd331a3608566d7f7ef02352338320/tesseract/messaging/messaging/src/lib.rs#L443

Filter out receipts where the request_fee_metadata  or response_fee_metadata on the source chain is zero, so we don't store unnecessary requests in the db.

tesseract/messaging/messaging/src/lib.rs
```diff
@@ -439,7 +440,33 @@ async fn fee_accumulation<A: IsmpProvider + Clone + Clone + HyperbridgeClaim + '
...
+				let fee = match receipt {
+					TxReceipt::Request { query, .. } =>
+						source_chain.query_request_fee_metadata(query.commitment).await,
+					TxReceipt::Response { query, .. } =>
+						source_chain.query_response_fee_metadata(query.commitment).await,
+				};
+
+				match fee {
+					Ok(fee_amount) if fee_amount > U256::zero() => Some(receipt),
+					_ => None,
```
Wizdave97: If the query returned an error, spawn a background task to store the receipt in the DB, instead of dropping it.

Wizdave97: Also in `create_claim_proof` function on the `TransactionPayment` struct filter out requests and responses with zero fees, then delete them in the background.

tesseract/messaging/fees/src/lib.rs
```diff
@@ -451,45 +453,127 @@ impl TransactionPayment {
...
+
+					match source.query_request_fee_metadata(hash).await {
+						Ok(fee) if fee.is_zero() => Some(Err(Key::Request(hash))),
+						Ok(_) => {
+							let source_key = source.request_commitment_full_key(hash);
+							//Get request receipt keys on dest chain
+							let dest_key = dest.request_receipt_full_key(hash);
+							Some(Ok((Key::Request(hash), source_key, dest_key)))
+						}
+						Err(err) => {
```
Wizdave97: This implementation is a bit unreadable. I think it's better if you just query the fee metadata inside this block and push keys with zero fees into the `keys_to_delete` array
https://github.com/polytope-labs/hyperbridge/blob/0f37db9c89cd331a3608566d7f7ef02352338320/tesseract/messaging/fees/src/lib.rs#L477

tesseract/messaging/fees/src/lib.rs
```diff
@@ -451,45 +453,127 @@ impl TransactionPayment {
...
+			.chain(resp_errors.into_iter().filter_map(|e| e.err()))
+			.collect();
+		let responses: Vec<_> = responses.into_iter().filter_map(|e| e.ok()).collect();
 
 		let mut keys_to_delete = vec![];
 		let mut keys_to_prove = vec![];
 
-		for key in requests.chain(responses) {
+		for key in requests.into_iter().chain(responses) {
 			if hyperbridge.check_claimed(key.0.clone()).await? {
```
Wizdave97, line 490: Yeah, just query fee metadata here and push keys with zero fee metadata into the keys to be deleted

Wizdave97 (changes requested):
One more thing,  make the signer field required in the substrate chain config in tesseract

Wizdave97:
> One more thing, make the signer field required in the substrate chain config in tesseract

Ignore this, we'll do it in a different PR.

tesseract/messaging/fees/src/lib.rs
```diff
@@ -475,19 +477,37 @@ impl TransactionPayment {
 		let mut keys_to_prove = vec![];
 
 		for key in requests.chain(responses) {
+			let has_fee = match &key.0 {
+				Key::Request(hash) => match source.query_request_fee_metadata(*hash).await {
```
Wizdave97: There's no need to match the result, this function can return an error if the request fails.

tesseract/messaging/fees/src/lib.rs
```diff
@@ -475,19 +477,37 @@ impl TransactionPayment {
...
+					Ok(fee) => !fee.is_zero(),
+					Err(_) => true,
+				},
+				Key::Response { request_commitment, .. } => {
+					match source.query_response_fee_metadata(*request_commitment).await {
```
Wizdave97: You are meant to use response commitment here
```suggestion
					match source.query_response_fee_metadata(*response_commitment).await {
```

Wizdave97 (approved):

PR #506, opened 2025-09-01
Allow for Delivery address Relayer Signature decoding and Backward compatibility

### LINKED ISSUE #504: Backwards compatibility for regular signers when accumulating fees
Since the hyperbridge relayer currently submits messages to substrate chains with the signer as an encoded `Signature` object, this causes relayer fees to be accumulated into the wrong address for deliveries to substrate chains.

Add a check that decodes the `delivery_address` into a Signature object only if it has a length greater than 32 bytes.
This way, we would support substrate chains that are not on the latest release of pallet-ismp.

This should be done after proof verification and before storing accumulated fees in `pallet-ismp-relayer`

dharjeezy:
closes #504

### DIFF
--- a/parachain/runtimes/nexus/src/lib.rs
+++ b/parachain/runtimes/nexus/src/lib.rs
@@ -226,7 +226,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("nexus"),
 	impl_name: Cow::Borrowed("nexus"),
 	authoring_version: 1,
-	spec_version: 3_800,
+	spec_version: 3_900,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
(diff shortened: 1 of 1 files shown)


modules/pallets/relayer/src/lib.rs
```diff
@@ -683,8 +685,19 @@ where
...
 						s if s.is_substrate() => {
 							use codec::Decode;
-							<Vec<u8>>::decode(&mut &*encoded_receipt)
-								.map_err(|_| Error::<T>::ProofValidationError)?
+							let relayer_bytes = <Vec<u8>>::decode(&mut &*encoded_receipt)
+								.map_err(|_| Error::<T>::ProofValidationError)?;
+							if relayer_bytes.len() > 32 {
+								let signature = Signature::decode(&mut &*relayer_bytes)
+									.map_err(|_| Error::<T>::SignatureDecodingError)?;
+								match signature {
```
Wizdave97: You can just call the signer method instead of matching

Wizdave97 (approved):

PR #514, opened 2025-09-09
Allowance for vesting/locked tokens for Candidacy bonding

dharjeezy:
allow bonding of locked/vesting tokens for Collator candidacy bonding.

### DIFF
--- a/modules/pallets/collator-manager/src/lib.rs
+++ b/modules/pallets/collator-manager/src/lib.rs
@@ -278,18 +278,13 @@ pub mod pallet {
 			who: &T::AccountId,
 			value: <T as pallet::Config>::Balance,
 		) -> (NegativeImbalanceOf<T>, <T as pallet::Config>::Balance) {
-			Bonded::<T>::mutate(who, |bonded| {
+			let amount = Bonded::<T>::mutate(who, |bonded| {
 				let to_slash = (*bonded).min(value);
 				*bonded = bonded.saturating_sub(to_slash);
 				to_slash
 			});
 
-			T::NativeCurrency::set_lock(
-				T::LockId::get(),
-				who,
-				Bonded::<T>::get(who),
-				WithdrawReasons::all(),
-			);
+			T::NativeCurrency::set_lock(T::LockId::get(), who, amount, WithdrawReasons::all());
 
 			let (imbalance, remainder) = T::NativeCurrency::slash(who, value);
 			(imbalance, remainder)


modules/pallets/collator-balances/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+		fn make_free_balance_be(
+			who: &T::AccountId,
+			balance: Self::Balance,
+		) -> SignedImbalance<Self::Balance, Self::PositiveImbalance> {
+			T::NativeCurrency::make_free_balance_be(who, balance)
+		}
+	}
+
+	/// The custom implementation of the `ReservableCurrency` trait.
+	impl<T: Config> ReservableCurrency<T::AccountId> for Pallet<T> {
```
Wizdave97: Is this how `pallet-conviction-voting` handles locks on balances?

dharjeezy: it is quite similar, conviction-voting and this pallet uses lock to to hold user's fund to make it non transferable, the difference is ours is a wrapper that translates the reserve, unreserve function into a lock hence why ReservableCurrency trait is needed to be implemented specifically.

modules/pallets/collator-manager/src/lib.rs
```diff
@@ -154,4 +208,189 @@ pub mod pallet {
...
+				.saturating_sub(T::NativeCurrency::reserved_balance(who));
+			unreserved_balance >= value
+		}
+
+		/// Reserves by placing a lock on the `NativeCurrency`.
+		/// Checks if the funds are available via `can_reserve`. If so, it updates the
+		/// `Bonded` storage and then places a lock on `pallet-balances` for the new total
+		/// bonded amount.
+		fn reserve(who: &T::AccountId, value: <T as pallet::Config>::Balance) -> DispatchResult {
+			dbg!("reserving for {:?} {:?}", who, value);
```
Wizdave97:
```suggestion
```

modules/pallets/collator-manager/src/lib.rs
```diff
@@ -154,4 +208,188 @@ pub mod pallet {
...
+		/// entirely.
+		fn unreserve(
+			who: &T::AccountId,
+			value: <T as pallet::Config>::Balance,
+		) -> <T as pallet::Config>::Balance {
+			let unreserved_amount = Bonded::<T>::mutate(who, |bonded| {
+				let to_unreserve = (*bonded).min(value);
+				*bonded = bonded.saturating_sub(to_unreserve);
+				to_unreserve
+			});
```
seunlanlege, line 256: I think the better behaviour is to throw an error if the amount to unreserve > bonded rather than swallowing this error

modules/pallets/collator-manager/src/lib.rs
```diff
@@ -154,4 +208,188 @@ pub mod pallet {
...
+			Bonded::<T>::mutate(who, |bonded| {
+				let to_slash = (*bonded).min(value);
+				*bonded = bonded.saturating_sub(to_slash);
+				to_slash
+			});
+
+			T::NativeCurrency::set_lock(
+				T::LockId::get(),
+				who,
+				Bonded::<T>::get(who),
```
seunlanlege: [`StorageMap::mutate`](https://docs.rs/frame-support/latest/frame_support/storage/trait.StorageMap.html#tymethod.mutate) returns the new vaule, so we can save ourselves a query here

modules/pallets/collator-manager/src/lib.rs
dharjeezy: but the api return type is not expecting a Result but an actual balance

PR #519, opened 2025-09-12
Create a call for withdrawing protocol fees from EVM

Wizdave97:


modules/pallets/relayer/src/lib.rs
```diff
@@ -211,6 +213,55 @@ pub mod pallet {
 			MinimumWithdrawalAmount::<T>::insert(state_machine, U256::from(amount));
 			Ok(())
 		}
+
+		/// Issues a call to withdraw protocol the fees from an evm chain
```
dharjeezy:
```suggestion
		/// Issues a call to withdraw protocol fees from an evm chain
```

PR #523, opened 2025-09-16
collect fees for non-incentivized chains

dharjeezy:


modules/pallets/messaging-fees/src/impls.rs
```diff
@@ -145,115 +145,119 @@ where
...
+						.ok_or(Error::<T>::CalculationOverflow)?;
+
+					let base_reward_12_decimals: u128 = cost_bridge_price_12_decimals_u256
+						.try_into()
+						.map_err(|_| Error::<T>::CalculationOverflow)?;
+
+					let base_reward_as_balance: T::Balance =
+						base_reward_12_decimals.saturated_into();
+
+					if u128::from(base_reward_as_balance) == 0 {
```
Wizdave97: Don't think we need this check again since we have the check for cost above

modules/pallets/messaging-fees/src/impls.rs
```diff
@@ -145,115 +145,115 @@ where
...
-
-					Self::deposit_event(Event::FeePaid {
-						relayer: relayer.clone(),
-						amount: base_reward_as_balance,
-					});
-					T::ReputationAsset::mint_into(&relayer, base_reward_as_balance)
-						.map_err(|_| Error::<T>::ReputationMintFailed)?;
+				};
+
+				if bytes_processed == 0 {
```
Wizdave97: The protocol requires a minimum of 32 bytes for a message. 
So let's remove this if check and add `bytes_processed = core::cmp::max(bytes_processed, 32);`

Wizdave97 (approved):

PR #533, opened 2025-09-22
fix: use pagination while getting polygon validators

royvardhan:


tesseract/consensus/polygon/src/client.rs
```diff
@@ -462,11 +462,54 @@ impl Client for HeimdallClient {
...
+		let mut total_validators = "0".to_string(); // Store the actual total from first response
+
+		loop {
+			let heimdall_response: HeimdallValidatorsResponse = self
+				.rpc_request(
+					"validators",
+					json!({
+						"height": height.to_string(),
+						"page": page.to_string(),
+						"per_page": page_size.to_string()
```
Wizdave97, line 480: What's the maximum page size the RPC endpoint allows?

royvardhan: It is [100](https://github.com/0xPolygon/cometbft/blob/72fcfe58cec621e3bfa0acb69d368055a7687ced/light/rpc/client.go#L635)

Wizdave97: Okay

tesseract/consensus/polygon/src/client.rs
```diff
@@ -462,11 +462,54 @@ impl Client for HeimdallClient {
...
+			all_validators.extend(heimdall_response.clone().validators);
+
+			if heimdall_response.validators.len() < page_size {
+				break;
+			}
+
+			page += 1;
+
+			// Safety check to prevent infinite loops
+			if page > 10 {
```
Wizdave97: Why not break only when `all_validators.len()  == total_validators` or page_size > 10?

tesseract/consensus/polygon/src/client.rs
```diff
@@ -462,11 +462,54 @@ impl Client for HeimdallClient {
...
-		// Use Heimdall-specific response type
-		let heimdall_response: HeimdallValidatorsResponse =
-			self.rpc_request("validators", json!({"height": height.to_string()})).await?;
+		let mut all_validators = Vec::new();
+		let mut page = 1;
+		let page_size = 100;
+		let mut total_validators = "0".to_string(); // Store the actual total from first response
+
+		loop {
+			let heimdall_response: HeimdallValidatorsResponse = self
```
Wizdave97, line 474: Add some trace logs for run of the loop

.github/workflows/ci.yml
```diff
@@ -373,6 +373,11 @@ jobs:
             #
             #          cargo test -p polygon-pos-prover -- --nocapture --ignored
 
+
+            - name: polygon pos integration tests
+              run: |
+                  RUST_LOG=tesseract=trace, cargo test -p tesseract-integration-tests -- --nocapture
```
Wizdave97: Just add a polygon mainnet test to tendermint prover, tesseract tests will not work in CI

royvardhan: In that case, the current test for prover is all we need with mainnet envs. Which you have already set. I have now removed the ci changes.

PR #534, opened 2025-09-22
count all bytes processed

dharjeezy:


seunlanlege:
Don't merge this yet until we alert the relevant stake holders

PR #542, opened 2025-10-01
Fetch mmr leaves concurrently in beefy prover

Wizdave97:


modules/consensus/beefy/prover/src/lib.rs
```diff
@@ -174,8 +182,13 @@ impl<R: Config, P: Config> Prover<R, P> {
 			.await?
 			.ok_or_else(|| anyhow!("Failed to query blockhash for blocknumber"))?;
 
+		log::trace!("\n\nQuerying mmr proof\n\n");
```
seunlanlege: we can probably remove these log lines

PR #545, opened 2025-10-03
check for minimum balance before attempting to transfer

dharjeezy:
modify how we check target message size
change target message size based on 0.01 per byte fee

### DIFF
--- a/modules/pallets/messaging-fees/src/impls.rs
+++ b/modules/pallets/messaging-fees/src/impls.rs
@@ -3,7 +3,10 @@ use alloc::{collections::BTreeMap, string::ToString, vec};
 use codec::{Decode, Encode};
 use hyperbridge_client_machine::OnRequestProcessed;
 use polkadot_sdk::{
-	frame_support::traits::{fungible::Mutate, tokens::Preservation},
+	frame_support::traits::{
+		fungible::{Inspect, Mutate},
+		tokens::Preservation,
+	},
 	sp_core::U256,
 	sp_runtime::traits::*,
 };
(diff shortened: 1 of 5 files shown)


modules/pallets/messaging-fees/src/impls.rs
```diff
@@ -225,27 +228,30 @@ where
...
+								)
+								.map_err(|_| Error::<T>::RewardTransferFailed)?;
+
+								Self::deposit_event(Event::FeeRewarded {
+									relayer: relayer.clone(),
+									amount: reward_amount.saturated_into(),
+								});
+							} else {
+								log::info!(target: "ismp", "Reward amount {reward_amount:?} below minimum balance");
+							}
```
seunlanlege, line 254: uhm maybe set the minimum balance to zero? otherwise people will miss rewards

parachain/runtimes/nexus/src/lib.rs
```diff
@@ -875,8 +875,8 @@ impl pallet_messaging_fees::Config for Runtime {
 	>;
 	type PriceOracle = FixedPriceOracle;
 
-	/// Target message size: 6.51 KB (6.51 * 1024 = 6666 bytes)
-	type TargetMessageSize = ConstU32<6666>;
+	/// Target message size: 195.3125 KB (195.3125 * 1024 = 200000 bytes)
+	type TargetMessageSize = ConstU32<200000>;
```
seunlanlege: Maybe remove entirely?

modules/pallets/messaging-fees/src/impls.rs
Wizdave97: Hmmn, how about state bloat from empty accounts?

seunlanlege: Or make the minimum balance absurdly small. eg 0.0000001 $BRIDGE

Wizdave97: Existential deposit has been changed, but keeping this just for the edge case where reward is zero

Wizdave97 (approved):

PR #554, opened 2025-10-10
Prepare tendermint state machine for deployment

Wizdave97:
Adds Tendermint ISMP state machine to the runtime
Some refactor the config for Tendermint consensus and messaging relayer.

### DIFF
--- a/tesseract/consensus/relayer/Cargo.toml
+++ b/tesseract/consensus/relayer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tesseract-consensus"
-version = "1.3.1"
+version = "1.4.0"
 edition = "2021"
 
 [lib]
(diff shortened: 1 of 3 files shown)


tesseract/consensus/relayer/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tesseract-consensus"
-version = "1.3.0"
+version = "1.3.1"
```
seunlanlege: Minor bump seems more appropriate

```suggestion
version = "1.4.0"
```

tesseract/messaging/relayer/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tesseract"
-version = "1.2.1"
+version = "1.2.2"
```
seunlanlege:
```suggestion
version = "1.4.0"
```

PR #564, opened 2025-10-21
Support fulu hardfork in sync committee

Wizdave97:
Extends the beacon state with a new field added in the fusaka fork


seunlanlege (commented):
Looks good, can you revert the changes to the abi generated files?

PR #581, opened 2025-11-10
Remove Fee Decimals Scaling and Introduction of Optional Beneficiary Address for Relayer when Withdrawing Fees

### LINKED ISSUE #579: Add optional beneficiary to relayer fee withdrawal parameters
In pallet messaging relayer fees when we accumulate the protocol fees into the relayer's account, we collect it into the relayer's substrate account, because that's the only account we have access to, our current withdrawal implementation in Tesseract cannot handle this case; hence, relayers will have to withdraw this manually, but they need to provide a beneficiary(an evm address) that will receive this withdrawal on the evm chain, current withdrawal call does not support this.

### LINKED ISSUE #580: Wrong use of fee token decimals in `pallet-messaging-fees`
The fee token decimal value is being misused in `pallet-messaging-fees` in get_per_byte_fee. It should be used directly without any scaling.

dharjeezy:
This removes the scaling of fee token decimals when calculating `per_byte_fee` in `pallet-messaging-fees`. This closes #580 

This PR also introduce an optional beneficiary address when withdrawing fees, which closes #579

### DIFF
--- a/parachain/simtests/src/migration_test.rs
+++ b/parachain/simtests/src/migration_test.rs
@@ -148,8 +148,8 @@ async fn wait_for_port(port: u16, timeout: Duration) -> Result<(), anyhow::Error
 	Err(anyhow!("Timed out waiting for connection to port"))
 }
 
-#[tokio::test]
-#[ignore]
+// #[tokio::test]
+// #[ignore]
 async fn test_runtime_upgrade_and_fee_migration() -> Result<(), anyhow::Error> {
 	// Check nexus runtime version - only run test if version is below 6000
 	println!("Connecting to Nexus at: {} to check runtime version...", NEXUS_RPC);


tesseract/messaging/substrate/src/lib.rs
```diff
@@ -61,6 +61,8 @@ pub struct SubstrateConfig {
 	pub max_rpc_payload_size: Option<u32>,
 	/// Relayer account seed
 	pub signer: String,
+	/// Optional withdrawal beneficiary address
```
Wizdave97: There's no no need to modify anything in tesserract, just use None when it's trying to withdraw

modules/pallets/relayer/src/lib.rs
Wizdave97: The message here should contain the beneficiary address if it's available

tesseract/messaging/evm/src/lib.rs
```diff
@@ -68,6 +68,8 @@ pub struct EvmConfig {
 	pub ismp_host: H160,
 	/// Relayer account private key
 	pub signer: String,
+	/// Optional Beneficiary withdrawal address
+	pub withdrawal_beneficiary_address: Option<String>,
```
Wizdave97: Not needed

tesseract/messaging/substrate/src/calls.rs
```diff
@@ -177,7 +177,11 @@ where
 			counterparty.sign(&message)
 		};
 
-		let input_data = WithdrawalInputData { signature, dest_chain: chain };
+		let input_data = WithdrawalInputData {
+			signature,
+			dest_chain: chain,
+			beneficiary: counterparty.withdrawal_beneficiary_address(),
```
Wizdave97:
```suggestion
			beneficiary: None
```

modules/pallets/messaging-fees/src/impls.rs
```diff
@@ -117,8 +117,7 @@ where
 				else {
 					return None;
 				};
-				let scaling_power = 18u8.saturating_sub(decimals); // assumption that the decimals will always be less than 18
-				let scaling_factor = U256::from(10u128.pow(scaling_power as u32));
+				let scaling_factor = U256::from(10u128.pow(decimals as u32));
```
Wizdave97: The per_byte_fee is already in the correct decimals for the source chain, don't even think we need this

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,132 @@
...
+                weight.saturating_accrue(weight_per_item);
+                items_processed.saturating_inc();
+
+                let mut state_machine_key_part = &key_suffix[16..];
+
+                log::info!(target: "ismp", "MIGRATION: Attempting to decode key part: {:?}", &state_machine_key_part);
+
+                if let Ok(state_machine) = StateMachine::decode(&mut state_machine_key_part) {
+
+                    if state_machine.is_substrate() {
```
Wizdave97: The state machine is not substrate, we want to check that the state machine is evm and the address is 32 bytes

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,132 @@
...
+                    Box::new(KeyPrefixIterator::new(
+                        fee_storage_prefix.to_vec(),
+                        fee_storage_prefix.to_vec(),
+                        |k| Ok(k.to_vec()),
+                    ))
+                };
+
+            let mut items_processed = 0u32;
+            let max_weight = T::BlockWeights::get().max_block;
+            let weight_per_item = Weight::from_parts(200_000_000, 0);
```
Wizdave97: We should benchmark the weight of iterating through one item and use that as the base value  instead of randomly picking a value

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,132 @@
...
+            weight.saturating_accrue(T::DbWeight::get().reads(1));
+
+            while let Some(key_suffix) = iter.next() {
+                let full_key = [fee_storage_prefix.as_slice(), key_suffix.as_slice()].concat();
+
+                if weight.saturating_add(weight_per_item).ref_time() >= max_weight.ref_time() {
```
Wizdave97: We should not use more than half of the block ref_time

modules/pallets/testsuite/src/tests/pallet_messaging_fees.rs
```diff
@@ -451,3 +457,55 @@ fn test_protocol_fee_accumulation() {
...
+			);
+
+			assert_eq!(
+				pallet_ismp_relayer::Fees::<Test>::get(&evm_chain, &relayer_address),
+				initial_fee
+			);
+
+			assert_eq!(StorageVersion::get::<pallet_messaging_fees::Pallet<Test>>(), 1);
+
+			assert_eq!(MigrationInProgress::<Test>::get(), false);
```
Wizdave97: The test should run over multiple blocks also

modules/pallets/relayer/src/lib.rs
```diff
@@ -809,3 +824,11 @@ impl<T: Config> Pallet<T> {
 pub fn message(nonce: u64, dest_chain: StateMachine) -> [u8; 32] {
```
Wizdave97: This can just take an optional beneficiary instead of having a new function

tesseract/messaging/substrate/src/calls.rs
```diff
@@ -162,6 +159,26 @@ where
 		counterparty: Arc<dyn IsmpProvider>,
 		chain: StateMachine,
 	) -> anyhow::Result<WithdrawFundsResult> {
+		let hyperbridge_account_balance =
+			relayer_account_balance(&self.client, &self.rpc, chain, self.address.clone()).await?;
+
+		if hyperbridge_account_balance > U256::zero() {
```
Wizdave97: This has to be greater than $10 before you can withdraw it

tesseract/messaging/substrate/src/calls.rs
```diff
@@ -162,6 +159,26 @@ where
...
+				.chain_get_block_hash(None)
+				.await?
+				.ok_or_else(|| anyhow!("Failed to query latest block hash"))?;
+			let raw_value = self.client.storage().at(block_hash).fetch_raw(key.clone()).await?;
+			let nonce =
+				if let Some(raw_value) = raw_value { Decode::decode(&mut &*raw_value)? } else { 0u64 };
+
+			let message = message_with_beneficiary(nonce, chain, &counterparty.address());
+
+			execute_withdrawal(self, None, message, counterparty.clone(), chain).await?;
```
Wizdave97: What happens to the result of this withdrawal?

modules/pallets/messaging-fees/src/impls.rs
Wizdave97: Thinking about this, we do this scaling because of the bridge token incentives calculation, but it should only be used for that; we should use the per-byte fee directly when accumulating fees

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,132 @@
...
+
+                let mut state_machine_key_part = &key_suffix[16..];
+
+                log::info!(target: "ismp", "MIGRATION: Attempting to decode key part: {:?}", &state_machine_key_part);
+
+                if let Ok(state_machine) = StateMachine::decode(&mut state_machine_key_part) {
+
+                    if state_machine.is_substrate() {
+                        log::info!(target: "ismp", "MIGRATION: Is Substrate. Resetting fee for key {:?}", &full_key);
+                        storage::unhashed::put(&full_key, &U256::zero());
```
Wizdave97: Since we know that the accumulated fees is in 18 decimals, we should scale the value there down to the correct fee token decimals for that chain instead of resetting it to zero

modules/pallets/messaging-fees/src/impls.rs
Wizdave97, line 101: This fucntion should return the unscaled per byte fee and the per byte fee scaled to 18 decimals so we can choose which to use downstream.

modules/pallets/relayer/src/lib.rs
```diff
@@ -267,37 +269,47 @@ where
 {
 	pub fn withdraw(withdrawal_data: WithdrawalInputData) -> DispatchResult {
 		let address = match &withdrawal_data.signature {
+			Signature::Evm { address, .. } => address.clone(),
+			Signature::Sr25519 { public_key, .. } => public_key.clone(),
+			Signature::Ed25519 { public_key, .. } => public_key.clone()
+		};
+
+		let msg = match &withdrawal_data.beneficiary {
+			Some(beneficiary) => {
```
Wizdave97: There's no need for this match statement, you can just do `msg = message(nonce, withdrawal_data.dest_chain, withdrawal_data.beneficiary)`

modules/pallets/testsuite/src/tests/pallet_messaging_fees.rs
```diff
@@ -451,3 +460,138 @@ fn test_protocol_fee_accumulation() {
...
+
+			assert_eq!(
+				pallet_ismp_relayer::Fees::<Test>::get(&evm_chain, &relayer_addr_20),
+				initial_fee
+			);
+		});
+	}
+
+	#[test]
+	fn migration_scales_evm_fees_for_32_byte_address_multi_block() {
```
Wizdave97, line 471: A multiblock test is sufficient, we don't need the other test.

tesseract/messaging/relayer/src/fees.rs
```diff
@@ -393,16 +393,21 @@ where
 async fn deliver_post_request<D: IsmpProvider>(
 	dest_chain: Arc<dyn IsmpProvider>,
 	hyperbridge: &D,
-	result: WithdrawFundsResult,
+	results: Vec<WithdrawFundsResult>,
 ) -> anyhow::Result<()> {
+	if results.is_empty() {
+		return Ok(());
+	}
+	let max_block = results.iter().map(|r| r.block).max().unwrap_or(0);
```
Wizdave97:
```suggestion
	let max_block = results.iter().map(|r| r.block).max().expect("Cannot be empty, checked in previous  block");
```

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,141 @@
...
+			};
+
+			if let Some(key_suffix) = iter.next() {
+				meter.consume(weight_per_item);
+				let full_key = [fee_storage_prefix.as_slice(), key_suffix.as_slice()].concat();
+				let mut key_part = &key_suffix[16..];
+
+				if let Ok(state_machine) = StateMachine::decode(&mut key_part) {
+					if state_machine.is_evm() {
+						if key_part.len() > 16 {
```
Wizdave97: These two if block can be collapsed into one

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,141 @@
...
+				return Err(SteppedMigrationError::InsufficientWeight { required: weight_per_item });
+			}
+
+			let fee_storage_prefix = storage_prefix(b"Relayer", b"Fees");
+
+			let mut iter: Box<dyn Iterator<Item = Vec<u8>>> = if let Some(c) = cursor.as_ref() {
+				let cursor_vec = c.as_slice().to_vec();
+				Box::new(
+					KeyPrefixIterator::new(
+						fee_storage_prefix.to_vec(),
```
Wizdave97: Why is this value used twice here?

Wizdave97: I believe the second  item should be the cursor when it exists

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,141 @@
...
+						fee_storage_prefix.to_vec(),
+						|k| Ok(k.to_vec()),
```
Wizdave97:
```suggestion
						|k| <Vec<u8>>::decode(&mut &*k),
```

modules/pallets/messaging-fees/src/migrations.rs
```diff
@@ -0,0 +1,141 @@
...
+					)
+					.skip_while(move |k_vec| k_vec != &cursor_vec)
```
Wizdave97: We should not need this skip_while

tesseract/messaging/substrate/src/calls.rs
```diff
@@ -161,7 +161,36 @@ where
...
+			let nonce = if let Some(raw_value) = raw_value {
+				Decode::decode(&mut &*raw_value)?
+			} else {
+				0u64
+			};
+
+			let message = message(nonce, chain, Some(counterparty.address().clone()));
+
+			results
+				.push(execute_withdrawal(self, None, message, counterparty.clone(), chain).await?);
```
Wizdave97: beneficiary should not be None in this case

Wizdave97 (commented):
Looks good, let's add a simnode test for the migration by forking the live nexus chain

parachain/runtimes/gargantua/src/lib.rs
```diff
@@ -829,6 +879,12 @@ mod runtime {
 	pub type IsmpOptimism = ismp_optimism::pallet;
 	#[runtime::pallet_index(85)]
 	pub type IsmpTendermint = ismp_tendermint::pallet;
+	#[runtime::pallet_index(86)]
+	pub type ConsensusIncentives = pallet_consensus_incentives::pallet;
```
Wizdave97: Remove ConsensusIncentives and MessagingFees

PR #592, opened 2025-12-01
IntentGateway V2

seunlanlege:
- [x] Intent Gateway can now execute "predispatch" swaps, before placing an order. Simplifying the user experience for cross-chain token swaps. The contract may collect any dust tokens from these swaps as a source of revenue
- [x] Introduce a `protocolFeeBps` as a means of collecting protocol fees from filler liquidity
- [x] Enable limit orders that do not expire and can only be cancelled from the counterparty chain.

### DIFF
--- a/evm/src/modules/IntentGatewayV2.sol
+++ b/evm/src/modules/IntentGatewayV2.sol
@@ -616,11 +616,7 @@ contract IntentGatewayV2 is HyperApp {
             IDispatcher(hostAddr).dispatch{value: msgValue}(request);
         } else {
             // try to pay for dispatch with fee token
-            address feeToken = IDispatcher(hostAddr).feeToken();
-            uint256 fee = quote(request);
-            IERC20(feeToken).safeTransferFrom(msg.sender, address(this), fee);
-            IERC20(feeToken).approve(hostAddr, fee);
-            IDispatcher(hostAddr).dispatch(request);
+            dispatchWithFeeToken(request, msg.sender);
         }
 
         emit OrderFilled({commitment: commitment, filler: msg.sender});
@@ -691,7 +687,7 @@ contract IntentGatewayV2 is HyperApp {
                 } else {
                     IERC20(token).safeTransfer(req.beneficiary, amount);
                 }
-                
+
                 emit RevenueWithdrawn(token, amount, req.beneficiary);
             }
         }
@@ -732,9 +728,8 @@ contract IntentGatewayV2 is HyperApp {
         }
 
         _filled[body.commitment] = incoming.relayer;
-        
+
         RequestKind kind = RequestKind(uint8(incoming.request.body[0]));
-        
         if (kind == RequestKind.RefundEscrow) {
             emit EscrowRefunded({commitment: body.commitment});
         } else {
@@ -796,11 +791,7 @@ contract IntentGatewayV2 is HyperApp {
             IDispatcher(hostAddr).dispatch{value: msg.value}(request);
         } else {
             // try to pay for dispatch with fee token
-            address feeToken = IDispatcher(hostAddr).feeToken();
-            uint256 fee = quote(request);
-            IERC20(feeToken).safeTransferFrom(msg.sender, address(this), fee);
-            IERC20(feeToken).approve(hostAddr, fee);
-            IDispatcher(hostAddr).dispatch(request);
+            dispatchWithFeeToken(request, msg.sender);
         }
     }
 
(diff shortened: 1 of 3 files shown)


evm/src/modules/IntentGatewayV2.sol
```diff
@@ -0,0 +1,842 @@
...
+                uint256 requiredAmount = order.inputs[i].amount;
+
+                if (token == address(0)) {
+                    uint256 balance = address(callDispatcher).balance;
+                    if (balance < requiredAmount) revert InsufficientNativeToken();
+                    transferCalls[i] = Call({
+                        to: address(this),
+                        value: balance,
+                        data: ""
+                    });
```
Wizdave97, line 469: Any reason we don't emit dust collected for native tokens

evm/src/modules/IntentGatewayV2.sol
```diff
@@ -0,0 +1,842 @@
...
+     * @param options The options to be used when filling the order.
+     * @dev This function is payable and can accept Ether.
+     */
+    function fillOrder(Order calldata order, FillOptions memory options) public payable {
+        address hostAddr = host();
+        // Ensure the order is being filled on the correct chain
+        if (keccak256(order.destChain) != keccak256(IDispatcher(hostAddr).host())) revert WrongChain();
+
+        // Ensure the order has not expired
+        if (order.deadline < block.number) revert Expired();
```
Wizdave97, line 553: Missing check for zero deadline

evm/src/modules/IntentGatewayV2.sol
```diff
@@ -0,0 +1,842 @@
...
+    /**
+     * @dev Enum representing the different kinds of incoming requests that can be executed.
+     */
+    enum RequestKind {
+        /// @dev Identifies a request for redeeming an escrow.
+        RedeemEscrow,
+        /// @dev Identifies a request for recording new contract deployments
+        NewDeployment,
+        /// @dev Identifies a request for updating parameters.
+        UpdateParams
```
Wizdave97: Think we need a command to do something with accrued protocol fees

evm/src/modules/IntentGatewayV2.sol
seunlanlege: Actually for limit orders, we should use `type(uint256).max` so they never expire

evm/src/modules/IntentGatewayV2.sol
```diff
@@ -0,0 +1,854 @@
...
+
+                    (bool success, ) = callDispatcher.call{value: amount}("");
+                    require(success, "Native token transfer failed");
+                } else {
+                    IERC20(token).safeTransferFrom(msg.sender, callDispatcher, amount);
+                }
+            }
+
+            // Execute the call dispatcher with predispatch call
+            ICallDispatcher(callDispatcher).dispatch(order.predispatch.call);
```
royvardhan, line 452: I think we should block the commitment of the order before reaching this line. Wdyt?

royvardhan: Or we should have a simple state bool that makes this function locked. And the lock is released after the execution of full placeOrder is over.

evm/src/modules/IntentGatewayV2.sol
```diff
@@ -0,0 +1,854 @@
...
+            }
+        }
+
+        // dispatch calls if any
+        if (order.callData.length > 0) {
+            ICallDispatcher(_params.dispatcher).dispatch(order.callData);
+        }
+
+        // Process inputs and collect protocol fees
+        TokenInfo[] memory inputs = _processInputsAndCollectFees(order);
```
royvardhan, line 594: If we are using this to redeem, then we need to emit protocol fees collected event in the redeem function?

seunlanlege: The model for protocol fees is to deduct them from the filler's redemption, by only allowing them redeem the escrowed assets minus the protocol fees

evm/src/modules/IntentGatewayV2.sol
seunlanlege: Are you envisioning any re-entrancy attacks?

royvardhan: Havn't thought of an exact way to attack but not comfortable executing arbitrary calldata before blocking.

evm/src/modules/IntentGatewayV2.sol
seunlanlege: Done

evm/src/modules/IntentGatewayV2.sol
seunlanlege: done

evm/src/modules/IntentGatewayV2.sol
```diff
@@ -0,0 +1,899 @@
...
+     * This address is used to store or reference the transaction fees within the contract.
+     */
+    address private constant TRANSACTION_FEES = address(uint160(uint256(keccak256("txFees"))));
+
+    /**
+     * @notice Constant representing a filled slot in big endian format
+     * @dev Hex value 0x05 padded with leading zeros to fill 32 bytes
+     */
+    bytes32 constant FILLED_SLOT_BIG_ENDIAN_BYTES =
+        hex"0000000000000000000000000000000000000000000000000000000000000005";
```
seunlanlege: I think we might have to update this filled slot @royvardhan

Wizdave97: Forge can list it `forge inspect IntentGatewayV2 storage`

PR #598, opened 2025-12-08
[evm]: add UniV4 wrapper, update UniV3 wrapper

royvardhan:
- [x] Added UniV4 wrapper
- [x] The UniV3 wrapper now uses SwapRouter02 and QuoterV2

### DIFF
--- a/evm/foundry.toml
+++ b/evm/foundry.toml
@@ -2,7 +2,8 @@
 src = "src"
 out = "out"
 libs = ["lib"]
-auto_detect_solc = true
+auto_detect_solc = false
+solc = "0.8.30"
 optimizer = true
 evm_version = "cancun"
 # via-ir = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10788,7 +10788,7 @@ dependencies = [
  "ckb-merkle-mountain-range",
  "ethers",
  "ethers-contract-abigen",
- "forge-testsuite",
+ "ethers-solc",
  "ismp",
  "mmr-primitives",
  "polkadot-sdk",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -350,6 +350,10 @@ default-features = false
 git = "https://github.com/polytope-labs/ethers-rs"
 rev = "45239225c50247e049892125d281442c084a2a92"
 
+[workspace.dependencies.ethers-solc]
+git = "https://github.com/polytope-labs/ethers-rs"
+rev = "45239225c50247e049892125d281442c084a2a92"
+
 [workspace.dependencies.forge-testsuite]
 git = "https://github.com/polytope-labs/forge-testsuite"
 rev = "ff9fdc69c0200e3bb471198aa44ba933e043b78d"
--- a/evm/abi/Cargo.toml
+++ b/evm/abi/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 [build-dependencies]
 anyhow.workspace = true
 ethers-contract-abigen = { workspace = true }
-forge-testsuite = { workspace = true }
+ethers-solc = { workspace = true }
 
 [dependencies]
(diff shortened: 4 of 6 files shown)


evm/src/modules/UniV3UniswapV2Wrapper.sol
```diff
@@ -17,24 +17,61 @@ pragma solidity ^0.8.17;
...
+    function exactOutputSingle(ExactOutputSingleParams calldata params) external payable returns (uint256 amountIn);
+}
+
+/**
+ * @title ISwapRouter02
+ * @notice Extended interface for SwapRouter02 with multicall support
+ */
+interface ISwapRouter02 is IV3SwapRouter {
+    function multicall(uint256 deadline, bytes[] calldata data) external payable returns (bytes[] memory);
+}
```
seunlanlege: shouldn't these interfaces come from their sdk?

royvardhan: Yeah, installing.

evm/src/modules/UniV4UniswapV2Wrapper.sol
```diff
@@ -0,0 +1,167 @@
...
+            abi.encodePacked(bytes1(uint8(Commands.V4_SWAP))),
+            inputs,
+            deadline
+        );
+
+        amounts = new uint256[](2);
+        amounts[0] = msg.value;
+        amounts[1] = amountOut;
+
+        IERC20(path[1]).safeTransfer(recipient, amounts[1]);
```
royvardhan: Looking to remove this extra transfer through commands.

evm/lib/swap-router-contracts
seunlanlege, line 1: can we use npm for these dependencies, gitsubmodules are a bit large

royvardhan: Yes sure.

.gitmodules
```diff
@@ -7,3 +7,18 @@
...
+	url = https://github.com/Uniswap/v4-periphery
+[submodule "evm/lib/universal-router"]
+	path = evm/lib/universal-router
+	url = https://github.com/Uniswap/universal-router
+[submodule "evm/lib/permit2"]
+	path = evm/lib/permit2
+	url = https://github.com/Uniswap/permit2
+[submodule "evm/lib/swap-router-contracts"]
+	path = evm/lib/swap-router-contracts
+	url = https://github.com/Uniswap/swap-router-contracts
```
seunlanlege: we need to remove these

royvardhan:
> @royvardhan, what's left here?

Sweeping the dust, and getting the Universal router to send directly to the recipient.

evm/foundry.toml
```diff
@@ -4,10 +4,11 @@ out = "out"
 libs = ["lib"]
 auto_detect_solc = true
 optimizer = true
-evm_version = "paris"
+evm_version = "cancun"
```
seunlanlege, line 8: why do we need this

royvardhan: So that we can have opcode available in cancun used by v4.

evm/package.json
```diff
@@ -1,30 +1,36 @@
...
+    "version": "1.0.0",
+    "main": "index.js",
+    "directories": {
+        "example": "examples",
+        "lib": "lib",
+        "test": "test"
+    },
+    "scripts": {
+        "test": "echo \"Error: no test specified\" && exit 1",
+        "postinstall": "forge install"
```
Wizdave97: You should add forge to the CI if we need to use this.

Wizdave97: @royvardhan is the post install script necessary?

royvardhan: I had to add it because the build was failing. Now that I have cleaned gitmodules need to check the build again without it

PR #599, opened 2025-12-08
[intentv2]: Solver Selection

seunlanlege:
Introduces a solver selection mechanism, this involves a session key added to the order which can authorize a selected solver to fill an order after order auctions take place.

The "client" will sign a message, authorizing a specific solver's solution, while the protocol keeps the surplus as revenue

### DIFF
--- a/evm/src/modules/IntentGatewayV2.sol
+++ b/evm/src/modules/IntentGatewayV2.sol
@@ -624,10 +624,8 @@ contract IntentGatewayV2 is HyperApp {
             address token = address(uint160(uint256(order.output.assets[i].token)));
             uint256 requestedAmount = order.output.assets[i].amount;
 
-            // Verify that solver's output matches the order output token and beneficiary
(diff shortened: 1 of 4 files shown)


evm/src/modules/IntentGatewayV2.sol
```diff
@@ -538,73 +542,141 @@ contract IntentGatewayV2 is HyperApp {
...
+                if (msgValue < solverAmount) revert InsufficientNativeToken();
 
-                // Send amount after fee to beneficiary
-                (bool sent, ) = beneficiary.call{value: amount}("");
+                // Send requested amount to beneficiary
+                (bool sent, ) = beneficiary.call{value: requestedAmount}("");
                 if (!sent) revert InsufficientNativeToken();
 
-                msgValue -= amount;
+                msgValue -= solverAmount;
```
Wizdave97: Shouldn't we subtract `requestedAmount` here?

PR #600, opened 2025-12-08
fix: withdraw fees input param

darwinsubramaniam:
There was issue with payload missing param beneficiary
image

This fixes the issue by checking if there is any value given

if there is not value then None Value variant is being created , if not create Some Value Variant with vec  byte of the address being used.


Wizdave97 (approved):

Wizdave97:
Thank you @darwinsubramaniam

PR #605, opened 2025-12-14
[evm]: use EIP-712 in `IntentGatewayV2`

seunlanlege:
It's a better framework for signed message security.


evm/src/modules/IntentGatewayV2.sol
```diff
@@ -585,7 +626,8 @@ contract IntentGatewayV2 is HyperApp {
         // Ensure the order has not expired
         if (order.deadline < block.number) revert Expired();
         bytes32 commitment = keccak256(abi.encode(order));
-        if (_params.solverSelection) {
+        // cross-chain limit orders don't need solver selection, they can be filled directly
+        if (_params.solverSelection && order.deadline != type(uint256).max) {
```
seunlanlege: I'm a bit torn on this decision

royvardhan: For a limit order, the solver can still provide better quotes? Allowing protocol to pocket the difference?

royvardhan (approved):

evm/src/modules/IntentGatewayV2.sol
seunlanlege: Nahh for limit orders they can be filled directly, so no solver competition or selection.

The rationale is clients might be offline when their order becomes eligible for filling so they'd miss the window.

I'm not sure how we can enable solver selection for limit orders.

PR #614, opened 2025-12-22
[evm]: `ERC4337` Account for solvers

seunlanlege:
Introduces the `SolverAccount` an [`ERC-4337`](https://eips.ethereum.org/EIPS/eip-4337) compliant account for solvers, specifically for sponsoring `select` transactions.

This is built on openzeppelin's ERC-4337 `Account`.

Actual diff is [02c3904](https://github.com/polytope-labs/hyperbridge/pull/614/commits/02c3904d318fefe57a98813db973ef492c7a6c58). Had to refactor the project to upgrade openzeppelin

### DIFF
--- a/evm/src/SolverAccount.sol
+++ b/evm/src/SolverAccount.sol
@@ -20,7 +20,7 @@ import {ERC7821} from "@openzeppelin/contracts/account/extensions/draft-ERC7821.
 import {PackedUserOperation} from "@openzeppelin/contracts/interfaces/draft-IERC4337.sol";
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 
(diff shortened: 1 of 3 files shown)


evm/src/SolverAccount.sol
```diff
@@ -0,0 +1,235 @@
...
+
+        // Decode SelectOptions from calldata (skip 4-byte selector)
+        SelectOptions memory options = abi.decode(op.callData[4:], (SelectOptions));
+        if (options.solver != address(this)) return ERC4337Utils.SIG_VALIDATION_FAILED;
+
+        address sessionKey = address(bytes20(op.signature[0:20]));
+        bytes calldata solverSignature = op.signature[20:];
+
+        // Get nonce from EntryPoint using first 192 bits of commitment as key
+        uint192 nonceKey = uint192(uint256(options.commitment) >> 64);
```
Wizdave97: Should we use a constant for the key just so that we don't bloat the chain's storage?

Issue #620, opened 2025-12-30
get_rotation_block  will infinite loop when validator_size / 2 >= epoch_length


borngraced:
Came across https://github.com/polytope-labs/hyperbridge/blob/acfec13d9814c37b83f2a5b51b4bc1fd5dd33c29/modules/consensus/bsc/prover/src/lib.rs#L173C1-L186C2 while reading the code...

When `validator_size / 2 >= epoch_length`, the condition can never be satisfied since `block % epoch_length` will always be in the range `[0, epoch_length - 1]`. This will causes an infinite loop...and moreover, the entire code can be replaced with `O(1)` version

PR #621, opened 2025-12-31
fix infinite loop in get_rotation_block with O(1) implementation

borngraced:
#620


Wizdave97:
Thanks, I'll take a look.

PR #628, opened 2026-01-13
Pharos Network Integration

### LINKED ISSUE #593: Pharos Network Integration
Pharos is an upcoming EVM-compatible L1.

This describes the implementation details for integrating Pharos network with Hyperbridge.
 Network Info
1. The active validator set for the network is stored on-chain in a system contract at address `0x4100000000000000000000000000000000000000`. More information in the link
https://silken-muskox-24e.notion.site/Pharos-Staking-Contract-2b18ec314f7580c1b885e2fa8d8a70e9

Mainnet Epoch length: 4 hours
Testnet Epoch Length: 0.5 hours
The validator set is updated in the last block of each epoch, and these changes take effect in the next block, that is, the first block of the new epoch.

A proof for a block can be fetched using the `debug_getBlockProof` api https://docs.pharosnetwork.xyz/api-and-sdk/json-rpc-api-methods#debug_getblockproof
The response contains the aggregated BLS signature for the block and the public keys of participating validators.

Authority set rotation:
Authority-set rotations occur at the last block of an epoch, so a consensus update for that block must be submitted to the consensus client.

The consensus update for the last block in an epoch should be accompanied by a state proof of the new validator set as recorded in the staking contract described above.

- [ ] Pharos prover 
       Fetch block proofs from the debug_BlockProof Api
       For the last block in an epoch fetch a state proof of the new validator set
- [ ] Pharos verifier
       Verify bls signature of the block
        Vwerify authority set state proof if the block is an epoch boundary block
- [ ] Pharos Ismp Consensus Client
- [ ] Pharos tesseract Consensus Client

dharjeezy:
This PR introduces the integration to Pharos network for  Hyperbridge.

- [x] Pharos network verifier
- [x] ISMP consensus 
- [x] Pharos Prover
- [x] Tesseract Relayer Intergation
- [x] Pharos State Machine

closes #593

### DIFF
--- a/parachain/runtimes/gargantua/src/lib.rs
+++ b/parachain/runtimes/gargantua/src/lib.rs
@@ -238,7 +238,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("gargantua"),
 	impl_name: Cow::Borrowed("gargantua"),
 	authoring_version: 1,
-	spec_version: 5_100,
+	spec_version: 5_200,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,


modules/consensus/pharos/primitives/src/types.rs
```diff
@@ -0,0 +1,235 @@
...
+pub struct ValidatorSetProof {
+	/// The new validator set
+	pub validator_set: ValidatorSet,
+	/// Merkle-Patricia trie proof nodes from the staking contract storage
+	/// to the state root of the block
+	pub storage_proof: Vec<Vec<u8>>,
+	/// The account proof for the staking contract
+	pub account_proof: Vec<Vec<u8>>,
+	/// The storage slot(s) being proven
+	pub storage_slots: Vec<H256>,
```
Wizdave97: The light client should calculate the slot hash on chain

modules/consensus/pharos/primitives/src/types.rs
```diff
@@ -0,0 +1,235 @@
...
+///
+/// This state is updated as new blocks are verified and represents
+/// the current view of the chain from the light client's perspective.
+#[derive(Debug, Clone, Encode, Decode)]
+pub struct VerifierState {
+	/// The current (active) validator set
+	pub current_validator_set: ValidatorSet,
+	/// The next validator set (if known from a recent epoch boundary)
+	/// This becomes current_validator_set at the next epoch
+	pub next_validator_set: Option<ValidatorSet>,
```
Wizdave97: Probably don't need this field, we should rotate the validator set immediately we get a proof for the next validators

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+pub const PHAROS_BLS_DST: &str = "BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
+
+/// Verifies a Pharos block proof and update the verifier state.
+pub fn verify_pharos_block<C: Config, H: Keccak256 + Send + Sync>(
+	trusted_state: VerifierState,
+	update: VerifierStateUpdate,
+) -> Result<VerifierState, Error> {
+	let update_block_number = update.block_number();
+	let current_block_number = trusted_state.finalized_block_number();
+
```
Wizdave97, line 43: We need a check to show that the update block is in the same epoch as the verifier state

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+		// If not an epoch boundary, then, no validator set update expected
+		if update.validator_set_proof.is_some() {
+			return Err(Error::UnexpectedValidatorSetProof { block_number: update_block_number });
+		}
+		None
+	};
+
+	let new_state = if let Some(new_set) = new_validator_set.clone() {
+		VerifierState {
+			current_validator_set: trusted_state
```
Wizdave97: current validator set should immediately be set as new_validator_set

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+				.next_validator_set
+				.unwrap_or(trusted_state.current_validator_set),
+			next_validator_set: Some(new_set),
```
Wizdave97:
```suggestion
```

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/// Verify that participating validators have more than 2/3 of total stake.
+fn verify_stake_threshold(
+	validator_set: &ValidatorSet,
+	participants: &[BlsPublicKey],
+) -> Result<(), Error> {
+	let participating_stake = validator_set.participating_stake(participants);
+	let total_stake = validator_set.total_stake;
+
```
Wizdave97, line 142: The check should be a simple 
```rust 
     if  participating_stake >= ((2/3 *total_stake) + 1) { Ok(()) } else { Er(()) }
```

modules/consensus/pharos/primitives/src/constants.rs
```diff
@@ -0,0 +1,109 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Constants and configuration for Pharos consensus.
+
+use alloy_primitives::Address;
+
+/// Re-export BLS types from sync-committee-primitives
+pub use sync_committee_primitives::constants::{
```
seunlanlege: Maybe it makes sense to move this to utils/crypto?

modules/consensus/pharos/verifier/src/error.rs
```diff
@@ -0,0 +1,113 @@
...
+	#[error("Account proof verification failed: {0}")]
+	AccountProofFailed(String),
+
+	/// Storage proof verification failed
+	#[error("Storage proof verification failed: {0}")]
+	StorageProofFailed(String),
+
+	/// General verification error
+	#[error("Verification failed: {0}")]
+	VerificationFailed(String),
```
seunlanlege: Let's try to avoid string errors

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,636 @@
...
+	const LENGTH: usize = 32;
+
+	fn hash(x: &[u8]) -> Self::Out {
+		H::keccak256(x)
+	}
+}
+
+/// The ethereum account stored in the global state trie.
+#[derive(RlpDecodable, RlpEncodable, Debug, Clone)]
+pub struct Account {
```
seunlanlege: Feel like this should come from alloy or some other crate

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Verify the BLS aggregate signature.
+fn verify_bls_signature(
+	participants: &[BlsPublicKey],
+	block_proof: &BlockProof,
+) -> Result<(), Error> {
+	if participants.is_empty() {
+		return Err(Error::NoParticipants);
+	}
+
+	let aggregate_pubkey = bsc_verifier::aggregate_public_keys(participants);
```
Wizdave97: I think we have this method in s crypto utils, if not let's move it there

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+		}
+	};
+
+	Ok(new_state)
+}
+
+/// Compute the hash of a block header using RLP encoding and Keccak256.
+///
+/// This follows the standard Ethereum header hash computation.
+pub fn compute_header_hash<H: Keccak256>(header: &CodecHeader) -> H256 {
```
Wizdave97: We already have this in the geth primitives, there's a hash function implemented on the header

modules/consensus/pharos/primitives/src/types.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// State proof for validator set stored in the staking contract.
+///
+/// This proof is required when the validator set changes at epoch boundaries.
+/// It proves the new validator set against the state root of the epoch boundary block.
+#[derive(Debug, Clone, PartialEq, Eq, Default, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct ValidatorSetProof {
+	/// The new validator set
+	pub validator_set: ValidatorSet,
```
Wizdave97: Because of how EVM slots work, we don't need the ValidatorSet in the proof, 

The staking contract has the following in storage
```
    mapping(bytes32 => Validator) public validators;
    bytes32[] public activePoolIds;
 ```
 
 1. We need the storage proof to contain the activePoolIds array.
 2. Since the validator map key is the poolId, for each poolId in activePoolId, we need the proof to contain the bls_public_key and total_stake of that validator. That's all we need.
 
 Note that you can't fetch a proof for a full struct in EVM, you can only fetch a proof for each field in the struct using the slot hash for each field

File: modules/consensus/pharos/verifier/src/state_proof.rs [hunk 13]
```diff
@@ -0,0 +1,708 @@
...
+		.map_err(|_| Error::AccountTrieLookupFailed)?
+		.ok_or(Error::StakingContractNotFound)?;
+
+	let account = Account::decode(&mut &*result).map_err(|_| Error::AccountRlpDecodeFailed)?;
+
+	Ok(account)
+}
+
+/// Get values from storage proof.
+fn get_values_from_proof<H: Keccak256 + Send + Sync>(
```
Wizdave97: Import all these EVM utility methods from the evm state machine instead of duplicating

modules/consensus/pharos/primitives/src/types.rs
Wizdave97: I see you have the implementation in the decode_validator_set

Wizdave97:
```suggestion
```
The proof doesn't need to come with the validator set, we can trust the values we decode from the proof

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,708 @@
...
+/// in the staking contract at the given block.
+pub fn verify_validator_set_proof<H: Keccak256 + Send + Sync>(
+	state_root: H256,
+	proof: &ValidatorSetProof,
+) -> Result<(), Error> {
+	let account = get_staking_contract_account::<H>(&proof.account_proof, state_root)?;
+	let storage_root = H256::from_slice(account.storage_root.as_slice());
+
+	let layout = StakingContractLayout::default();
+	let validator_count = proof.validator_set.validators.len() as u64;
```
Wizdave97: We can just have a value activePoolIdTotalCount on the `ValidatorSetProof`

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,708 @@
...
+	// Compute and fetch global storage slots (epoch, totalStake, array length, pool IDs)
+	let global_slots = layout.get_validator_set_keys::<H>(validator_count);
+	let global_values = get_values_from_proof::<H>(
+		global_slots.iter().map(|s| H::keccak256(s.as_bytes()).0.to_vec()).collect(),
+		storage_root,
+		proof.storage_proof.clone(),
+	)?;
+
+	// Extract pool IDs from the proven values
+	let pool_set_start = 3; // After epoch, totalStake, array length
```
Wizdave97: We should get active poolIds   from a proof of this  array ` bytes32[] public activePoolIds;`

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,708 @@
...
+			.unwrap_or_default();
+		pool_ids.push(pool_id);
+	}
+
+	let mut validator_slots = Vec::new();
+	for pool_id in &pool_ids {
+		validator_slots.extend(layout.get_validator_keys::<H>(pool_id));
+	}
+
+	let validator_values = get_values_from_proof::<H>(
```
Wizdave97: We then use the poolIds decoded from the proof to get the values for each validator's bls_public_key and stake,these are the only values we need from the ValidatorInfo

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,708 @@
...
+		validator_slots.iter().map(|s| H::keccak256(s.as_bytes()).0.to_vec()).collect(),
+		storage_root,
+		proof.storage_proof.clone(),
+	)?;
+
```
Wizdave97, line 60: Our proof should also contain a proof of the ` uint256 public totalStake;` item

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,708 @@
...
+	let decoded_set = decode_validator_set_from_storage::<H>(&all_slots, &all_values)?;
+
+	verify_validator_set_matches(&proof.validator_set, &decoded_set)?;
+
+	Ok(())
+}
+
+/// Get the staking contract account from the account proof.
+fn get_staking_contract_account<H: Keccak256 + Send + Sync>(
```
Wizdave97: We have utility contracts for these in evm-state machine

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,708 @@
...
+
+	let mut all_slots = global_slots;
+	all_slots.extend(validator_slots);
+
+	let mut all_values = global_values;
+	all_values.extend(validator_values);
+
+	let decoded_set = decode_validator_set_from_storage::<H>(&all_slots, &all_values)?;
+
+	verify_validator_set_matches(&proof.validator_set, &decoded_set)?;
```
Wizdave97: Not needed, the values decoded from the proof are correct

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,708 @@
...
+/// Pharos staking contract's actual storage layout.
+///
+/// ## Pharos Storage Layout
+///
+/// The contract at `0x4100000000000000000000000000000000000000` uses:
+/// - Slot 0: validators mapping (mapping(bytes32 => Validator))
+/// - Slot 5: currentEpoch (uint256)
+/// - Slot 6: totalStake (uint256)
+/// - Slot 22: activePoolSets (EnumerableSet._values array)
+fn decode_validator_set_from_storage<H: Keccak256>(
```
Wizdave97, line 86: We only need to decode the validator's bls_public_key and stake

modules/consensus/pharos/verifier/src/state_proof.rs
Wizdave97: I see this is in line with the StakingV2Storage

modules/consensus/pharos/verifier/src/state_proof.rs
Wizdave97: The totalStake can change

dharjeezy: yes, this is what is being done

modules/consensus/pharos/primitives/src/types.rs
```diff
@@ -0,0 +1,218 @@
...
+	/// The block number this proof is for (from JSON "blockNumber")
+	pub block_number: u64,
+	/// The block hash (derived from header)
+	pub block_hash: H256,
+	/// The block proof hash - this is the message that validators sign
+	/// (from JSON "blockProofHash")
+	///
+	/// This is distinct from the block hash and is the actual message
+	/// that the BLS signature is computed over.
+	pub block_proof_hash: H256,
```
Wizdave97: We should generate this hash onchain, this is the message signed by validators the light client should generate it, it should not be submitted as part of the proof, rather submit all the data required to generate this hash

modules/consensus/pharos/primitives/src/types.rs
```diff
@@ -0,0 +1,218 @@
...
+///     SignedBlsKeys          []string `json:"signedBlsKeys"`
+/// }
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, Default, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct BlockProof {
+	/// The block number this proof is for (from JSON "blockNumber")
+	pub block_number: u64,
+	/// The block hash (derived from header)
+	pub block_hash: H256,
```
Wizdave97: The block header should be submitted not the block header hash, the light client will compute the header hash on chain

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,584 @@
...
+		return Err(Error::InsufficientStorageValues { expected: 2, got: values.len() });
+	}
+
+	// Parse global state
+	// Index 0: totalStake
+	let total_stake = values[0]
+		.as_ref()
+		.map(|v| decode_u256_from_storage(v))
+		.transpose()?
+		.unwrap_or_default();
```
Wizdave97: This should fail if not available `unwrap_or_default()`

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,584 @@
...
+				.unwrap_or_default();
+
+			let validator = ValidatorInfo { bls_public_key: bls_key, pool_id, stake };
+
+			validator_set.validators.push(validator);
+		}
+	} else {
+		// Minimal proof - only pool IDs, no detailed validator data
+		// This is used when the validator set is provided separately
+		// and only pool IDs need to be verified
```
Wizdave97: This should fail if the data is incomplete

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,584 @@
...
+	if validator_set.validators.is_empty() {
+		return Err(Error::EmptyValidatorSet);
+	}
+
+	let computed_total: U256 = validator_set
+		.validators
+		.iter()
+		.fold(U256::zero(), |acc, v| acc.saturating_add(v.stake));
+
+	if computed_total != validator_set.total_stake {
```
Wizdave97: This check is not needed

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,584 @@
...
+		return Err(Error::ComputedStakeMismatch {
+			computed: computed_total,
+			claimed: validator_set.total_stake,
+		});
+	}
+
+	for (i, v1) in validator_set.validators.iter().enumerate() {
```
Wizdave97: You can just use a BtreeSet, then we don't need to do this

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+	)?;
+
+	verify_stake_threshold(
+		&trusted_state.current_validator_set,
+		&update.block_proof.participant_keys,
+	)?;
+
+	let computed_hash = Header::from(&update.header).hash::<H>();
+
+	verify_bls_signature(&update.block_proof.participant_keys, &update.block_proof, computed_hash)?;
```
Wizdave97: Is the `block_proof_hash` the same as the header hash?

dharjeezy: oh...yes. According to the implementation in pharos' codebase.
https://github.com/PharosNetwork/AntChainAldabaNG/blob/pharos-reaudit-contract/aldaba-ng/libraries/structures/evm/block_header.cpp#L90

modules/consensus/pharos/primitives/Cargo.toml
```diff
@@ -15,6 +15,7 @@ crypto-utils = { workspace = true, default-features = false }
 primitive-types = { workspace = true, features = ["serde_no_std", "impl-codec"] }
 hex-literal = { workspace = true }
 serde = { workspace = true, optional = true, features = ["derive"] }
+sha2 = { version = "0.10", default-features = false }
```
Wizdave97: We should hashing from sp_io

modules/consensus/pharos/prover/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+		let base_keys = vec![
+			self.storage_layout.raw_slot_key(self.storage_layout.total_stake_slot),
+			self.storage_layout.raw_slot_key(self.storage_layout.active_pool_set_slot),
+		];
+
+		let base_proof = self.rpc.get_proof(address, base_keys.clone(), block_number).await?;
+
+		// Get validator count from activePoolIds length (slot 1)
+		let validator_count = if base_proof.storage_proof.len() >= 2 {
+			hex_to_u64(&base_proof.storage_proof[1].value).unwrap_or(0)
```
Wizdave97: We should return an error is the value is invalid instead of using a default value of zero

modules/consensus/pharos/prover/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+		// Fetch pool IDs from the activePoolIds array
+		let mut pool_id_keys = Vec::new();
+		for i in 0..validator_count {
+			pool_id_keys.push(self.array_element_key(self.storage_layout.active_pool_set_slot, i));
+		}
+
+		let pool_id_proof = if !pool_id_keys.is_empty() {
+			self.rpc.get_proof(address, pool_id_keys.clone(), block_number).await?
+		} else {
+			base_proof.clone()
```
Wizdave97: Return an error instead

modules/consensus/pharos/prover/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+		for pool_id in &pool_ids {
+			let validator_keys = self.get_validator_storage_keys(pool_id);
+
+			match self.rpc.get_proof(address, validator_keys.clone(), block_number).await {
+				Ok(val_proof) =>
+					for sp in &val_proof.storage_proof {
+						self.collect_proof_nodes(&sp.proof, &mut storage_proof)?;
+						storage_values.push(hex_to_bytes(&sp.value)?);
+					},
+				Err(_) => {
```
Wizdave97: Return an error if it fails,  we cannot verify the proof successfully with incomplete proof nodes

modules/consensus/pharos/prover/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+				proof_node: hex_to_bytes(&n.proof_node)?,
+				next_begin_offset: n.next_begin_offset,
+				next_end_offset: n.next_end_offset,
+			})
+		})
+		.collect()
+}
+
+/// Simple keccak256 implementation using tiny_keccak.
+fn keccak256(data: &[u8]) -> [u8; 32] {
```
Wizdave97, line 437: sp_io provides a keccak implementation

modules/consensus/pharos/prover/src/rpc.rs
```diff
@@ -0,0 +1,273 @@
...
+			return Err(ProverError::RpcError { code: error.code, message: error.message });
+		}
+
+		rpc_response.result.ok_or(ProverError::MissingRpcResult)
+	}
+
+	/// Fetch block proof using `debug_getBlockProof`.
+	pub async fn get_block_proof(&self, block_number: u64) -> Result<RpcBlockProof, ProverError> {
+		let block_hex = format!("0x{:x}", block_number);
+		self.call("debug_getBlockProof", vec![block_hex]).await
```
Wizdave97, line 274: We should use alloy client for these rpc queries except  eth_getProof and the debug endpoints

modules/consensus/pharos/prover/src/rpc.rs
```diff
@@ -0,0 +1,273 @@
...
+	client: reqwest::Client,
+	request_id: AtomicU64,
+}
+
+impl PharosRpcClient {
+	/// Create a new RPC client.
+	pub fn new(endpoint: impl Into<String>) -> Self {
+		Self {
+			endpoint: endpoint.into(),
+			client: reqwest::Client::new(),
```
Wizdave97, line 174: Can we use the Alloy client for the Ethereum-compatible endpoints?

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,502 @@
...
+	// - 1 BLS string slot (header)
+	// - 4 BLS data slots (for long string: 98 chars with "0x" prefix)
+	// - 1 totalStake
+	const VALIDATOR_FIELDS: usize = 6;
+
+	let validators_data_start = pool_ids_end;
+	let expected_total = validators_data_start + (count * VALIDATOR_FIELDS);
+
+	// If we have detailed validator data
+	if values.len() >= expected_total {
```
Wizdave97:
```suggestion
	if values.len() == expected_total {
```

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+	error::Error,
+	host::IsmpHost,
+	messaging::{hash_request, hash_response, Keccak256, Proof},
+	router::RequestResponse,
+};
+use pallet_ismp_host_executive::EvmHosts;
+use pharos_primitives::{spv, PharosProofNode};
+use primitive_types::{H160, H256, U256};
+
+/// Slot index for requests commitments map in the ISMP contract
```
Wizdave97: These constants should be imported from the evm-state-machine

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,502 @@
...
+	) {
+		return Err(Error::AccountTrieLookupFailed);
+	}
+
+	let layout = StakingContractLayout::default();
+
+	// Decode the validator set from the storage proof
+	let decoded_set = decode_validator_set_from_proof::<H>(proof, &layout, epoch)?;
+
+	validate_validator_set(&decoded_set)?;
```
Wizdave97: I don't see any place where we verify the storage proof for the validator set

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+	if !spv::verify_account_proof(
+		&pharos_proof.contract_proof,
+		&address,
+		&pharos_proof.raw_account_value,
+		&state_root.0,
+	) {
+		return Err(Error::Custom("Invalid contract account proof".to_string()));
+	}
+
+	let storage_hash = pharos_proof.storage_hash;
```
Wizdave97: This is an unverified storage_hash, the storage_hash should be extracted from the raw_account_value after it's verified. The account value should be an rlp encoded ethereum account that can be decoded and it should have the storage root for that contract

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+		let mut trie_key = Vec::with_capacity(52);
+		trie_key.extend_from_slice(&address);
+		trie_key.extend_from_slice(&slot_hash);
+
+		let storage_proof_nodes = pharos_proof
+			.storage_proof
+			.get(&slot_hash)
+			.ok_or_else(|| Error::Custom("Missing storage proof for commitment key".to_string()))?;
+
+		let leaf = storage_proof_nodes
```
Wizdave97: Why not use the functions defined in the spv crate to do this state proof verification instead of duplicating the algorithm here

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+///
+/// Contains Pharos hexary hash tree proof data with SHA-256 hashing.
+#[derive(Encode, Decode, Clone)]
+pub struct PharosStateProof {
+	/// Account proof nodes for the   contract
+	pub contract_proof: Vec<PharosProofNode>,
+	/// Map of storage key (slot hash) to storage proof nodes
+	pub storage_proof: BTreeMap<Vec<u8>, Vec<PharosProofNode>>,
+	/// Storage trie root from eth_getProof
+	pub storage_hash: H256,
```
Wizdave97:
```suggestion
```
This should not be supplied by the rpc, it should be extracted from the raw_account_value after it's verified

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+			let leaf = storage_proof_nodes.last();
+			if let Some(leaf) = leaf {
+				if leaf.proof_node.len() == 65 {
+					// Verify the proof chain
+					let expected_key_hash = spv::sha256(&trie_key);
+					if leaf.proof_node[1..33] == expected_key_hash {
+						let mut current_hash = spv::sha256(&leaf.proof_node);
+						let mut valid = true;
+
+						for i in (0..storage_proof_nodes.len() - 1).rev() {
```
Wizdave97: Same thing here

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	Ok(map)
+}
+
+/// Compute Solidity storage slot keys for ISMP request/response commitments.
+///
+/// Solidity layout: `mapping(bytes32 => FungibleAssetRegistration)` at a given slot.
+/// Key derivation: `keccak256(keccak256(commitment || uint256(slot)) + offset)`
+fn req_res_commitment_key<H: Keccak256>(item: RequestResponse) -> Vec<Vec<u8>> {
```
Wizdave97: Why are we redefining this function? Please import it from the evm-state-machine

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+				);
+				keys.push(key.0.to_vec())
+			},
+	}
+	keys
+}
+
+/// Compute Solidity storage slot keys for ISMP request/response receipts.
+fn req_res_receipt_keys<H: Keccak256>(item: RequestResponse) -> Vec<Vec<u8>> {
+	let mut keys = vec![];
```
Wizdave97: Import

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+					derive_unhashed_map_key::<H>(commitment.0.to_vec(), RESPONSE_RECEIPTS_SLOT);
+				keys.push(key.0.to_vec())
+			},
+	}
+	keys
+}
+
+/// Derive a Solidity mapping key with offset:
+/// `keccak256(uint256(keccak256(key || uint256(slot))) + offset)`
+fn derive_map_key_with_offset<H: Keccak256>(mut key: Vec<u8>, slot: u64, offset: u64) -> H256 {
```
Wizdave97: Import if you need to use this

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+	key.extend_from_slice(&U256::from(slot).to_big_endian());
+	let root_key = H::keccak256(&key).0;
+	let number = U256::from_big_endian(root_key.as_slice()) + U256::from(offset);
+	H::keccak256(&number.to_big_endian())
+}
+
+/// Derive an unhashed Solidity mapping key: `keccak256(key || uint256(slot))`
+fn derive_unhashed_map_key<H: Keccak256>(mut key: Vec<u8>, slot: u64) -> H256 {
```
Wizdave97: Same

modules/pallets/testsuite/src/tests/pharos_state_machine.rs
```diff
@@ -0,0 +1,256 @@
...
+	rpc::{hex_to_bytes, hex_to_h256, PharosRpcClient, RpcProofNode},
+	PharosProver,
+};
+use primitive_types::{H160, H256, U256};
+use std::sync::Arc;
+
+const ATLANTIC_RPC: &str = "https://atlantic-rpc.dplabs-internal.com";
+
+/// Convert RPC proof nodes to PharosProofNode format.
+fn rpc_to_proof_nodes(nodes: &[RpcProofNode]) -> Vec<PharosProofNode> {
```
Wizdave97: Duplicate function, import it

modules/pallets/testsuite/src/tests/pharos_state_machine.rs
```diff
@@ -0,0 +1,256 @@
...
+
+	let account_valid =
+		spv::verify_account_proof(&account_proof_nodes, &address_bytes, &raw_value, &state_root.0);
+	assert!(account_valid, "Account proof verification should pass");
+	println!("Account proof verification: PASSED");
+
+	assert!(!proof.storage_proof.is_empty(), "Should have at least one storage proof");
+	let storage_entry = &proof.storage_proof[0];
+	let storage_proof_nodes = rpc_to_proof_nodes(&storage_entry.proof);
+	let storage_hash = hex_to_h256(&proof.storage_hash).expect("Failed to parse storage_hash");
```
Wizdave97: Extract storage hash from verified account value

modules/pallets/testsuite/src/tests/pharos_state_machine.rs
```diff
@@ -0,0 +1,256 @@
...
+	let raw_value = hex_to_bytes(&proof_stake.raw_value).expect("Failed to parse raw_value");
+	let address_bytes: [u8; 20] = address.0;
+
+	let account_valid =
+		spv::verify_account_proof(&account_proof_nodes, &address_bytes, &raw_value, &state_root.0);
+	assert!(account_valid, "Account proof verification should pass");
+	println!("Account proof verification: PASSED");
+
+	let storage_hash =
+		hex_to_h256(&proof_stake.storage_hash).expect("Failed to parse storage_hash");
```
Wizdave97: Same thing

tesseract/consensus/pharos/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+		})?;
+
+		let latest_block = self.prover.get_latest_block().await?;
+		let update = self.prover.fetch_block_update(latest_block).await?;
+
+		Ok(Some(CreateConsensusState {
+			consensus_state: initial_consensus_state.encode(),
+			consensus_client_id: PHAROS_CONSENSUS_CLIENT_ID,
+			consensus_state_id: self.consensus_state_id,
+			unbonding_period: 60 * 60 * 27,
```
Wizdave97: Is this the pharos unstaking period?

tesseract/consensus/pharos/src/notification.rs
```diff
@@ -0,0 +1,94 @@
...
+	);
+
+	// Determine the target block for the update
+	// If we're still in the same epoch, just get the latest block
+	// If we've crossed epoch boundaries, we need to sync epoch by epoch
+	let target_block = if latest_epoch > current_epoch {
+		// Find the first epoch boundary block we need to sync
+		// Epoch boundary is at (epoch + 1) * epoch_length - 1
+		let next_epoch_boundary = (current_epoch + 1) * C::EPOCH_LENGTH_BLOCKS - 1;
+		if next_epoch_boundary <= latest_block {
```
Wizdave97: Don't think we need this if check,  next_epoch_boundary cannot be greater than latest_block since we've ascertained that it's in a different epoch.

modules/ismp/clients/pharos/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+
+		let trusted_state = VerifierState {
+			current_validator_set: consensus_state.current_validators.clone(),
+			finalized_block_number: consensus_state.finalized_height,
+			finalized_hash: consensus_state.finalized_hash,
+			current_epoch: consensus_state.current_epoch,
+		};
+
+		verify_pharos_block::<C, H>(trusted_state.clone(), update_1)
+			.map_err(|_| Error::Custom("Failed to verify first header".to_string()))?;
```
seunlanlege: we lose the underlying error at this point, lets add an `AnyHow` variant to `ismp::error::Error` so we can bubble up the underlying error, using `anyhow::Error::from`

modules/pallets/testsuite/src/tests/ismp_pharos.rs
```diff
@@ -0,0 +1,214 @@
...
+use ismp::{
+	consensus::{ConsensusClient, StateMachineId},
+	host::StateMachine,
+};
+use ismp_pharos::{ConsensusState, PharosClient, PHAROS_CONSENSUS_CLIENT_ID};
+use pharos_primitives::{Config, Testnet, PHAROS_ATLANTIC_CHAIN_ID};
+use pharos_prover::PharosProver;
+use primitive_types::H256;
+
+const ATLANTIC_RPC: &str = "https://atlantic-rpc.dplabs-internal.com";
```
seunlanlege: should be a secret env

modules/pallets/testsuite/src/tests/ismp_pharos.rs
```diff
@@ -0,0 +1,214 @@
...
+			assert_eq!(
+				new_state.current_epoch, initial_consensus_state.current_epoch,
+				"Epoch should not change for non-epoch-boundary blocks"
+			);
+
+			let state_id = StateMachineId {
+				state_id: StateMachine::Evm(PHAROS_ATLANTIC_CHAIN_ID),
+				consensus_state_id: PHAROS_CONSENSUS_CLIENT_ID,
+			};
+			assert!(commitments.contains_key(&state_id), "Should have state commitment");
```
seunlanlege, line 102: make assertions about height

modules/ismp/core/src/error.rs
```diff
@@ -214,4 +214,6 @@ pub enum Error {
 	},
 	/// Error decoding signature
 	SignatureDecodingFailed,
+	/// Anyhow error: {0}
+	AnyHow(String),
```
seunlanlege:
```suggestion
	AnyHow(anyhow::Error),
```

modules/ismp/clients/pharos/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+publish = false
+
+[dependencies]
+log = { workspace = true, default-features = false }
+codec = { workspace = true, default-features = false }
+scale-info = { version = "2.1.1", default-features = false, features = ["derive"] }
+
+ismp = { workspace = true }
+pharos-verifier = { path = "../../../consensus/pharos/verifier", default-features = false }
+pharos-primitives = { path = "../../../consensus/pharos/primitives", default-features = false }
```
seunlanlege: better to use workspace dependencies

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,217 @@
...
+	let key_hash = sha256(key);
+	if leaf_data[1..33] != key_hash {
+		return None;
+	}
+
+	// Extract value hash from the leaf
+	let mut value_hash = [0u8; 32];
+	value_hash.copy_from_slice(&leaf_data[33..65]);
+
+	// Walk bottom-up, hashing each node and checking the parent contains it
```
Wizdave97: We should use the verify_pharos_proof for this, this verification code is duplicated

Wizdave97: They are basically the same function

modules/consensus/pharos/verifier/src/state_proof.rs
```diff
@@ -0,0 +1,637 @@
...
+/// 3. Verify each storage value against its per-key proof path
+/// 4. Decode the verified storage values into a ValidatorSet
+pub fn verify_validator_set_proof<H: Keccak256 + Send + Sync>(
+	state_root: H256,
+	proof: &ValidatorSetProof,
+	epoch: u64,
+) -> Result<ValidatorSet, Error> {
+	// Verify the account proof against the state root using Pharos SPV
+	let address: [u8; 20] = STAKING_CONTRACT_ADDRESS.0 .0;
+	if !spv::verify_account_proof(
```
Wizdave97: I don't think we need an account proof since slot hashes can be verified against the state_root directly

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,217 @@
...
+/// Verify a storage proof for a single key.
+///
+/// `address` is the 20-byte contract address.
+/// `storage_key` is the 32-byte storage slot hash.
+/// `storage_value` is the 32-byte padded storage value.
+/// `storage_hash` is the storage trie root from the account proof.
+pub fn verify_storage_proof(
+	proof_nodes: &[PharosProofNode],
+	address: &[u8; 20],
+	storage_key: &[u8; 32],
```
seunlanlege:
```suggestion
	slot_hash: &[u8; 32],
```

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,217 @@
...
+	storage_value: &[u8; 32],
+	storage_hash: &[u8; 32],
```
seunlanlege:
```suggestion
	storage_root: &[u8; 32],
```

modules/ismp/clients/pharos/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+
+		let trusted_state = VerifierState {
+			current_validator_set: consensus_state.current_validators.clone(),
+			finalized_block_number: consensus_state.finalized_height,
+			finalized_hash: consensus_state.finalized_hash,
+			current_epoch: consensus_state.current_epoch,
+		};
+
+		let new_state = verify_pharos_block::<C, H>(trusted_state, update.clone())
+			.map_err(|e| Error::Custom(alloc::format!("Verification failed: {:?}", e)))?;
```
seunlanlege:
```suggestion
			.map_err(|e| Error::AnyHow(anyhow::Error::from(e)))?;
```

modules/ismp/clients/pharos/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+{
+	fn verify_consensus(
+		&self,
+		_host: &dyn IsmpHost,
+		consensus_state_id: ConsensusStateId,
+		trusted_consensus_state: Vec<u8>,
+		proof: Vec<u8>,
+	) -> Result<(Vec<u8>, ismp::consensus::VerifiedCommitments), Error> {
+		let update = VerifierStateUpdate::decode(&mut &proof[..])
+			.map_err(|_| Error::Custom("Cannot decode pharos client update".to_string()))?;
```
seunlanlege: lets also use anyhow::error::from here as well so we don't swallow errors

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	// Pharos uses a flat trie — storage proofs verify directly against state_root.
+	let mut map = BTreeMap::new();
+
+	for key in keys {
+		let (contract_addr, slot_hash) = if key.len() == 52 {
+			// First 20 bytes = contract address, last 32 = slot hash
+			let addr = H160::from_slice(&key[..20]);
+			let slot = H::keccak256(&key[20..]).0.to_vec();
```
seunlanlege: why do we hash the slot_hash?

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+			return Err(Error::Custom(
+				"Unsupported key type: expected length 20, 32, or 52".to_string(),
+			));
+		};
+
+		let contract_address: [u8; 20] = contract_addr.0;
+
+		// Look up the storage proof for this slot and verify it
+		if let Some(storage_proof_nodes) = pharos_proof.storage_proof.get(&slot_hash) {
+			if let Ok(slot_key) = <[u8; 32]>::try_from(slot_hash.as_slice()) {
```
seunlanlege: if the key is invalid, this should be an error.

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+					&state_root.0,
+				) {
+					map.insert(key, Some(value_hash.to_vec()));
+					continue;
+				}
+			}
+			// Proof verification failed - key not found or invalid
+			map.insert(key, None);
+		} else {
+			// No proof provided for this key
```
seunlanlege: this should also be an error

tesseract/consensus/pharos/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+
+pub use pharos_primitives::{Mainnet, Testnet};
+
+/// Host configuration for Pharos relayer
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PharosHostConfig {
+	/// Frequency (in seconds) to check for new updates
+	pub consensus_update_frequency: Option<u64>,
+	/// Pharos JSON-RPC URL
+	pub rpc_url: String,
```
seunlanlege: isn't this already in the evm config

modules/ismp/state-machines/pharos/src/lib.rs
```diff
@@ -196,24 +172,24 @@ pub fn verify_state_proof<H: Keccak256 + Send + Sync>(
...
+		if let Some(storage_proof_nodes) = pharos_proof.storage_proof.get(slot_key.as_slice()) {
+			if let Some(value_hash) = spv::verify_storage_membership_proof(
+				storage_proof_nodes,
+				&contract_address,
+				&slot_key,
+				&state_root.0,
+			) {
+				map.insert(key, Some(value_hash.to_vec()));
+			} else {
+				map.insert(key, None);
```
seunlanlege: i'm not convinced this is correct

modules/ismp/clients/pharos/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+{
+	fn verify_consensus(
+		&self,
+		_host: &dyn IsmpHost,
+		consensus_state_id: ConsensusStateId,
+		trusted_consensus_state: Vec<u8>,
+		proof: Vec<u8>,
+	) -> Result<(Vec<u8>, ismp::consensus::VerifiedCommitments), Error> {
+		let update = VerifierStateUpdate::decode(&mut &proof[..])
+			.map_err(|e| Error::AnyHow(anyhow::anyhow!("{:?}", e).into()))?;
```
seunlanlege, line 106: doesn't `parity_scale_code::Error` implement `core::Error`

dharjeezy: `parity_scale_codec::Error` only implements `std::error::Error` which is gated behind the `std` feature, not `core::error::Error` actually.

modules/consensus/pharos/primitives/src/types.rs
```diff
@@ -0,0 +1,271 @@
...
+///
+/// This contains the aggregated BLS signature for a block and the list
+/// of participating validators who signed it.
+#[derive(Debug, Clone, PartialEq, Eq, Default, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct BlockProof {
+	/// The aggregated BLS signature from participating validators (96 bytes)
+	pub aggregate_signature: Vec<u8>,
+	/// List of BLS public keys of validators who participated in signing
+	pub participant_keys: Vec<BlsPublicKey>,
```
Wizdave97, line 141: Let's use BtreeSet so there's no duplicates

modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -109,8 +120,14 @@ fn verify_validator_membership(
 	validator_set: &ValidatorSet,
 	participants: &[BlsPublicKey],
 ) -> Result<(), Error> {
-	if let Some(key) = participants.iter().find(|key| !validator_set.contains(key)) {
-		return Err(Error::UnknownValidator { key: key.clone() });
+	let mut seen = alloc::collections::BTreeSet::new();
+	for key in participants {
```
Wizdave97: YOu can just collect the participants into a Btreeset instead of doing this.

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+	if depth % 2 == 0 {
+		key_hash[byte_index] & 0x0F
+	} else {
+		(key_hash[byte_index] >> 4) & 0x0F
+	}
+}
+
+fn is_zero_slot(slot: &[u8]) -> bool {
+	slot.iter().all(|&b| b == 0)
+}
```
seunlanlege, line 90: Wait we're checking every byte in every hash if its zero?

Why not compare it to a zeroed out 32 byte array
```suggestion
fn is_zero_slot(slot: &[u8; 32]) -> bool {
	slot == &[0u8; 32]
}
```

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+	data.extend_from_slice(&proof_node[..INTERNAL_NODE_HEADER]);
+	for i in 0..INTERNAL_NODE_SLOTS {
+		let start = INTERNAL_NODE_HEADER + i * INTERNAL_NODE_SLOT_SIZE;
+		let slot = &proof_node[start..start + INTERNAL_NODE_SLOT_SIZE];
+		if !is_zero_slot(slot) {
+			data.extend_from_slice(slot);
+		}
+	}
+	sha256(&data)
+}
```
seunlanlege, line 110: This could be redesigned, we don't need two loops

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+/// Uses `nibble_at_depth(sha256(key))` to locate child slots in internal nodes
+/// (index > 0), ensuring the proof path follows the key's trie path. The MSU root
+/// (index 0) uses `next_begin_offset` because its 256-slot addressing scheme is
+/// Pharos-specific and opaque; the MSU root content is pinned to the state root
+/// via its hash, so an attacker cannot substitute a different MSU root.
+fn verify_proof_walk(
+	proof_nodes: &[PharosProofNode],
+	key: &[u8],
+	root: &[u8; 32],
+) -> bool {
```
seunlanlege: Prefer Result API

```suggestion
) -> Result<(), Error> {
```

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+	key
+}
+
+/// Verify existence of a key-value pair in the trie.
+pub fn verify_pharos_proof(
+	proof_nodes: &[PharosProofNode],
+	key: &[u8],
+	value: &[u8],
+	root: &[u8; 32],
+) -> bool {
```
seunlanlege: Result API here as well

```suggestion
) -> Result<(), Error> {
```

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+
+fn build_storage_key(address: &[u8; 20], slot_hash: &[u8; 32]) -> [u8; 52] {
+	let mut key = [0u8; 52];
+	key[..20].copy_from_slice(address);
+	key[20..].copy_from_slice(slot_hash);
+	key
+}
+
+/// Verify existence of a key-value pair in the trie.
+pub fn verify_pharos_proof(
```
seunlanlege:
```suggestion
pub fn verify_inclusion_proof(
```

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+}
+
+pub fn verify_account_proof(
+	proof_nodes: &[PharosProofNode],
+	address: &[u8; 20],
+	raw_value: &[u8],
+	state_root: &[u8; 32],
+) -> bool {
+	verify_pharos_proof(proof_nodes, address, raw_value, state_root)
+}
```
seunlanlege: dead code

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+	let key = build_storage_key(address, slot_hash);
+	verify_pharos_proof(proof_nodes, &key, storage_value, storage_root)
+}
+
+/// Verify key membership without the raw value. Returns `Some(value_hash)` from the leaf.
+pub fn verify_pharos_proof_membership(
+	proof_nodes: &[PharosProofNode],
+	key: &[u8],
+	root: &[u8; 32],
+) -> Option<[u8; 32]> {
```
seunlanlege: Result API

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+
+	let depth = proof_nodes.len().saturating_sub(2);
+	let nibble = nibble_at_depth(&key_hash, depth) as usize;
+	let slot_start = INTERNAL_NODE_HEADER + nibble * INTERNAL_NODE_SLOT_SIZE;
+
+	if slot_start + INTERNAL_NODE_SLOT_SIZE > last.len() ||
+		!is_zero_slot(&last[slot_start..slot_start + INTERNAL_NODE_SLOT_SIZE])
+	{
+		return false;
+	}
```
seunlanlege, line 261: what does this do?

modules/consensus/pharos/primitives/src/spv.rs
```diff
@@ -0,0 +1,612 @@
...
+		None => false,
+	}
+}
+
+/// Verify that `key` does NOT exist in the trie.
+///
+/// Case 1: Proof ends at a leaf with a different key_hash (path collision).
+/// Case 2: Proof ends at an internal node where the target nibble slot is empty.
+///         Sibling proofs pin the non-empty slots to the same root, preventing forgery.
+pub fn verify_non_existence_proof(
```
seunlanlege, line 232: I think this function can be further simplified

modules/consensus/pharos/primitives/src/types.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// Block proof containing the BLS signature data.
+///
+/// This contains the aggregated BLS signature for a block and the list
+/// of participating validators who signed it.
+#[derive(Debug, Clone, PartialEq, Eq, Default, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct BlockProof {
+	/// The aggregated BLS signature from participating validators (96 bytes)
+	pub aggregate_signature: Vec<u8>,
```
seunlanlege, line 139: we should use arkworks types here, Affine representation is best

modules/consensus/pharos/primitives/src/spv.rs
Wizdave97: In the second non-existence proof scenario, we want to confirm that the key terminates in a slot that's empty(zero slot), before we go ahead to verify the terminating node since empty slots are skipped when hashing the node.

We can use `last.get(slot_start..slot_start + INTERNAL_NODE_SLOT_SIZE)` so we don't need this extra check `slot_start + INTERNAL_NODE_SLOT_SIZE > last.len()`

modules/consensus/pharos/primitives/src/types.rs
Wizdave97: This is an rpc response, arkworks Affine type can't be deserialized from a hex string.

modules/ismp/clients/pharos/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+
+		let consensus_state = ConsensusState::decode(&mut &trusted_consensus_state[..])
+			.map_err(|_| Error::Custom("Cannot decode trusted consensus state".to_string()))?;
```
seunlanlege: we shouldn't swallow errors

PR #642, opened 2026-01-24
emit event after collator is rewarded

dharjeezy:
Introduces an event after a collator is rewarded, this is most specially needed for indexing

### DIFF
--- a/docs/content/developers/network/collator.mdx
+++ b/docs/content/developers/network/collator.mdx
@@ -286,19 +286,6 @@ At the beginning of each new session, our `pallet-collator-manager` runs its sel
 - It ranks the remaining candidates from highest to lowest based on their Controller's Reputation Asset balance.
 - It selects the top candidates to fill the available collator slots for the new session.
 
-### Technical Note: Validator Registration
-
-The stash/controller separation is enforced through a custom `ValidatorRegistration` implementation in `pallet-collator-manager`.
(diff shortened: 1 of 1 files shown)


docs/content/developers/network/collator.mdx
```diff
@@ -272,14 +278,27 @@ This tells the network: "I  bonded the Stash, designate this Controller to perfo
...
 - It ranks the remaining candidates from highest to lowest based on their Controller's Reputation Asset balance.
 - It selects the top candidates to fill the available collator slots for the new session.
 
+### Technical Note: Validator Registration
+
+The stash/controller separation is enforced through a custom `ValidatorRegistration` implementation in `pallet-collator-manager`.
+When `pallet-collator-selection` checks if a candidate is valid, it calls `ValidatorRegistration::is_registered(stash)`.
+Our implementation:
+
+1. Looks up the **Controller** account mapped to the given **Stash** via the `Controller` storage map
```
Wizdave97: I think we should just list the steps they should take before calling `registerCandidate`

dharjeezy: Yeah, so the existing doc already give guidelines on the step to take already.

dharjeezy: i can go ahead and remove the note since it is not needed

PR #647, opened 2026-01-30
update polkadot-sdk to v2512

dharjeezy:
This PR updates the repo to polkadot-sdk 2512.1.0

### DIFF
--- a/modules/utils/serde/Cargo.toml
+++ b/modules/utils/serde/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde-hex-utils"
-version = "0.2.0"
+version = "0.1.0"
 edition = "2021"
 authors = ["Polytope Labs <hello@polytope.technology>"]
 license = "Apache-2.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23786,7 +23786,7 @@ dependencies = [
 
 [[package]]
 name = "serde-hex-utils"
-version = "0.2.0"
+version = "0.1.0"
 dependencies = [
  "anyhow",
  "hex",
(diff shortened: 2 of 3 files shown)


modules/pallets/token-gateway/src/lib.rs
```diff
@@ -611,6 +611,30 @@ where
 							))?
 						}
 					},
+					MultiSignature::Eth(sig) => {
```
Wizdave97: We should not support this here, we already have ECDSA support, let's just return an error

parachain/runtimes/nexus/src/lib.rs
seunlanlege, line 1: Please add the pallet-intents-coprocessor to nexus in this PR

.github/workflows/ci.yml
```diff
@@ -10,23 +10,15 @@ on:
             - "**/Cargo.lock"
             - ".github/workflows/ci.yml"
 
-    pull_request_target:
```
Wizdave97, line 13: We can restore this

.github/workflows/ci.yml
```diff
@@ -10,23 +10,15 @@ on:
...
+    pull_request:
+        branches:
+            - main
         paths:
             - "**/*.rs"
             - "**/Cargo.toml"
             - "**/Cargo.lock"
             - ".github/workflows/ci.yml"
 
-    # pull_request:
```
Wizdave97, line 21: And this

modules/utils/serde/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "serde-hex-utils"
-version = "0.1.0"
+version = "0.2.0"
```
Wizdave97: This does not need to be bumped, only the crates whose dependencies have been updated need a bump

PR #655, opened 2026-02-14
[tesseract] Fix state queries for tron evm provider

Wizdave97:
Ethers uses "null" as the default parameter for the block number in eth_call, this is unsupported by tron, this PR updates eth_calls to use the latest block tag

Successful message submission: https://nile.tronscan.org/#/transaction/a0b3cdc57db1a86995bf68cc47616399e9e987dd4c07277c7930b1b912d7c7a1/event-logs

### DIFF
--- a/tesseract/messaging/evm/src/lib.rs
+++ b/tesseract/messaging/evm/src/lib.rs
@@ -268,8 +268,11 @@ impl EvmClient {
 		counterparty: Arc<dyn IsmpProvider>,
 	) -> Result<(), anyhow::Error> {
 		if self.config.initial_height.is_none() {
-			self.initial_height =
-				counterparty.query_latest_height(self.state_machine_id()).await?.into();
+			self.initial_height = counterparty
+				.query_latest_height(self.state_machine_id())
+				.await
+				.unwrap_or(self.initial_height as u32)
+				.into();
 		}
 
 		log::info!("Initialized height for {:?} at {}", self.state_machine, self.initial_height);


tesseract/messaging/evm/src/provider.rs
```diff
@@ -138,22 +155,63 @@ impl IsmpProvider for EvmClient {
...
-		let value =
-			contract.state_machine_commitment_update_time(height.try_into()?).call().await?;
-		Ok(Duration::from_secs(value.low_u64()))
+		let destination = ethers::types::H160(self.config.ismp_host.0);
+		let state_id = match height.id.state_id {
+			StateMachine::Polkadot(id) => U256::from(id),
+			StateMachine::Kusama(id) => U256::from(id),
+			StateMachine::Evm(id) => U256::from(id),
+			StateMachine::Substrate(id) => U256::from(u32::from_be_bytes(id)),
+			StateMachine::Tendermint(id) => U256::from(u32::from_be_bytes(id)),
```
seunlanlege: don't really use these variants

tesseract/messaging/evm/src/provider.rs
seunlanlege, line 1: lets override these methods in tesseract-tron/provider instead

tesseract/messaging/evm/src/provider.rs
Wizdave97: Yeah, I'll remove them

tesseract/messaging/evm/src/provider.rs
Wizdave97: Yeah, it won't work because some of them are used inside the state_machine_notification function, and the client referenced inside that function would be the evm client, not the tron client

Wizdave97: I can change the implementation of only the ones called inside that function.

PR #665, opened 2026-02-24
[evm]: Deploy IntentGatewayV2

seunlanlege:
- [x] Added support for partial fills
 - [x] Audit fixes 
 - [x] Deploy new addresses


evm/src/apps/ExtrinsicIntents.sol
```diff
@@ -0,0 +1,310 @@
+// Copyright (C) Polytope Labs Ltd.
```
royvardhan: `ExtrinsicIntents`, `IntrinsicIntents` and `IntentsBase` are already there in the `/apps/intentsv2` dir

evm/src/apps/intentsv2/IntrinsicIntents.sol
```diff
@@ -0,0 +1,171 @@
...
+                fillAmount = totalRequired;
+                uint256 dust = solverAmount - totalRequired;
+                if (order.output.call.length > 0) {
+                    protocolShare = dust;
+                } else {
+                    protocolShare = (dust * _params.surplusShareBps) / 10_000;
+                    beneficiaryShare = dust - protocolShare;
+                }
+            } else {
+                fillAmount = solverAmount > remaining ? remaining : solverAmount;
```
royvardhan, line 89: In this partial-fill continuation branch, if `solverAmount > remaining` we cap `fillAmount` to `remaining` and the extra amount is effectively ignored (no dust calculation). Is this intentional?

evm/test/VWAPOracleTest.sol
```diff
@@ -59,15 +59,15 @@ contract VWAPOracleTest is Test {
         VWAPOracle.TokenDecimalsUpdate[] memory updates = new VWAPOracle.TokenDecimalsUpdate[](0);
         vm.prank(user);
         vm.expectRevert(VWAPOracle.Unauthorized.selector);
-        oracle.init(host, updates);
+        oracle.init(host, address(this), updates);
```
royvardhan, line 62: Now that the `VWAPOracle` needs three items, the `InitVWAPOracle.s.sol` script needs an update.

royvardhan (commented):
lgtm

evm/src/apps/intentsv2/IntrinsicIntents.sol
Wizdave97: Yes it's intentional, we don't want to overcharge fillers. There could be race conditions for partial fills because multiple fill transactions could be sent at the same time. Depending on the order in which they are filled, the last fill might have a very large surplus remaining.

royvardhan: I am currently at this part in the sdk implementation right now and was going to ask exactly this. Thanks.

PR #682, opened 2026-03-10
[intent-coprocessor]: make `StorageDepositFee` an on chain storage value

dharjeezy:
Changes `StorageDepositFee` from being a constant to an on chain storage value


modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -79,10 +78,6 @@ pub mod pallet {
 		/// A currency implementation for handling storage deposits
 		type Currency: ReservableCurrency<Self::AccountId>;
 
-		/// The storage deposit fee per bid
```
Wizdave97, line 82: We should keep this, but the storage value should take precedence

PR #684, opened 2026-03-11
[intent-coprocessor]: Price Discovery Protocol

dharjeezy:
## Overview

This PR introduces a deposit-based price submission protocol for the intents coprocessor. The intents system needs on-chain price data for token pairs to function correctly. Rather than depending on external oracles, this protocol allows anyone to submit prices for governance-approved token pairs by putting up a one-time deposit. Subsequent updates to the same pair are free, and the deposit can be reclaimed through a two-phase withdrawal process.

The implementation spans the pallet, RPC, SDK, and simplex filler.

### Pallet

The `submit_pair_price` extrinsic accepts price submissions for governance-approved token pairs. Each submission contains one or more price entries specifying a base token amount range and the corresponding price (all encoded as U256 scaled by 10^18). On the first submission per (account, pair), a configurable deposit is reserved from the submitter's balance. Further updates to the same pair are free.

Deposits are withdrawn through a two-phase process via `withdraw_price_deposit`. The first call records the unlock block (current block + governance-configured lock duration in blocks). The second call, after the unlock block is reached, unreserves the tokens and removes the deposit record. Calling too early fails with `DepositStillLocked`. 

Prices are organized into time-based windows. When a window expires, existing price data is lazily cleared on the first new submission rather than in `on_initialize`.

Governance extrinsics control recognized token pairs, deposit amounts, lock duration, and window duration.

### RPC

The `intents_getPairPrices(pair_id)` endpoint returns all price entries for a given pair. On-chain U256 values (scaled by 10^18) are converted to human-readable decimal strings with fractional precision preserved (e.g. `1414500000000000000000` becomes `"1414.5"`).

### SDK

`IntentsCoprocessor` in `@hyperbridge/sdk` exposes `submitPairPrice(pairId, entries)` and `withdrawPriceDeposit(pairId)` for interacting with the pallet from TypeScript. The `signAndSendExtrinsic` helper checks for dispatch errors within the `isInBlock`/`isFinalized` status handler so that on-chain failures are correctly reported.

### Simplex filler integration

A `PriceUpdateService` in `@hyperbridge/simplex` handles periodic price submission on a configurable interval (default 5 minutes). It iterates through all configured pairs and calls `submitPairPrice` for each one. The `IntentFiller` starts this service automatically when a `[priceUpdates]` section is present in the TOML config, with human-readable amounts converted to 18-decimal format at parse time.


modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -144,6 +145,86 @@ pub struct Bid<AccountId> {
...
+	}
+}
+
+/// An individual price submission stored on-chain
+#[derive(Clone, Debug, Encode, Decode, DecodeWithMemTracking, TypeInfo, PartialEq, Eq)]
+pub struct PriceEntry<AccountId> {
+	/// The submitter's substrate account
+	pub submitter: AccountId,
+	/// The submitted price
+	pub price: U256,
```
Wizdave97, line 183: Let the price be a bounded string, so we don't have to worry about decimals

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -144,6 +145,86 @@ pub struct Bid<AccountId> {
...
+		data.extend_from_slice(&self.base.0);
+		data.extend_from_slice(&self.quote.0);
+		sp_io::hashing::keccak_256(&data).into()
+	}
+}
+
+/// An individual price submission stored on-chain
+#[derive(Clone, Debug, Encode, Decode, DecodeWithMemTracking, TypeInfo, PartialEq, Eq)]
+pub struct PriceEntry<AccountId> {
+	/// The submitter's substrate account
```
Wizdave97: Also, the price entry should specify a range denominated in the base token. Fillers would have different rates for different price ranges, e.g., for the pair USDC/CNGN `0 - 999 -> cNGN 1414`, `1000 - 5000  -> cNGN 1420`

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -144,6 +145,86 @@ pub struct Bid<AccountId> {
...
+	pub submitter: AccountId,
```
Wizdave97: Let's use the filler's evm address here

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -466,6 +723,195 @@ pub mod pallet {
...
+			// Extract the filler address from the proof value
+			// EVM addresses are 20 bytes, left-padded to 32 bytes in storage
+			let filler_address = H160::from_slice(
+				filler_bytes.get(12..32).ok_or(Error::<T>::MembershipProofFailed)?,
+			);
+			ensure!(filler_address != H160::zero(), Error::<T>::MembershipProofFailed);
+
+			// Check proof freshness
+			let threshold = ProofFreshnessThresholdValue::<T>::get();
+			let proof_gap = commitment_h2.timestamp.saturating_sub(commitment_h1.timestamp);
```
Wizdave97: To complete the validation that this commitment is fresh enough, we also need to do `ensure!(commitment_h2.timestamp.saturating_sub(now) <= threshold, Error::<T>::ProofNotFresh );`

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -466,6 +723,195 @@ pub mod pallet {
 			offchain_bid_key_raw(commitment, &filler.encode())
 		}
 
+		/// Process a verified (high confidence) price submission with ISMP proofs and
+		/// EVM signature verification.
+		fn submit_verified_price(
+			submitter: T::AccountId,
+			pair_id: H256,
```
Wizdave97:
```suggestion
			entries: BoundedVec<PriceEntry>,
```

modules/pallets/intents-coprocessor/src/types.rs
Wizdave97: You can leave this, we can assume that all prices are 18 decimals

modules/pallets/intents-coprocessor/src/types.rs
Wizdave97: Let's also assume 18 decimals for the price range here

docs/content/developers/intent-gateway/price-submission.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+### Verified Prices (High Confidence)
+
+Fillers who have actually filled orders on the IntentGateway contract submit prices along with cryptographic proof of their fill. This path exists because fillers are the most trustworthy source of price data. They have real skin in the game and their prices reflect actual market activity.
+
+A verified submission requires three things:
+
+1. **Two ISMP state proofs** against the IntentGateway contract's `_filled` mapping. A non-membership proof at height H1 shows the order was not yet filled, and a membership proof at height H2 shows it was filled. Together these bracket when the fill occurred and extract the filler's EVM address from storage.
+
+2. **Proof freshness.** Both the gap between H1 and H2, and the age of H2 relative to the current time, must fall within a governance-configured threshold. This prevents stale or manufactured proofs from being accepted.
```
Wizdave97:
```suggestion
2. **Proof freshness.** Both the gap between H1 and H2, and the age of H2 relative to the current time, must fall within a governance-configured threshold. This check ensures we are getting verified price data from fillers that are actually active and have filled orders very recently. Allowing us offer high confidence rates to users
```

modules/pallets/intents-coprocessor/rpc/src/lib.rs
```diff
@@ -259,6 +287,34 @@ where
...
+		};
+
+		// Decode Vec<PriceEntry>
+		// PriceEntry SCALE-encodes as (U256, U256, U256, u64)
+		type Entry = (primitive_types::U256, primitive_types::U256, primitive_types::U256, u64);
+		match Vec::<Entry>::decode(&mut &data[..]) {
+			Ok(entries) => Ok(entries
+				.into_iter()
+				.map(|(range_start, range_end, price, timestamp)| RpcPriceEntry {
+					range_start: range_start.to_string(),
```
Wizdave97: Let's divide the prices by 10**18 here to cnvert it to human readable value before returning

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +553,181 @@ pub mod pallet {
...
+			ensure!(!deposit_amount.is_zero(), Error::<T>::PriceDepositsNotConfigured);
+
+			let now = T::Dispatcher::default().timestamp().as_secs();
+
+			// Reserve deposit on first submission per (account, pair)
+			if !PriceDeposits::<T>::contains_key(&submitter, &pair_id) {
+				<T as Config>::Currency::reserve(&submitter, deposit_amount)
+					.map_err(|_| Error::<T>::InsufficientBalance)?;
+
+				PriceDeposits::<T>::insert(&submitter, &pair_id, (deposit_amount, now));
```
Wizdave97: The deposit lock duration is only relevant to withdrawals.

When a Filler wants to withdraw, they submit an extrinsic to initiate it, then we note the block at which the tokens can be unreserved using the configured pricelockduration, The filler can complete the withdrawal only after that block has elapsed

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +553,181 @@ pub mod pallet {
...
+		///
+		/// # Parameters
+		/// - `pair_id`: The token pair the deposit was made for
+		///
+		/// # Errors
+		/// - `DepositNotFound`: No deposit exists for this account and pair
+		/// - `DepositStillLocked`: The lock duration has not yet elapsed
+		#[pallet::call_index(13)]
+		#[pallet::weight(T::DbWeight::get().reads(3).saturating_add(T::DbWeight::get().writes(1)))]
+		pub fn withdraw_price_deposit(origin: OriginFor<T>, pair_id: H256) -> DispatchResult {
```
Wizdave97: This extrinsic should note the block at which the tokens can be unlocked on the first call

On the second call it should unreserve the tokens

modules/pallets/intents-coprocessor/rpc/src/lib.rs
Wizdave97: Preserve the decimals during this conversion

docs/content/developers/intent-gateway/price-submission.mdx
```diff
@@ -0,0 +1,144 @@
...
+
+Submissions are batched. Each call accepts up to `MaxPriceEntries` entries (a compile-time constant configurable per runtime), where each entry specifies a base token amount range and the corresponding price of the base token in terms of the quote token. This makes it possible to quote different rates for different order sizes in a single transaction. For example, a submitter might quote USDC/CNGN at 1414 for orders between 0 and 999, and 1420 for orders between 1000 and 5000.
+
+Each price entry contains three fields. The `range_start` field is the lower bound of the base token amount range (inclusive). The `range_end` field is the upper bound (also inclusive). The `price` field is the price of the base token in terms of the quote token. The pallet validates that `range_start` is less than or equal to `range_end` for every entry and rejects empty submissions. When stored on-chain, each entry becomes a `PriceEntry` that also includes the submission timestamp.
```
Wizdave97:
```suggestion
Each price entry contains three fields. The `range_start` field is the lower bound of the base token amount range (inclusive). The `range_end` field is the upper bound (also inclusive). The `price` field is the cost of one unit of the base token in terms of the quote token. The pallet validates that `range_start` is less than or equal to `range_end` for every entry and rejects empty submissions. When stored on-chain, each entry becomes a `PriceEntry` that also includes the submission timestamp.
```

docs/content/developers/intent-gateway/price-submission.mdx
```diff
@@ -0,0 +1,144 @@
...
+The simplex filler supports periodic price submission out of the box. When the filler starts, it checks whether a `[priceUpdates]` section is present in the configuration file. If price pairs are configured and a Hyperbridge WebSocket connection is available (via `substratePrivateKey` and `hyperbridgeWsUrl` in the `[simplex]` section), the filler creates a `PriceUpdateService` and starts it alongside the main order monitoring loop.
+
+The `PriceUpdateService` runs on a configurable interval (defaulting to every 5 minutes). On each tick, it iterates through all configured pairs and calls `IntentsCoprocessor.submitPairPrice()` for each one. An initial submission is also triggered shortly after startup so that prices are available immediately rather than waiting for the first interval to elapse. The service logs success or failure for each pair and continues with the remaining pairs even if one fails.
+
+### Configuration
+
+Price updates are configured by adding a `[priceUpdates]` section to the filler TOML configuration file.
+
+```toml
+[priceUpdates]
```
Wizdave97: We do not need this, the fx strategy config already has a price policy, we should just use the values specified there to submit price updates

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +568,206 @@ pub mod pallet {
...
+				entries.iter().all(|e| e.range_start <= e.range_end),
+				Error::<T>::InvalidPriceRange
+			);
+			ensure!(RecognizedPairs::<T>::contains_key(&pair_id), Error::<T>::PairNotRecognized);
+
+			let deposit_amount = PriceDepositAmount::<T>::get();
+			ensure!(!deposit_amount.is_zero(), Error::<T>::PriceDepositsNotConfigured);
+
+			let now = T::Dispatcher::default().timestamp().as_secs();
+
```
Wizdave97: They should not be able to submit a price update if they have initiated a withdrawal

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -144,6 +143,52 @@ pub struct Bid<AccountId> {
 	pub user_op: Vec<u8>,
 }
 
+/// A recognized token pair for price tracking
+#[derive(Clone, Debug, Encode, Decode, DecodeWithMemTracking, TypeInfo, PartialEq, Eq)]
+pub struct TokenPair {
+	/// The base token address
+	pub base: H160,
```
Wizdave97: Let's use padded H256 instead

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -144,6 +143,52 @@ pub struct Bid<AccountId> {
...
+	/// The quote token address
+	pub quote: H160,
```
Wizdave97: Same

parachain/simtests/src/price_submission.rs
```diff
@@ -0,0 +1,254 @@
...
+use polkadot_sdk::*;
+use primitive_types::{H160, H256, U256};
+use sc_consensus_manual_seal::CreatedBlock;
+use sp_core::{crypto::Ss58Codec, hashing::keccak_256, Bytes};
+use sp_keyring::sr25519::Keyring;
+use subxt::ext::subxt_rpcs::rpc_params;
+use subxt_utils::Hyperbridge;
+
+/// Compute the same pair_id as `TokenPair::pair_id()` — keccak256(base ++ quote).
+fn compute_pair_id(base: &H160, quote: &H160) -> H256 {
```
Wizdave97: This is a duplicate, method is already implemented on TokenPair

sdk/packages/sdk/src/chains/intentsCoprocessor.ts
```diff
@@ -261,11 +261,23 @@ export class IntentsCoprocessor {
 					if (result.status.isInBlock || result.status.isFinalized) {
 						resolved = true
 						clearTimeout(timeoutId)
-						resolve({
-							success: true,
-							blockHash: result.status.asInBlock.toHex() as HexString,
-							extrinsicHash: extrinsic.hash.toHex() as HexString,
-						})
+						const blockHash = result.status.isInBlock
```
Wizdave97:
```suggestion
						const txHash = result.status.isInBlock
```

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -431,6 +449,30 @@ program
 				logger.info("Rebalancing service initialized")
 			}
 
+			// Parse price update configuration
+			let priceUpdateConfig: PriceUpdateConfig | undefined
+			if (config.priceUpdates?.pairs && config.priceUpdates.pairs.length > 0) {
+				priceUpdateConfig = {
+					intervalSeconds: config.priceUpdates.intervalSeconds,
```
Wizdave97: For now price updates should only be submitted when the filler is initializing until we have a better strategy for updating prices in the filler config

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -52,6 +55,7 @@ export class IntentFiller {
 		privateKey: HexString,
 		rebalancingService?: RebalancingService,
 		bidStorage?: BidStorageService,
+		priceUpdateConfig?: PriceUpdateConfig,
```
Wizdave97: price update service should run as part of  the fx strategy

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -94,6 +101,8 @@ export class FXFiller implements FillerStrategy {
 		askPricePolicy: FillerPricePolicy,
 		maxOrderUsdStr: string,
 		exoticTokenAddresses: Record<string, HexString>,
+		askCurvePoints: PriceCurvePoint[],
```
Wizdave97: Why do we need to store these? We can compute them at the point of submitting prices

Also we should submit prices for asks and bids

sdk/packages/simplex/src/tests/strategies/fx.mainnet.test.ts
```diff
@@ -923,6 +925,11 @@ function createFxOnlyIntentFiller(
 	const extAsset = chainConfigService.getExtAsset(mainnetId)
 	const exoticTokenAddresses: Record<string, HexString> = extAsset ? { [mainnetId]: extAsset as HexString } : {}
 
+	const askCurvePoints = [
```
Wizdave97: We already have price policy that does this, please use what's available

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -119,6 +130,82 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub type PriceWindowStart<T: Config> = StorageValue<_, u64, ValueQuery>;
+
+	/// Price window duration in milliseconds
+	#[pallet::storage]
+	pub type PriceWindowDurationValue<T: Config> = StorageValue<_, u64, ValueQuery>;
+
+	/// Price entries per pair
+	#[pallet::storage]
+	pub type Prices<T: Config> = StorageMap<_, Blake2_128Concat, H256, Vec<PriceEntry>, ValueQuery>;
```
Wizdave97: Let's use a BtreeSet instead of a Vec
```suggestion
	pub type Prices<T: Config> = StorageMap<_, Blake2_128Concat, H256, BtreeSet<PriceEntry>, ValueQuery>;
```

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -466,6 +797,17 @@ pub mod pallet {
 			offchain_bid_key_raw(commitment, &filler.encode())
 		}
 
+		/// Clear all prices if this is the first submission in a new window.
+		///
+		/// Prices from the previous window persist until the first new submission
+		/// in the new window, at which point all entries across all pairs are cleared.
+		fn maybe_clear_stale_prices() {
+			if !PricesClearedThisWindow::<T>::get() {
+				let _ = Prices::<T>::clear(u32::MAX, None);
```
Wizdave97: We should only clear the price for the pair id that was submitted, not all pairs at once
`PricesClearedThisWindow` should probably be a with pair id as key

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -90,6 +91,8 @@ interface FxStrategyConfig {
 	maxOrderUsd: string
 	/** Map of chain identifier (e.g. "EVM-97") to exotic token contract address */
 	exoticTokenAddresses: Record<string, HexString>
+	/** On-chain pair ID (H256) for price submission to the intents coprocessor */
+	pairId?: HexString
```
Wizdave97:
```suggestion
```

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -64,6 +65,8 @@ export class FXFiller implements FillerStrategy {
 	private maxOrderUsd: Decimal
 	private account: ReturnType<typeof privateKeyToAccount>
 	private logger = getLogger("fx-simplex")
+	/** On-chain pair ID for price submission */
+	private pairId?: HexString
```
Wizdave97:
```suggestion
```

Wizdave97: Compute from the exotic token address and base token address

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -116,6 +122,65 @@ export class FXFiller implements FillerStrategy {
...
+		}
+	}
+
+	/**
+	 * Convert the ask curve points into on-chain PriceInput entries.
+	 * Each adjacent pair of points defines a price range.
+	 * The last point extends to a large upper bound.
+	 */
+	private buildPriceEntries(): PriceInput[] {
+		const points = this.askPricePolicy.getPoints()
```
Wizdave97, line 231: You should build prices for the bidPricePolicy which is the reverse case `Stable/Exotic` pair
ask us `Exotic/Stable` pair

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -111,6 +112,15 @@ export class IntentFiller {
 			}
 		}
 
+		// Submit initial prices on FX strategies during initialization
+		if (this.hyperbridge) {
+			for (const strategy of this.strategies) {
+				if (strategy instanceof FXFiller) {
+					await strategy.submitInitialPrices(this.hyperbridge)
```
Wizdave97, line 121: The submissions should  be done in the background

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -123,45 +118,80 @@ export class FXFiller implements FillerStrategy {
...
-	 * Submit initial prices using the ask price curve.
+	 * Compute pair ID from token symbols.
+	 * Hashes each symbol to H256, then hashes the concatenation to match
+	 * the pallet's TokenPair::pair_id() = keccak256(base_H256 || quote_H256).
+	 *
+	 * Governance registers pairs with base = keccak256(baseSymbol),
+	 * quote = keccak256(quoteSymbol), producing the same pair_id.
+	 */
+	private computeSymbolPairId(baseSymbol: string, quoteSymbol: string): HexString {
+		const base = keccak256(encodePacked(["string"], [baseSymbol]))
```
Wizdave97: let's just do keccak256(`${baseSymbol}/${quoteSymbol}`)

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -112,11 +112,13 @@ export class IntentFiller {
 			}
 		}
 
-		// Submit initial prices on FX strategies during initialization
+		// Submit initial prices on FX strategies in the background
 		if (this.hyperbridge) {
 			for (const strategy of this.strategies) {
 				if (strategy instanceof FXFiller) {
-					await strategy.submitInitialPrices(this.hyperbridge)
+					strategy.submitInitialPrices(this.hyperbridge).catch((err) => {
```
Wizdave97: On second thought we should reverse this, you can't submit transactions in parallel to hyperbridge.
Let's add some logs though

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -144,6 +143,54 @@ pub struct Bid<AccountId> {
...
+)]
+pub struct PriceEntry {
+	/// Lower bound of the base token amount range (inclusive), with 18 decimal places
+	pub range_start: U256,
+	/// Upper bound of the base token amount range (inclusive), with 18 decimal places
+	pub range_end: U256,
+	/// The price of the base token in the quote token, with 18 decimal places
+	pub price: U256,
+	/// Timestamp of submission (seconds)
+	pub timestamp: u64,
```
Wizdave97, line 185: Let's change this to filler address

modules/pallets/intents-coprocessor/rpc/src/lib.rs
```diff
@@ -259,6 +306,39 @@ where
...
+
+		// Decode Vec<PriceEntry>
+		// PriceEntry SCALE-encodes as (U256, U256, U256, H256)
+		type Entry = (
+			primitive_types::U256,
+			primitive_types::U256,
+			primitive_types::U256,
+			primitive_types::H256,
+		);
+		match Vec::<Entry>::decode(&mut &data[..]) {
```
Wizdave97: Why not just decode into BTreeSet<PriceEntry>, this would fail

modules/pallets/intents-coprocessor/rpc/src/lib.rs
```diff
@@ -259,6 +306,39 @@ where
...
+		let storage_key = sp_core::storage::StorageKey(key);
+
+		let data = match self.client.storage(best_hash, &storage_key) {
+			Ok(Some(data)) => data.0,
+			_ => return Ok(Vec::new()),
+		};
+
+		// Decode Vec<PriceEntry>
+		// PriceEntry SCALE-encodes as (U256, U256, U256, H256)
+		type Entry = (
```
Wizdave97: We don't need to define this type

parachain/simtests/src/price_submission.rs
```diff
@@ -0,0 +1,246 @@
...
+
+	// Deposit amount: 100 units
+	let deposit_amount: u128 = 100_000_000_000_000;
+
+	// Lock duration: 5 blocks
+	let lock_duration: u32 = 5;
+
+	// Price entries: 0-999 at 1414.5, 1000-5000 at 1420
+	let price_entries = vec![
+		PriceInput {
```
Wizdave97, line 100: This test has not yet been updated

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -116,6 +117,132 @@ export class FXFiller implements FillerStrategy {
...
+	 * - Ask pair (stable/exotic): submitted with askPricePolicy entries
+	 *   Ranges in stable amounts, price = exotic tokens per 1 stable
+	 * - Bid pair (exotic/stable): submitted with bidPricePolicy entries
+	 *   Ranges in exotic amounts, price = stable tokens per 1 exotic
+	 */
+	async submitInitialPrices(coprocessor: Promise<IntentsCoprocessor>): Promise<void> {
+		try {
+			const chain = Object.keys(this.exoticTokenAddresses)[0]
+			const exoticAddress = this.exoticTokenAddresses[chain]
+			const usdcAddress = this.configService.getUsdcAsset(chain)
```
Wizdave97: Let's add the stablecoin addresses to the fx strategy config like we have for exotic tokens, so when we support USDT, we can just call `getUsdToken(chain)`

modules/pallets/intents-coprocessor/src/benchmarking.rs
```diff
@@ -186,5 +188,147 @@ mod benchmarks {
...
+
+		#[extrinsic_call]
+		_(origin as T::RuntimeOrigin, 1000u32.into());
+
+		assert_eq!(StorageDepositFee::<T>::get(), 1000u32.into());
+		Ok(())
+	}
+
+	#[benchmark]
+	fn submit_pair_price() {
```
Wizdave97: This should take a Linear iterator over the number of price entries

Wizdave97 (commented):
Looks good.

docs/content/developers/intent-gateway/price-submission.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+### Configuration
+
+Price submission is enabled by adding a `pairId` to the `hyperfx` strategy in the filler TOML configuration file. The ask price curve points are converted into price entries automatically, where each point on the curve defines a price range.
+
+```toml
+[[strategies]]
+type = "hyperfx"
+pairId = "0x..."                    # On-chain pair ID (keccak256 of base_address ++ quote_address)
+priceSubmissionIntervalSeconds = 300 # How often to submit prices (default: 300 = 5 minutes)
```
Wizdave97:
```suggestion
```

docs/content/developers/intent-gateway/price-submission.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+[[strategies.bidPriceCurve]]
+amount = "0"
+price = "1500"
+
+[[strategies.bidPriceCurve]]
+amount = "1000"
+price = "1490"
+
+[strategies.exoticTokenAddresses]
```
Wizdave97, line 68: Let's update the simplex config in the docs with stablecoin addresses

sdk/packages/sdk/src/chains/intentsCoprocessor.ts
```diff
@@ -316,6 +328,59 @@ export class IntentsCoprocessor {
...
+	async withdrawPriceDeposit(pairId: HexString): Promise<BidSubmissionResult> {
+		try {
+			const extrinsic = this.api.tx.intentsCoprocessor.withdrawPriceDeposit(pairId)
+			return await this.signAndSendExtrinsic(extrinsic)
+		} catch (error) {
+			return {
+				success: false,
+				error: error instanceof Error ? error.message : "Unknown error",
+			}
+		}
```
Wizdave97: @dharjeezy  Let's add a method to this `getQuotes(pairid: HexString, amount: number) -> Quotes[]`,  this method will fetch all quotes, group them by filler address, run the linear interpolation found in price policy and return the quote for that amount from each filler.

sdk/packages/sdk/src/chains/intentsCoprocessor.ts
```diff
@@ -316,6 +335,124 @@ export class IntentsCoprocessor {
...
+			byFiller.set(entry.filler, points)
+		}
+
+		const quotes: Quote[] = []
+		for (const [filler, points] of byFiller) {
+			// Sort by amount ascending
+			points.sort((a, b) => a.amount - b.amount)
+
+			const interpolatedPrice = this.interpolatePrice(points, amount)
+			quotes.push({ filler, price: interpolatedPrice.toString() })
```
Wizdave97:
```suggestion
			quotes.push({ filler, amount: interpolatedPrice.toString() })
```

sdk/packages/sdk/src/chains/intentsCoprocessor.ts
```diff
@@ -316,6 +335,124 @@ export class IntentsCoprocessor {
...
+		return quotes
+	}
+
+	/**
+	 * Piecewise linear interpolation over sorted price points.
+	 * Below the minimum amount, returns the first point's price.
+	 * Above the maximum amount, returns the last point's price.
+	 * Between two points, linearly interpolates.
+	 */
+	private interpolatePrice(points: { amount: number; price: number }[], amount: number): number {
```
Wizdave97: Can we move the interpolation code from simplex to sdk to avoid code duplication

sdk/packages/sdk/src/types/index.ts
```diff
@@ -1350,6 +1350,24 @@ export type IntentOrderStatusUpdate =
...
+	amount: bigint
+	price: bigint
+}
+
+/** A quote from a single filler for a given amount, produced by interpolating their price entries */
+export interface Quote {
+	/** The filler's on-chain account address */
+	filler: string
+	/** The interpolated price at the requested amount */
+	price: string
```
Wizdave97:
```suggestion
	amount: string
```

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +568,181 @@ pub mod pallet {
...
+		/// lock duration via `withdraw_price_deposit`.
+		///
+		/// Each entry in `entries` specifies a base token amount threshold and the
+		/// corresponding price of the base token in terms of the quote token.
+		#[pallet::call_index(7)]
+		#[pallet::weight(T::WeightInfo::submit_pair_price(entries.len() as u32))]
+		pub fn submit_pair_price(
+			origin: OriginFor<T>,
+			pair_id: H256,
+			side: Side,
```
Wizdave97: This is not necessary, the pair ID already makes this distinction
```suggestion
```

Wizdave97 (commented):
Much Simpler @seunlanlege take another look.

sdk/packages/sdk/src/chains/intentsCoprocessor.ts
```diff
@@ -316,6 +336,75 @@ export class IntentsCoprocessor {
...
+	 * Fetches all price entries for a token pair, groups them by filler,
+	 * runs piecewise linear interpolation on each filler's price curve,
+	 * and returns the interpolated quote at the requested amount from each filler.
+	 *
+	 * @param pairId - The token pair identifier (H256 / bytes32)
+	 * @param amount - The base token amount to get quotes for (human-readable number, e.g. 500 for 500 tokens)
+	 * @returns Array of Quote objects, one per filler who has price entries for this pair
+	 */
+	async getQuotes(pairId: HexString, amount: number): Promise<Quote[]> {
+		const entries: RpcPriceEntry[] = await (this.api as any)._rpcCore.provider.send(
```
Wizdave97, line 382: We should filter out price entries that are older than 24 hours

seunlanlege:
Lets put this on ice for now

PR #696, opened 2026-03-12
[simplex,sdk]: support cross-chain fx, optimize gas estimation

royvardhan:
Fx strategy now supports cross-chain swaps.


sdk/packages/simplex/src/strategies/fx.ts
seunlanlege, line 1: Whats the confirmation policy? Also please update the docs in tandem

royvardhan: Add the confirmation policy, will wait for #697 to be merged, so that I can pull updated docs from there, and reupdate.

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -387,20 +389,24 @@ program
...
-							strategyConfig.maxOrderUsd,
-							strategyConfig.exoticTokenAddresses,
-						)
-					}
+				case "hyperfx": {
+					const bidPricePolicy = new FillerPricePolicy({ points: strategyConfig.bidPriceCurve })
+					const askPricePolicy = new FillerPricePolicy({ points: strategyConfig.askPriceCurve })
+					const fxConfirmationPolicy = strategyConfig.confirmationPolicies
+						? new ConfirmationPolicy(strategyConfig.confirmationPolicies)
+						: undefined
```
seunlanlege, line 397: confirmation policy is required

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -315,17 +311,19 @@ export class IntentFiller {
...
-					requiredConfirmations = Math.max(
-						requiredConfirmations,
-						strategy.confirmationPolicy.getConfirmationBlocks(
-							getChainId(order.source)!,
-							inputUsdValue.toNumber(),
-						),
-					)
+				if (isCrossChain) {
+					for (const [strategy, canFill] of canFillCache) {
+						if (!canFill || !strategy.confirmationPolicy) continue
```
Wizdave97, line 328: We should skip the crosschain order if the confirmation policy is not configured and add a log

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -387,20 +389,24 @@ program
...
-							bidPricePolicy,
-							askPricePolicy,
-							strategyConfig.maxOrderUsd,
-							strategyConfig.exoticTokenAddresses,
-						)
-					}
+				case "hyperfx": {
+					const bidPricePolicy = new FillerPricePolicy({ points: strategyConfig.bidPriceCurve })
+					const askPricePolicy = new FillerPricePolicy({ points: strategyConfig.askPriceCurve })
+					const fxConfirmationPolicy = strategyConfig.confirmationPolicies
```
Wizdave97, line 395: We should add a log that crosschain orders will be skipped without a confirmation policy

sdk/packages/simplex/src/bin/simplex.ts
royvardhan: It is kept optional because the filler may run this only for same chain orders as well. 

Filler with no confirmation policy will see logs for a cross chain order that is being skipped,

docs/content/developers/evm/filler.mdx
seunlanlege, line 1: we should delete this file, it doesn't show up in the docs

PR #706, opened 2026-03-17
[sdk]: Fix order placement test

ddboy19912:


sdk/packages/simplex/src/tests/strategies/basic.testnet.test.ts
```diff
@@ -263,14 +263,7 @@ describe("Filler V2 - Solver Selection ON", () => {
...
 
-		for await (const status of gen) {
-			if (status.status === "BID_SELECTED") {
-				console.log(`Bid selected: solver=${status.selectedSolver}, userOpHash=${status.userOpHash}`)
-			}
-			if (status.status === "FAILED") {
-				throw new Error(`Order execution failed: ${status.error}`)
-			}
-		}
+		await gen.return(undefined)
```
Wizdave97, line 266: Why this change?

ddboy19912: the test is failing on main

Screenshot 2026-03-17 at 10 44 35

the test is only focused on order placement  so this closes the generator right after `ORDER_PLACED` else it will keep failing

seunlanlege: I don't think this is the right fix either

Wizdave97: Remove this test and enable `Should place order, filler submits bid, user selects bid, order filled` instead, but use USDH(the testnet fee token) as the token to be filled instead of USDC

PR #707, opened 2026-03-17
[simplex,sdk]: support signing with MPC Vault

### LINKED ISSUE #699: [simplex]: External Wallet Provider Support
We need to implement the following wallet services for signing transactions in Simplex:

- [x] MPC vault https://docs.mpcvault.com/api-reference/overview
- [ ] Blockradar https://docs.blockradar.co/en/essentials/wallets

royvardhan:
Closes #699 

Todos: 
- [x] MPC Vault
- [x] Rewire Rebalancers to work with Account
- [x] Tests

On Standby:
- Blockradar

### DIFF
--- a/sdk/packages/simplex/src/services/DelegationService.ts
+++ b/sdk/packages/simplex/src/services/DelegationService.ts
@@ -1,5 +1,5 @@
 import type { HexString } from "@hyperbridge/sdk"
-import { concat, keccak256, serializeTransaction, toHex, toRlp, zeroAddress } from "viem"
+import { concat, keccak256, toHex, toRlp, zeroAddress } from "viem"
 import { ChainClientManager } from "./ChainClientManager"
 import { FillerConfigService } from "./FillerConfigService"
 import { getLogger } from "./Logger"
@@ -77,47 +77,14 @@ export class DelegationService {
 		const walletClient = this.clientManager.getWalletClient(chain)
 		const publicClient = this.clientManager.getPublicClient(chain)
 
-		if (this.signer.mode === "mpcVault") {
-			const chainNonce = await publicClient.getTransactionCount({
-				address: authorityAddress,
-				blockTag: "pending",
-			})
-
-			const txRequest = await walletClient.prepareTransactionRequest({
-				to: authorityAddress,
-				value: 0n,
-				authorizationList: [authorization],
-				chain: walletClient.chain,
-				nonce: chainNonce,
-			})
-
-			const numericChainId = txRequest.chainId ?? this.configService.getChainId(chain)
-
-			const txForSerialization = {
-				...txRequest,
-				type: "eip7702" as const,
-				chainId: numericChainId,
-				authorizationList: [authorization],
-				nonce: chainNonce,
-				gas: DELEGATION_TX_GAS_FLOOR,
-			}
-
-			const unsignedSerialized = serializeTransaction(txForSerialization)
-			const txSigningHash = keccak256(unsignedSerialized)
-			const signature = await this.signer.signRawHash(txSigningHash as HexString)
-			const signedSerialized = serializeTransaction(txForSerialization, signature)
-
-			return (await publicClient.sendRawTransaction({
-				serializedTransaction: signedSerialized,
-			})) as HexString
-		}
-
-		return (await walletClient.sendTransaction({
-			to: authorityAddress,
-			value: 0n,
-			authorizationList: [authorization],
-			chain: walletClient.chain,
-		})) as HexString
+		return this.signer.sendEip7702DelegationTransaction({
+			walletClient,
+			publicClient,
+			authorityAddress,
+			authorization,
+			chainIdFallback: this.configService.getChainId(chain),
+			gasFloor: DELEGATION_TX_GAS_FLOOR,
+		})
 	}
 
 	/**
--- a/sdk/packages/simplex/src/services/wallet/signer.ts
+++ b/sdk/packages/simplex/src/services/wallet/signer.ts
@@ -1,68 +1,17 @@
-import type { HexString } from "@hyperbridge/sdk"
-import { privateKeyToAccount } from "viem/accounts"
-import { sign } from "viem/accounts"
-import { createMpcVaultAccount } from "./mpcvault"
 import { SignerType, type SignerConfig, type SigningAccount } from "./types"
+import { createMpcVaultSigningAccount } from "./accounts/mpc"
+import { createPrivateKeySigningAccount } from "./accounts/privatekey"
 
 export function createSimplexSigner(config: SignerConfig): SigningAccount {
 	if (config.type === SignerType.PrivateKey) {
-		const account = privateKeyToAccount(config.privateKey)
-		const signRawHash = async (hash: HexString) => {
-			const signature = await sign({
-				hash,
-				privateKey: config.privateKey,
-			})
-			const yParity =
-				signature.yParity ??
-				(signature.v !== undefined ? Number(signature.v >= 27n ? signature.v - 27n : signature.v) : undefined)
-			if (yParity !== 0 && yParity !== 1) {
-				throw new Error("Failed to derive yParity from private key signature")
-			}
-			return {
-				r: signature.r as HexString,
-				s: signature.s as HexString,
-				yParity,
-			}
-		}
-		return {
-			mode: "privateKey",
-			account,
-			signMessage: (messageHash: HexString, _chainId: number) =>
-				account.signMessage({ message: { raw: messageHash } }),
-			signRawHash,
-		}
+		return createPrivateKeySigningAccount(config.privateKey)
 	}
 
 	if (config.type === SignerType.MpcVault) {
-		const { account, service } = createMpcVaultAccount(config.mpcVault)
-		const signRawHash = async (hash: HexString) => {
-			const signature = await service.signRawHash(hash)
-			return parseSignature(signature)
-		}
-		return {
-			mode: "mpcVault",
-			account,
-			signMessage: (messageHash: HexString, chainId: number) => service.signPersonalMessage(messageHash, chainId),
-			signRawHash,
-		}
+		return createMpcVaultSigningAccount(config.mpcVault)
 	}
 
 	throw new Error(`Unsupported signer mode: ${(config as { type?: string }).type ?? "unknown"}`)
-}
-
-function parseSignature(signature: HexString): { r: HexString; s: HexString; yParity: number } {
-	const hex = signature.slice(2)
-	if (hex.length !== 130) {
-		throw new Error(`Invalid signature length: expected 65 bytes, got ${hex.length / 2} bytes`)
-	}
-	const r = `0x${hex.slice(0, 64)}` as HexString
-	const s = `0x${hex.slice(64, 128)}` as HexString
-	const v = Number.parseInt(hex.slice(128, 130), 16)
-	const yParity = v >= 27 ? v - 27 : v
-	if (yParity !== 0 && yParity !== 1) {
-		throw new Error(`Invalid signature v/yParity value: ${v}`)
-	}
-	return { r, s, yParity }
 }
 
 export function validateSignerConfig(config: SignerConfig): void {
--- a/sdk/packages/simplex/src/services/wallet/accounts/mpc.ts
+++ b/sdk/packages/simplex/src/services/wallet/accounts/mpc.ts
@@ -1,0 +1,49 @@
+import type { HexString } from "@hyperbridge/sdk"
+import { keccak256, serializeTransaction } from "viem"
+import { createMpcVaultAccount } from "../mpcvault"
+import type { MpcVaultSignerConfig, SigningAccount } from "../types"
+
+export function createMpcVaultSigningAccount(config: MpcVaultSignerConfig): SigningAccount {
+	const { account, service } = createMpcVaultAccount(config)
+	const signRawHash = (hash: HexString) => service.signRawHashComponents(hash)
+	return {
+		mode: "mpcVault",
+		account,
+		signMessage: (messageHash: HexString, chainId: number) => service.signPersonalMessage(messageHash, chainId),
+		signRawHash,
+		sendEip7702DelegationTransaction: async (args) => {
+			const chainNonce = await args.publicClient.getTransactionCount({
+				address: args.authorityAddress,
+				blockTag: "pending",
+			})
+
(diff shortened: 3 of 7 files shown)


sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -237,9 +239,18 @@ interface BinanceConfig {
...
+	vaultUuid: string
+	accountAddress: HexString
+	callbackClientSignerPublicKey: string
+	baseUrl?: string
+}
+
 interface FillerTomlConfig {
 	simplex: {
-		privateKey: string
+		privateKey?: string
```
Wizdave97: Why not make the signer config an enum that can be either a private key or any of the external signers?

seunlanlege: yeah i think the signer should be a tagged union

sdk/packages/simplex/src/services/ChainClientManager.ts
```diff
@@ -58,14 +58,20 @@ export const ViemClientFactory = new ViemClientFactoryImpl()
...
+	private account: Account
 	private configService: FillerConfigService
 	private clientFactory: ViemClientFactoryImpl
 
-	constructor(configService: FillerConfigService, privateKey?: HexString) {
-		this.privateKey = privateKey || generatePrivateKey()
+	constructor(configService: FillerConfigService, accountOrPrivateKey?: Account | HexString) {
 		this.configService = configService
 		this.clientFactory = ViemClientFactory
+		if (!accountOrPrivateKey) {
```
Wizdave97: We should just make `accountOrPrivateKey` required

royvardhan: Roger that!

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -315,6 +316,7 @@ program
 			const fillerConfigForService: FillerServiceConfig = {
 				privateKey: config.simplex.privateKey,
 				mpcVault: config.simplex.mpcVault,
+				delegationSubmitterPrivateKey: config.simplex.delegationSubmitterPrivateKey,
```
seunlanlege: I don't think we need a new account

sdk/packages/simplex/src/services/DelegationService.ts
```diff
@@ -88,24 +125,29 @@ export class DelegationService {
...
+			const authorization = await this.buildAuthorization(chain, solverAccountContract)
 
 			const hash = await walletClient.sendTransaction({
-				account,
-				to: account.address,
+				account: submitter,
+				to: submitter.address,
 				value: 0n,
-				authorizationList: [authorization],
+				authorizationList: [authorization as never],
```
seunlanlege: whats wrong with the types?

sdk/packages/simplex/README.md
```diff
@@ -154,6 +155,27 @@ solverAccountContractAddress = "0x..."
...
+apiToken = "..."
+vaultUuid = "..."
+accountAddress = "0x..."
+callbackClientSignerPublicKey = "ssh-ed25519 AAAA..."
+```
+
+Notes:
+- `delegationSubmitterPrivateKey` is required when `simplex.mpcVault` is configured and the filler is not running in all-chain watch-only mode.
+- This submitter key is used for delegation setup/revoke transactions, while the delegation authorization itself is still signed by MPC Vault.
+
```
seunlanlege: I think this should go in docs.hyperbridge.network

sdk/packages/sdk/src/protocols/intents/BidManager.ts
```diff
@@ -93,9 +94,15 @@ export class BidManager {
...
-		const solverSignature = await solverAccount_.signMessage({ message: { raw: messageHash } })
+		let solverSignature: HexString
+		if (solverSignMessage) {
+			solverSignature = await solverSignMessage(messageHash)
+		} else if (solverPrivateKey) {
+			const solverAccount_ = privateKeyToAccount(solverPrivateKey as import("viem").Hex)
+			solverSignature = await solverAccount_.signMessage({ message: { raw: messageHash } })
+		} else {
+			throw new Error("Missing solver signer: provide solverSignMessage or solverPrivateKey")
+		}
```
seunlanlege: I think we should use polymorphic objects here, doesn't make sense that we update all callsites whenever we add a new signing scheme, which will be a lot

seunlanlege: aggregate all signing schemes, including raw private keys signing behind a polymorphic interface

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -303,6 +315,8 @@ program
 
 			const fillerConfigForService: FillerServiceConfig = {
 				privateKey: config.simplex.privateKey,
+				mpcVault: config.simplex.mpcVault,
```
seunlanlege: rather than passing privateKey and mpcVault, pass an abstract signing interface

sdk/packages/simplex/src/bin/simplex.ts
royvardhan: Delegation is a type 4 transaction. 

When an authorisarion list is submitted via viem, it automatically infers the transaction type.

While `evmSendCustom` in MPC Vault does not have an auth list field so it would become a type 2 transaction.

sdk/packages/simplex/src/services/DelegationService.ts
royvardhan: This was because I was infering when the signer and the submitter of the delegation are the same and auth list is not supported by MPC. 

Will remove the `never`

sdk/packages/sdk/src/protocols/intents/BidManager.ts
royvardhan: Agreed. Thanks.

sdk/packages/simplex/src/bin/simplex.ts
royvardhan: Makes sense.

sdk/packages/simplex/README.md
royvardhan: Shall I delete this file? It has some basic configuration data.

royvardhan: Or make it point to docs.hyperbridge.network

sdk/packages/simplex/src/bin/simplex.ts
seunlanlege: yeah but the mpc vault can sign the surrounding tx too

sdk/packages/simplex/README.md
seunlanlege: Yeah point to https://docs.hyperbridge.network/developers/intent-gateway/simplex/

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -356,12 +376,27 @@ program
 
 			// Create shared services to avoid duplicate RPC calls and reuse connections
 			const sharedCacheService = new CacheService()
-			const privateKey = config.simplex.privateKey as HexString
-			const chainClientManager = new ChainClientManager(configService, privateKey)
+			// Keep signer construction isolated from broader runtime config.
+			const signerConfig: SignerConfig | undefined =
```
Wizdave97: Let's have a signer service that's initialized here  directly from the `SimplexSignerTomlConfig`, it should encapsulate everything required for signing internally and expose all functionality via an interface

sdk/packages/simplex/src/services/rebalancers/binance.ts
```diff
@@ -113,12 +113,15 @@ export class BinanceRebalancer {
...
 		chainClientManager: ChainClientManager,
 		configService: FillerConfigService,
-		privateKey: HexString,
+		accountOrPrivateKey: Account | HexString,
 		config: BinanceCexConfig,
 	) {
 		this.chainClientManager = chainClientManager
 		this.configService = configService
-		this.privateKey = privateKey
+		this.account =
```
Wizdave97: Let this be provided by the signing service interface instead

sdk/packages/simplex/src/bin/simplex.ts
royvardhan: Done ser.

sdk/packages/sdk/src/protocols/intents/BidManager.ts
```diff
@@ -93,9 +94,19 @@ export class BidManager {
...
-		const solverAccount_ = privateKeyToAccount(solverPrivateKey as import("viem").Hex)
-		const solverSignature = await solverAccount_.signMessage({ message: { raw: messageHash } })
+		let solverSignature: HexString
+		switch (solverSigner.type) {
+			case SolverBidSignerType.External:
+				solverSignature = await solverSigner.signMessage(messageHash)
+				break
+			case SolverBidSignerType.PrivateKey: {
+				const solverAccount_ = privateKeyToAccount(solverSigner.privateKey as import("viem").Hex)
+				solverSignature = await solverAccount_.signMessage({ message: { raw: messageHash } })
```
seunlanlege: I don't think any internal subsystems should be exposed to privateKey. They should only interact with an abstract signer interface

sdk/packages/simplex/src/services/wallet/signer.ts
```diff
@@ -0,0 +1,130 @@
...
+import { MpcVaultService } from "./mpcvault"
+import { SignerType, type MpcVaultSignerConfig, type SignerConfig, type SigningAccount } from "./types"
+
+function requireChainId(value: unknown, context: string): number {
+	if (typeof value === "number" && Number.isFinite(value)) return value
+	if (typeof value === "bigint") return Number(value)
+	throw new Error(`Missing chainId for MPCVault ${context}`)
+}
+
+function createMpcVaultAccount(config: MpcVaultSignerConfig): { account: Account; service: MpcVaultService } {
```
seunlanlege: should this be here?

sdk/packages/sdk/src/types/index.ts
```diff
@@ -1164,11 +1164,20 @@ export interface PackedUserOperation {
 	signature: HexString
 }
 
+export interface SolverBidSigner {
```
Wizdave97: We should use one interface for all things related to signing

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -512,8 +513,24 @@ function validateConfig(config: FillerTomlConfig): void {
...
 
-	if (!config.simplex?.privateKey && !allChainsWatchOnly) {
-		throw new Error("Filler private key is required (unless all chains are in watchOnly mode)")
+	const signer = config.simplex?.signer
+
+	if (!signer && !allChainsWatchOnly) {
+		throw new Error("Signer configuration is required via [simplex.signer]")
+	}
+
+	if (signer?.type === SignerType.PrivateKey) {
```
Wizdave97: I think we can have a validateConfig on the simplexSignerConfig interface that does this instead of having it here

sdk/packages/sdk/src/types/index.ts
royvardhan: I did a better abstraction now

sdk/packages/simplex/src/services/wallet/signer.ts
```diff
@@ -0,0 +1,95 @@
...
+	if (yParity !== 0 && yParity !== 1) {
+		throw new Error(`Invalid signature v/yParity value: ${v}`)
+	}
+	return { r, s, yParity }
+}
+
+export function createSimplexSigner(config: SignerConfig): SigningAccount {
+	if (config.type === SignerType.PrivateKey) {
+		const account = privateKeyToAccount(config.privateKey)
+		const signRawHash = async (hash: HexString) => {
```
Wizdave97: Why not have a PrivateKeySigner type that implements this, then we just initialize that type here

sdk/packages/simplex/src/services/wallet/signer.ts
```diff
@@ -0,0 +1,95 @@
...
+		account,
+		signMessage: (messageHash: HexString, _chainId: number) =>
+			account.signMessage({ message: { raw: messageHash } }),
+		signRawHash,
+	}
+	}
+
+	if (config.type === SignerType.MpcVault) {
+		const { account, service } = createMpcVaultAccount(config.mpcVault)
+		const signRawHash = async (hash: HexString) => {
```
Wizdave97: Let's implement this directly on  the mpc vault type

sdk/packages/simplex/src/services/ChainClientManager.ts
```diff
@@ -58,30 +73,43 @@ export const ViemClientFactory = new ViemClientFactoryImpl()
...
+	private signer: SigningAccount
 	private configService: FillerConfigService
 	private clientFactory: ViemClientFactoryImpl
 
-	constructor(configService: FillerConfigService, privateKey?: HexString) {
-		this.privateKey = privateKey || generatePrivateKey()
+	constructor(configService: FillerConfigService, signer?: SigningAccount) {
 		this.configService = configService
 		this.clientFactory = ViemClientFactory
+		this.signer =
```
Wizdave97, line 83: Why is the signer optional in the constructor? Seems like it should not be

sdk/packages/simplex/src/services/DelegationService.ts
```diff
@@ -1,26 +1,123 @@
...
+			r: HexString
+			s: HexString
+			yParity: number
+		},
+	): Promise<HexString> {
+		const authorityAddress = this.signer.account.address
+		const walletClient = this.clientManager.getWalletClient(chain)
+		const publicClient = this.clientManager.getPublicClient(chain)
+
+		if (this.signer.mode === "mpcVault") {
```
Wizdave97: Seems this would work better once we add more signers
```suggestion
		if (this.signer.mode !== "privateKey") {
```

seunlanlege: I think we should use our abstract signer here

sdk/packages/simplex/src/services/ChainClientManager.ts
royvardhan: it is kept optional so that this can be run for watch only mode as well for any chain.

sdk/packages/simplex/src/services/wallet/signer.ts
royvardhan: Done

PR #713, opened 2026-03-17
[sdk]: estimate order in source native

royvardhan:
We should always use source native value for estimates, works for both same chain and cross chain orders.

### DIFF
--- a/sdk/packages/simplex/src/tests/strategies/basic.testnet.test.ts
+++ b/sdk/packages/simplex/src/tests/strategies/basic.testnet.test.ts
@@ -263,15 +263,6 @@ describe("Filler V2 - Solver Selection ON", () => {
 			console.log(`Order placed successfully with ID: ${order.id}`)
 		}
 
-		for await (const status of gen) {
-			if (status.status === "BID_SELECTED") {
-				console.log(`Bid selected: solver=${status.selectedSolver}, userOpHash=${status.userOpHash}`)
-			}
-			if (status.status === "FAILED") {
-				throw new Error(`Order execution failed: ${status.error}`)
-			}
-		}
-
 		expect(order.id).toBeDefined()
 		expect(order.user).toBe(bytes20ToBytes32(beneficiaryAddress))
 		expect(order.source).toBe(toHex(bscChapelId))


sdk/packages/simplex/src/tests/strategies/basic.testnet.test.ts
```diff
@@ -263,15 +263,6 @@ describe("Filler V2 - Solver Selection ON", () => {
 			console.log(`Order placed successfully with ID: ${order.id}`)
 		}
 
-		for await (const status of gen) {
```
seunlanlege, line 266: Why remove

sdk/packages/sdk/src/protocols/intents/utils.ts
```diff
@@ -219,3 +219,50 @@ export async function convertGasToFeeToken(
...
+	} catch {
+		const nativeCurrency = client.chain?.nativeCurrency
+		const chainId = Number.parseInt(evmChainID.split("-")[1])
+		const feeTokenAmountInToken = new Decimal(formatUnits(feeTokenAmount, feeToken.decimals))
+		const nativeTokenPriceUsd = await fetchPrice(nativeCurrency?.symbol, chainId)
+		const feeTokenAmountUsd = feeTokenAmountInToken.times(new Decimal(1))
+		const nativeAmount = feeTokenAmountUsd.dividedBy(nativeTokenPriceUsd)
+		return parseUnits(nativeAmount.toFixed(nativeCurrency?.decimals ?? 18), nativeCurrency?.decimals ?? 18)
+	}
+}
```
seunlanlege, line 268: Which is faster?

sdk/packages/simplex/src/tests/strategies/basic.testnet.test.ts
royvardhan: This was always meant to be “should place order only” test.

sdk/packages/sdk/src/protocols/intents/utils.ts
royvardhan: The catch block runs on testnets only.

seunlanlege (approved):

PR #717, opened 2026-03-18
[sdk]: thread relayerFee into cancel order calldata

ddboy19912:
This PR fixes cross-chain cancellation in the SDK by quoting and passing the correct relayer fee during cancel submission. Before this, some cancel transactions were built with a zero relayer fee even when one was required, which could break cancellation and lead to confusing wallet fee errors. The update adds a simple cancel quote return shape with both native value and relayer fee, uses that fee in the cancel calldata, and exposes it publicly so integrators can prepare approvals and submit cancels correctly.
Screenshot 2026-03-18 at 15 40 01

### DIFF
--- a/sdk/packages/sdk/src/protocols/intents/IntentGateway.ts
+++ b/sdk/packages/sdk/src/protocols/intents/IntentGateway.ts
@@ -267,20 +267,6 @@ export class IntentGateway {
 	}
 
 	/**
-	 * Quotes the native token cost for cancelling an order.
-	 *
-	 * Delegates to {@link OrderCanceller.quoteCancelNative}.
-	 *
-	 * @param order - The order to quote cancellation for.
-	 * @param fromDest - If `true`, quotes the destination-initiated cancellation fee.
-	 *   Defaults to `false` (source-side cancellation).
-	 * @returns The native token amount required to submit the cancel transaction.
-	 */
-	async quoteCancelNative(order: Order, fromDest: boolean = false): Promise<bigint> {
-		return this.orderCanceller.quoteCancelNative(order, fromDest)
-	}
-
-	/**
 	 * Returns both the native token cost and the relayer fee for cancelling an
 	 * order. Use `relayerFee` to approve the ERC-20 spend before submitting.
 	 *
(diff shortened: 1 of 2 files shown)


sdk/packages/sdk/src/protocols/intents/IntentGateway.ts
royvardhan: Do you need this anymore? We should keep one cancel quote function.

sdk/packages/sdk/src/protocols/intents/OrderCanceller.ts
```diff
@@ -62,10 +62,25 @@ export class OrderCanceller {
 	 * @returns The native token amount required to submit the cancel transaction.
 	 */
 	async quoteCancelNative(order: Order, fromDest: boolean = false): Promise<bigint> {
```
royvardhan: Same

sdk/packages/sdk/src/protocols/intents/IntentGateway.ts
ddboy19912: Done

sdk/packages/sdk/src/protocols/intents/OrderCanceller.ts
ddboy19912: Done

sdk/packages/sdk/src/protocols/intents/OrderCanceller.ts
```diff
@@ -109,7 +111,11 @@ export class OrderCanceller {
 			context,
 		}
 
-		return await this.ctx.source.quoteNative(getRequest, 0n)
+		const feeInSourceFeeToken = await convertGasToFeeToken(this.ctx, 400_000n, "source", sourceStateMachine)
```
Wizdave97, line 114: Where does 400_000n come from?

ddboy19912: https://github.com/polytope-labs/hyperbridge/blob/abf43ed567f79f801762ef0449d176330cba2257/sdk/packages/sdk/src/protocols/intents/OrderCanceller.ts#L592-L594

PR #768, opened 2026-04-10
[evm, sdk, simplex]: add SimplexPaymaster

royvardhan:
**Contract**: Added `SimplexPaymaster.sol` — onchain ERC-4337 v0.8 paymaster accepting USDC/USDT via Chainlink oracles, with `PERMIT` and `APPROVE` modes.                                                      
   
**Deployment**: Added `DeploySimplexPaymaster.s.sol` and paymaster config entries to `config.mainnet.toml`.                                      
                                                                                    
**SDK**: Abstracted paymaster module `(src/services/paymaster/)` with unified `buildPaymasterAndData()` entry point. Prefers Circle Paymaster (USDC permit) where available and filler has USDC balance, falls back to `SimplexPaymaster` (USDC/USDT, permit or capped approval).                            

**Simplex**: `DelegationService` now delegates via bundler `UserOp` with paymaster paying gas — solver needs zero native tokens. Supports Pimlico, Alchemy, and generic bundler gas pricing. Falls back to direct tx. `ContractInteractionService` approval scoped to exact required amount instead of `maxUint256`.


evm/src/utils/SimplexPaymaster.sol
```diff
@@ -0,0 +1,376 @@
...
+        return super._validatePaymasterUserOp(userOp, userOpHash, maxCost);
+    }
+
+    /// @dev Parse and execute EIP-2612 permit from paymasterData.
+    ///      Layout: mode(1) + token(20) + permitAmount(32) + deadline(32) + v(1) + r(32) + s(32)
+    ///      Total:  150 bytes
+    function _executePermit(PackedUserOperation calldata userOp) internal {
+        bytes calldata data = userOp.paymasterData();
+        // Minimum length: 1 + 20 + 32 + 32 + 1 + 32 + 32 = 150
+        require(data.length >= 150, "SimplexPaymaster: permit data too short");
```
Wizdave97: Can't we just use equalto?
```suggestion
        require(data.length == 150, "SimplexPaymaster: permit data too short");
```

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -0,0 +1,376 @@
...
+        try IERC20Permit(tokenAddr).permit(owner, address(this), permitAmount, deadline, v, r, s) {
+            emit PermitExecuted(tokenAddr, owner, permitAmount);
+        } catch {
+            revert PermitFailed(tokenAddr);
+        }
+    }
+
+    // ── Oracle helpers ───────────────────────────────────────────────
+
+    /// @dev Fetch price from a Chainlink feed, normalized to 8 decimals.
```
Wizdave97, line 317: Why normalize 8 instead of 18 decimals

royvardhan: The oracle already returns the price in 8 decimals.

sdk/packages/simplex/src/services/DelegationService.ts
```diff
@@ -131,12 +137,141 @@ export class DelegationService {
...
+				solverAccount,
+				paymasterAddress,
+				chain,
+				this.configService,
+				true, // forceApproveMode
+			)
+			const paymasterAndData = packPaymasterAndData(pm)
+
+			// 3. Get gas prices from bundler
+			const gasPriceResult = await this.sendBundlerRpc<{
```
Wizdave97: This would fail when they use the alchemy bundler which is now our default bundler

evm/src/utils/SimplexPaymaster.sol
Wizdave97: I think 18 decimals is better since we'll still have to do some math with native token gas

sdk/packages/simplex/src/services/DelegationService.ts
Wizdave97: It should support both alchemy and pimlico

Wizdave97:
@royvardhan  let's default to the circle paymaster when it's available and when the user has USDC balance. Our custom paymaster should only be used in cases where there's no circle paymaster or the user

evm/src/utils/SimplexPaymaster.sol
royvardhan: Claude interrupted with this and it makes sense: 

  

> Both nativeUsd and tokenUsd are normalized to the same decimal precision, so they 
>   cancel out in the ratio. Whether that's 8 or 18 doesn't matter — the division     
>   nativeUsd / tokenUsd gives the same result regardless of the shared normalization 
>   factor.                                                                           
>                                                                                     
>   For example with ETH=$3000, USDC=$1, tokenDecimals=6:                             
>   - At 8 decimals: (3000e8 * 1e6) / 1e8 = 3000 * 1e6 = 3e9                          
>   - At 18 decimals: (3000e18 * 1e6) / 1e18 = 3000 * 1e6 = 3e9
>                                                                                     
>   Same result. The shared normalization cancels.
>                                                                                     
>   8 decimals is actually preferable because:                
>   1. Chainlink feeds natively report 8 decimals for USD pairs — normalizing to 8    
>   means most feeds need zero scaling                                            
>   2. Using 18 would mean multiplying by 10^10 for most feeds, creating larger       
>   intermediate values and higher gas cost for the multiplication             
>                                                                                     
>   So the 8-decimal normalization is correct and gas-efficient. No precision is lost
>   since both prices use the same base.

Wizdave97 (commented):
@royvardhan  have we changed all the infinite approvals in simplex to per order approvals instead?

royvardhan:
> @royvardhan  have we changed all the infinite approvals in simplex to per order approvals instead?

Yes.

seunlanlege:
I'm not sure i like the fact that we're using chainlink? can't we get a twap/vwap from uniswap?

royvardhan:
> I'm not sure i like the fact that we're using chainlink? can't we get a twap/vwap from uniswap?

Chainlink aggregates prices off-chain from multiple independent sources rather than relying on a single DEX pool, making it significantly harder to manipulate. It's also more gas-efficient and eliminates the need to find and track high-liquidity pools for reliable pricing.

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -0,0 +1,376 @@
...
+///   0x01  APPROVE — Token must be pre-approved to this paymaster (BSC path,
+///                    or any chain where permit is unavailable). Batch the
+///                    approve() call in the UserOp's calls array.
+///
+/// paymasterData encoding:
+///   Mode 0x00 (permit):
+///     abi.encodePacked(uint8(0), address(token), uint256(permitAmount),
+///                      uint256(deadline), uint8(v), bytes32(r), bytes32(s))
+///   Mode 0x01 (approve):
+///     abi.encodePacked(uint8(1), address(token))
```
seunlanlege, line 41: I don't think we should support this approve mode. For safety users should only sign permits

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -0,0 +1,376 @@
...
+    /// @notice Treasury address — receives accumulated markup surplus only.
+    ///         Unused gas refunds are returned to the solver by OZ's base
+    ///         PaymasterERC20._postOp automatically.
+    address public treasury;
+
+    /// @notice Per-token configuration. token address → config.
+    mapping(address => TokenConfig) public tokenConfigs;
+
+    /// @notice Set of registered token addresses (for enumeration).
+    address[] public registeredTokens;
```
seunlanlege, line 93: likely unneeded

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -0,0 +1,376 @@
...
+/// Price conversion uses two Chainlink feeds per token:
+///   - token/USD (e.g. USDC/USD) → 8 decimals
+///   - nativeAsset/USD (e.g. BNB/USD, ETH/USD) → 8 decimals
+///   tokenPrice = (nativePrice * 10^tokenDecimals) / tokenPrice_USD
+///   This gives the cost in token-units per 1 wei of native gas.
+///
+/// Treasury: markup surplus accumulates in the contract and is withdrawable
+///           to a separate treasury address. Unused gas is always refunded
+///           to the solver (handled by OZ's PaymasterERC20._postOp).
+contract SimplexPaymaster is PaymasterERC20, Ownable {
```
seunlanlege, line 52: Can this contract receive eth? How does it make entry point deposits, maybe we override recieve so it automatically makes entry point deposits

seunlanlege:
Looking more at this PR it appears it still doesn't support USDT, since USDT doesn't have EIP-2612. I think we should pause work on this pending we have a solution that supports USDT otherwise circle's paymaster is still the best choice, no additional infra overhead

seunlanlege:
Can we split out 

DelegationService now delegates via bundler UserOp with paymaster paying gas — solver needs zero native tokens.

To a different PR?

seunlanlege:
Lets bring this back for bnb chain, so we can have a USDT paymaster

PR #778, opened 2026-04-17
[sdk,simplex]: Delegate EOA to SolverAccount using UserOp

royvardhan:
Ref: https://github.com/polytope-labs/hyperbridge/pull/768#issuecomment-4263106303

`DelegationService` now delegates via bundler `UserOp` with paymaster paying gas — solver needs zero native tokens. Supports Pimlico, Alchemy, and generic bundler gas pricing. Falls back to direct tx. `ContractInteractionService` approval scoped to exact required amount instead of `maxUint256`


Wizdave97 (commented):
Let's update the simplex docs

sdk/packages/simplex/src/services/paymaster/provider/circle.ts
```diff
@@ -0,0 +1,77 @@
...
+	if (existingAllowance >= threshold) {
+		// No permit data needed — paymaster will use existing allowance
+		const paymasterData = encodePacked(["uint8"], [0]) as HexString
+		return {
+			paymaster: paymasterAddress,
+			paymasterData,
+			paymasterVerificationGasLimit: VERIFICATION_GAS_LIMIT_CIRCLE,
+			paymasterPostOpGasLimit: POST_OP_GAS_LIMIT,
+		}
+	}
```
seunlanlege, line 53: prefer we always use permit2

sdk/packages/simplex/src/services/paymaster/types.ts
```diff
@@ -0,0 +1,61 @@
...
+	/** Which paymaster was selected. */
+	type: "circle" | "none"
+	/** Paymaster contract address (undefined when type is "none"). */
+	address?: HexString
+}
+
+// ── Authorization amount constants ──────────────────────────────────
+
+/** Dollar amount to authorize (permit). Safe upper bound — unused gas is refunded. */
+export const RECOMMENDED_AMOUNT_USD = 5n
```
seunlanlege, line 37: Lets only permit $1 per order.

sdk/packages/simplex/src/services/DelegationService.ts
```diff
@@ -45,10 +58,9 @@ export class DelegationService {
 		const authorityAddress = this.signer.account.address as HexString
 		const currentNonce = await publicClient.getTransactionCount({
 			address: authorityAddress,
-			blockTag: "pending",
+			blockTag: "latest",
```
seunlanlege, line 61: why?

sdk/packages/simplex/src/services/paymaster/provider/circle.ts
royvardhan: Yes we always use it only.

sdk/packages/simplex/src/services/paymaster/types.ts
royvardhan: What about expensive chains like Ethereum? I think $5 is okay since it leaves space for no permit for new orders in all chains until the threshold is hit, which is $2 here.

I also don't feel much difference between $1 and $5 in terms of security, but with $5, we save execution gas for the next orders.

sdk/packages/simplex/src/services/DelegationService.ts
royvardhan: Because the bundler submits the tx (not the EOA), the authority's nonce isn't bumped, so the auth must be signed against the actual on-chain nonce (latest) rather than pending, which could be inflated by unrelated pending txs from the EOA.

seunlanlege (approved):

royvardhan:
> Let's update the simplex docs

Yes

seunlanlege (approved):

Wizdave97 (approved):

PR #786, opened 2026-04-20
[sdk]: HandlerV2 support 

seunlanlege:
## Summary

adds HandlerV2 batch path. When the destination handler advertises `IHandlerV2` via ERC-165, the SDK queries the consensus proof from Hyperbridge's offchain storage and bundles `handleConsensus` + `handlePostRequests`/`handleGetResponses` into a single `batchCall`, replacing the old flow that waited for a separate consensus message to land on the destination. HandlerV1 path preserved as fallback.


modules/pallets/beefy-consensus-proofs/src/lib.rs
```diff
@@ -481,21 +416,12 @@ pub mod pallet {
...
-					.and_provides((ROT_TAG, outcome.current_set_id).encode())
-			} else {
-				// Single fixed slot — highest `proven_height` wins.
-				builder.priority(outcome.proven_height).and_provides(MSG_TAG.encode())
-			};
-
-			tx.build()
+				.propagate(true)
+				.priority(outcome.proven_height)
+				.and_provides(types::PROOF_TAG.encode())
```
Wizdave97, line 423: Shouldn't we maintain separate tags for mandatory and messaging updates to prevent potential conflicts, given the slim possibility of messaging updates being submitted while mandatory updates remain in the pool due to Substrate transaction pool inconsistencies?

royvardhan (approved):
The SDK side looks good to me.

modules/pallets/beefy-consensus-proofs/src/lib.rs
seunlanlege: Messaging proofs also rotate the authority set. We can't miss authority set rotations because the light client can only accept updates from either the current or next authority set

Wizdave97: Yeah thought about it, light client cannot progress without handoffs

PR #802, opened 2026-04-28
[runtime]: Uncle proofs in pallet-beefy-consensus-proofs

### LINKED ISSUE #798: [pallet-beefy-consensus-proofs]: Uncle proofs
## Summary

Currently `pallet-beefy-consensus-proofs` is winner-take-all: only the first prover to submit a valid BEEFY consensus proof for a given height range gets rewarded. This discourages multiple provers from operating since only the fastest survives economically.

This issue proposes **uncle proofs** — allowing the first N provers (configurable, e.g. 5) who independently generated valid proofs for the same height range to all receive rewards via a configurable decreasing reward curve.

Uncle proofs only apply to **SP1 ZK proofs** on mainnet. Naive proofs are used on testnet where the reward is 0, so there is no incentive structure to extend.

## Motivation

A healthy prover ecosystem requires multiple independent operators. Winner-take-all economics naturally consolidates to a single fast prover, creating a single point of failure for cross-chain message delivery. Uncle proofs incentivize redundancy by rewarding provers who did the work but weren't the fastest.

## Anti-Replay Design

A critical concern: if proof bytes are public (they're in the tx and in offchain storage), what prevents someone from copying a valid proof and re-signing it with their own key to claim uncle rewards without doing any proving work?

**SP1 Groth16 proofs use randomized witnesses**, so two provers independently proving the same statement produce **different proof bytes**. The pallet tracks `keccak256(proof)` for each accepted proof at a given height. If someone copies another prover's proof bytes and re-signs, the proof hash matches the original and gets rejected as `ProofAlreadySubmitted`.

## Signed Extrinsic with `Pays::No`

`submit_proof` becomes a **normal signed extrinsic** instead of an unsigned one. This simplifies the design significantly:

- **First proof and uncle proofs (positions 0–N)**: Return `Pays::No` — the prover pays no transaction fee and receives their reward. The tx fee is refunded since the proof did useful work.
- **Proofs beyond `MaxUncleProvers` or replayed proof hashes**: The extrinsic fails normally and the submitter **pays the transaction fee**. This creates a natural economic disincentive against spam — submitting junk or late proofs costs money.

This eliminates the need for `validate_unsigned` entirely, along with its complexity around discarded overlays, provides tags, and pre-dispatch hooks. The transaction pool handles deduplication naturally via nonce ordering.

### Why this is better than unsigned

- No `validate_unsigned` / `ValidateUnsigned` impl needed — simpler code
- No provides tag deduplication logic — the pool just orders by priority/fee
- Failed proofs pay tx fees → natural spam deterrent
- Provers need account balance for tx fees, but successful proofs refund via `Pays::No`
- No custom signature verification scheme needed — the runtime's standard signed extrinsic handles authentication

### Extrinsic change

The `submitter` field in `SubmitProofPayload` is no longer needed — the signer of the extrinsic IS the submitter. The SR25519 signature scheme over `(SIGNATURE_DOMAIN, submitter, proof_hash)` is also removed. The extrinsic simply takes the proof bytes:

```rust
#[pallet::call_index(1)]
#[pallet::weight(T::WeightInfo::submit_proof())]
pub fn submit_proof(
    origin: OriginFor<T>,
    proof: Vec<u8>,
) -> DispatchResultWithPostInfo {
    let submitter = ensure_signed(origin)?;
    // ... verification ...
    Ok(PostDispatchInfo { actual_weight: None, pays_fee: Pays::No })
}
```

## Decreasing Reward Curve

A configurable on-chain reward curve determines the reward for each position:

| Position | Example multiplier |
|---|---|
| 0 (first) | 100% |
| 1 (first uncle) | 80% |
| 2 | 60% |
| 3 | 40% |
| 4 | 20% |

Set via an admin extrinsic `set_reward_curve` that stores a `BoundedVec<(u32, u32), MaxUncleProvers>` of `(numerator, denominator)` fractions.

## Uncle Verification Path

The first proof advances state via the existing `verify_and_apply` path. Subsequent proofs for the same height will fail `verify_and_apply` with `StaleProof`/`NoNewWork` since state has already advanced. When this happens, `submit_proof` falls through to an uncle verification path that:

1. Checks the proof is an SP1 ZK proof (naive proofs are not eligible for uncle rewards)
2. Checks `MaxUncleProvers` hasn't been reached
3. Checks the proof hash hasn't been seen before (anti-replay)
4. Verifies the proof **cryptographically** against a saved pre-proof consensus state snapshot — calls `beefy_verifier::sp1::verify_sp1_consensus` directly (pure verification, no state mutation)
5. Pays the position-appropriate reward and mints reputation tokens
6. Returns `Pays::No`

### Consensus state snapshot

When the first proof lands, the pallet snapshots the consensus state bytes from `pallet-ismp` **before** `verify_and_apply` mutates them. Uncle proofs verify against this snapshot since the live consensus state has already been advanced by the first proof. The snapshot is pruned when the height is evicted from the ring buffers.

## New Config Items

```rust
/// Maximum number of unique provers rewarded per height (including the first).
#[pallet::constant]
type MaxUncleProvers: Get<u32>;
```

## New Storage

| Item | Type | Purpose |
|---|---|---|
| `ProofContext` | `Map<u64, Vec<u8>>` | Pre-proof consensus state snapshot for uncle verification. |
| `ProverCount` | `Map<u64, u32>` | Number of provers rewarded per height. |
| `AcceptedProofHashes` | `Map<u64, BoundedVec<H256>>` | Proof hashes accepted per height. Anti-replay. |
| `RewardCurve` | `BoundedVec<(u32, u32)>` | Decreasing reward fractions by position. |

All pruned when the corresponding height is evicted from `MessagingProofs`/`RotationProofs` ring buffers.

## New Admin Extrinsic

- `set_reward_curve(curve: BoundedVec<(u32, u32), MaxUncleProvers>)` — Sets the reward multiplier for each prover position.

## Modified Extrinsic: `submit_proof`

Changes from unsigned to signed. Unified flow:

```
let submitter = ensure_signed(origin)?;

try verify_and_apply(proof):
  success → first proof, position 0, store snapshot + metadata
  StaleProof/NoNewWork → fall through to verify_uncle()
    check proof_type == SP1 (naive not eligible for uncle)
    success → uncle proof, position = current ProverCount
    failure → return error (submitter pays tx fee)

check proof hash not in AcceptedProofHashes (anti-replay)
compute reward from RewardCurve[position]
pay reward + mint reputation tokens
record proof hash, increment ProverCount
emit ProofAccepted { position, ... }
return Ok(Pays::No)  // successful proofs don't pay tx fees
```

## Removals

- `ValidateUnsigned` impl — no longer needed
- `SubmitProofPayload` struct — replaced by bare proof bytes
- `Signature` type and `SIGNATURE_DOMAIN` — extrinsic signature handles auth
- `PROOF_TAG` and provides tag logic — tx pool handles naturally
- `PROOF_LONGEVITY` constant — not needed for signed txs

## Security Considerations

1. **Anti-replay**: Proof hash uniqueness prevents copied proofs from earning rewards. SP1 Groth16 randomization ensures independent provers produce different hashes.
2. **Spam deterrent**: Failed proofs pay transaction fees. Only successful first/uncle proofs get `Pays::No`.
3. **Bounded cost**: `MaxUncleProvers` caps rewards per height, bounding treasury drain.
4. **Same-block ordering**: If two proofs arrive in the same block, the block builder orders them. The first advances state; the second falls through to the uncle path.
5. **Prover balance requirement**: Provers now need enough balance to cover tx fees for submission. Successful proofs are refunded (`Pays::No`), so the net cost is zero for valid work.
6. **Naive proofs excluded**: Uncle rewards only apply to SP1 ZK proofs. Naive proofs (testnet-only, zero reward) are rejected at the uncle verification step.

## Suggested Defaults (Mainnet)

- `MaxUncleProvers`: 5
- `RewardCurve`: `[(1,1), (4,5), (3,5), (2,5), (1,5)]` (100%, 80%, 60%, 40%, 20%)

dharjeezy:
This PR adds uncle proofs so multiple independent `SP1` provers can race for the same target and all get paid, with a decreasing reward curve so the fastest still wins the largest share. Naive proofs are testnet-only with zero reward, so uncles apply to `SP1` only.    

`submit_proof` is now a signed extrinsic. The signer is the reward payee. Successful proofs return `Pays::No` (fee refunded along with the reward); failed proofs pay the fee, which is the spam deterrent. The old `ValidateUnsigned` path with hand-rolled `SR25519` signing is gone.
     

closes #798


modules/pallets/beefy-consensus-proofs/src/lib.rs
```diff
@@ -342,160 +470,191 @@ pub mod pallet {
...
+			// `latest_height` is for the *newer* first proof; the uncle's lower
+			// `proof.block_number` will fail SP1 verification against it (StaleHeight),
+			// so a stale uncle pays the tx fee.
+			let parachain_height = Self::latest_height();
+
+			let snapshot_bytes =
+				ProofContext::<T>::get(parachain_height).ok_or(Error::<T>::NoUncleContext)?;
+
+			let position = ProverCount::<T>::get(parachain_height);
+			if position >= T::MaxUncleProvers::get() {
```
seunlanlege:
```suggestion
			if position > T::MaxUncleProvers::get() {
```
So we can record positions 0(first proof)..=T::MaxUncleProvers

seunlanlege:
I think its a good idea to have some simnode tests with naive beefy to validate this behaviour

modules/pallets/beefy-consensus-proofs/src/lib.rs
```diff
@@ -504,6 +696,18 @@ pub mod pallet {
...
+					// Match the inner SP1 verifier's stale check before paying its cost. An
+					// uncle proof targets the same `block_number` the first prover advanced
+					// state to, so it would fail this same comparison inside SP1; surface
+					// it as `StaleProof` so the dispatcher routes to the uncle path.
+					let proof_block: u32 = abi_proof
+						.commitment
+						.blockNumber
+						.try_into()
+						.map_err(|_| Error::<T>::AbiDecodeFailed)?;
+					if prev_state.latest_beefy_height >= proof_block {
```
seunlanlege: Rather than duplicating this check here, we should match over this error from `beefy-verifier`

https://github.com/polytope-labs/hyperbridge/blob/main/modules/consensus/beefy/verifier/src/sp1.rs#L54-L57

seunlanlege:
Looking good, 2 things

- [x] Lets use a boundedvec for proof size so its rejected at txpool not at block execution
- [x] simnode should construct multiple proofs so we test the uncle proofs

seunlanlege (approved):

modules/pallets/beefy-consensus-proofs/src/lib.rs
```diff
@@ -519,29 +713,29 @@ pub mod pallet {
...
 			.map_err(|e| {
+				let msg = format!("{e:?}");
 				log::warn!(
 					target: "ismp",
-					"[beefy-consensus-proofs] handle_incoming_message failed: {e}",
+					"[beefy-consensus-proofs] handle_incoming_message failed: {msg}",
 				);
-				Error::<T>::VerificationFailed
+				// `beefy_verifier::error::Error::StaleHeight` propagates through
+				// `BeefyConsensusClient::verify_consensus` as a `Custom(...)` string;
```
Wizdave97: Let's add a StaleHeight variant to ismp error so we can use that directly instead of the String

seunlanlege: I think we already have an anyhow error variant on `ismp::Error` so technically we can match over that & do [downcast_ref](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.downcast_ref)

seunlanlege:
So essentially anyone can copy a proof, add some byte padding and it would still succeed? @dharjeezy we need to prevent this actually.

Either by ensuring decoding consumes all bytes, or re-encoding to abi before hashing to ensure proof uniqueness

modules/pallets/beefy-consensus-proofs/src/lib.rs
```diff
@@ -420,6 +423,41 @@ pub mod pallet {
...
+			// the input exactly — anything else fails decode at the boundary.
+			let abi_payload = &proof[1..];
+			match proof_type {
+				types::PROOF_TYPE_SP1 => {
+					let p =
+						<ismp_solidity_abi::sp1_beefy::SP1Beefy::SP1BeefyProof as SolType>::abi_decode_params(
+							abi_payload,
+						)
+						.map_err(|_| Error::<T>::AbiDecodeFailed)?;
+					if <ismp_solidity_abi::sp1_beefy::SP1Beefy::SP1BeefyProof as SolType>::abi_encode_params(&p)
```
seunlanlege: better to use the abi encoding to compute the proof hash instead

Wizdave97 (approved):

Wizdave97:
@dharjeezy  simtest is failing

dharjeezy:
> @dharjeezy simtest is failing

yes, it is because the test is against main, simtest passes against this branch when we use pull request target

PR #807, opened 2026-05-01
[sdk]: improve order interface

### LINKED ISSUE #806: Improve SDK interface for placing and cancelling orders
The SDK should not require hex string for state machine ids or padded addresses for tokens and user accounts.

We should accept the simplest inputs and do all the conversions internally.

We should also improve the cancel order interface.

ddboy19912:


docs/content/developers/sdk/api/intent-gateway.mdx
```diff
@@ -171,11 +171,11 @@ async *cancelOrder(
 |------|------|---------|-------------|
 | `order` | `Order` | — | The order to cancel |
 | `indexerClient` | `IsmpClient` | — | Indexer client for streaming ISMP request status |
-| `fromDest` | `boolean` | `false` | If `true`, initiates cancellation from the destination chain |
+| `options` | `{ from?: "source" \| "destination" }` | `{}` | Pass `{ from: "destination" }` to initiate on the destination chain. |
 
 **Yields:** `CancelEvent`
 
-**Throws:** If the order is same-chain and `fromDest` is `true`
+**Throws:** If the order is same-chain and destination-side cancellation is requested
```
Wizdave97: It should not throw, it should just proceed with cancelling, also if from or dest is not passed and the order is same chain, it should also proceed to cancel

Wizdave97 (approved):

PR #815, opened 2026-05-07
[runtime]: strip pallet-messaging-fees

royvardhan:
Strips `pallet-messaging-fees` down to a single responsibility — minting reputation tokens to relayers proportional to the bytes they deliver. Rewards math, `fee-distribution`, `target-message-size`, and `per-route` incentive tracking are gone.


modules/pallets/messaging-fees/src/lib.rs
```diff
@@ -13,193 +13,204 @@
...
-
-		/// Price oracle for usd price conversion to bridge tokens
-		type PriceOracle: PriceOracle;
-
-		/// The pallet-assets instance used for managing the reputation token.
-		type ReputationAsset: fungible::Mutate<
-			Self::AccountId,
-			Balance = <Self as pallet_ismp::Config>::Balance,
-		>;
-		/// Weight information for operations
```
Wizdave97, line 94: WeightInfo should not be removed, update the benchmarks

PR #826, opened 2026-05-13
[indexer]: align indexer with updated intents v2 events

### LINKED ISSUE #820: [Indexer] Intent Gateway Updates
Update the indexer with changes made to IntentGatewayV2 events

royvardhan:
Closes: #820


seunlanlege:
I think we should also index the actual amounts that are released from escrow/refunded

royvardhan:
@seunlanlege Done

seunlanlege (approved):

royvardhan:
@Wizdave97 Do I still need to change schema name for order v2? Even though we never officially launched it?

Wizdave97:
> @Wizdave97 Do I still need to change schema name for order v2? Even though we never officially launched it?



> @Wizdave97 Do I still need to change schema name for order v2? Even though we never officially launched it?

yes some data has been indexed, best to change everything relating to intent gateway to V3, since it's a new contract

sdk/packages/sdk/src/configs/chain.ts
```diff
@@ -306,8 +306,8 @@ export const chainConfigs: Record<number, ChainConfigData> = {
 		},
 		addresses: {
 			IntentGateway: "0x1a4ee689a004b10210a1df9f24a387ea13359acf",
-			IntentGatewayV2: "0x2d61624A17f361020679FaA16fbB566C344AaF4B",
-			SolverAccount: "0x66C4459fa61E5Ca647152EEb6dA56150EE975512",
+			IntentGatewayV2: "0x09a74AA91eb532C21862d4C6222753e4fC7e22dB",
```
Wizdave97, line 309: Shouldn't we rename this to V3 also?

Wizdave97 (dismissed):
Let's fix conflicts so we can synchronize changes with main

sdk/packages/sdk/src/configs/chain.ts
royvardhan: There is a dilemma that the contract has been deployed with the old naming.

Wizdave97: Ohh yeah, let's keep the current naming convention then, we can just keep the renaming to the graphql schema

seunlanlege (approved):

PR #839, opened 2026-05-15
[evm]: scope HandlerV2 epoch tracking per host address

seunlanlege:
## Summary
- Key `_currentEpoch` and `_epochs` mappings in HandlerV2 by `address(host)` to prevent a malicious host contract from advancing the epoch and front-running legitimate hosts
- Update `relayerOf()` and `currentEpoch()` signatures to accept a `host` parameter
- Update SDK ABI and callers to match new signatures

## Test plan
- [ ] HandlerV2Test passes with updated assertions
- [ ] SDK correctly passes host address to `currentEpoch` calls

### DIFF
--- a/evm/src/core/HandlerV2.sol
+++ b/evm/src/core/HandlerV2.sol
@@ -114,15 +114,15 @@ contract HandlerV2 is IHandlerV2, ERC165, Context {
     event NewEpoch(uint256 indexed authoritySetId, address indexed relayer);
 
     /**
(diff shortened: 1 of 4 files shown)


seunlanlege:
On second thoughts best to keep the handlers stateless & move it to the host

seunlanlege:
Superseded by #844, which moves epoch tracking to the Host (keeping HandlerV2 stateless) instead of scoping the handler-side mapping per-host.

PR #840, opened 2026-05-15
[ismp, evm, runtime]: switch to abi.encode and remove PostResponse

seunlanlege:
## Summary
- **Fix hash malleability**: Switch all ISMP type hashing from `abi.encodePacked` to `abi.encode` in both Solidity and Rust. `encodePacked` creates malleable requests/responses whose bytes can be shifted to create different valid encodings that hash to the same values, enabling critical vulnerabilities
- **Remove PostResponse**: Drop the `PostResponse` type and all associated handlers, dispatchers, events, and storage from the entire ISMP protocol stack (Solidity contracts, Rust ISMP core, pallets, and test suites)
- **New Rust ABI encoder**: Add `modules/ismp/core/src/abi.rs` implementing Solidity-compatible `abi.encode` semantics in Rust to ensure cross-chain encoding parity

## Test plan
- [x] Solidity compilation passes
- [x] Rust ISMP core compiles
- [x] Pallet compilation passes
- [x] EVM Rust test suite passes (post_response tests removed)
- [x] Foundry tests pass (PostResponse test files to be cleaned up)
- [x] Verify encoding parity between Rust and Solidity abi.encode output


parachain/runtimes/nexus/src/ismp.rs
```diff
@@ -358,22 +358,6 @@ impl IsmpModule for ProxyModule {
...
-				anyhow!(
-					"bandwidth gate: {err} (source={:?}, from={:x?})",
-					post.source_chain(),
-					post.source_module(),
-				)
-			})?;
-		}
-
 		if response.dest_chain() != HostStateMachine::get() {
 			Ismp::dispatch_response(
```
Wizdave97:
```suggestion
```

Wizdave97 (approved):

PR #843, opened 2026-05-15
[runtime]: multiple requests in a batch must share a single delivery address

dharjeezy:
This PR enforces single delivery address per accumulate_fees batch

introduces:

- New error variant `Error::MixedDeliveryAddressesInBatch` depicting rejection of batches with multiple addresses
- Replaced the per-iteration overwrite loop in `accumulate` with a single `ensure!(result.len() == 1, ...)` plus a one-entry extraction. 
- Concrete test for these changes

### DIFF
--- a/modules/pallets/relayer/src/lib.rs
+++ b/modules/pallets/relayer/src/lib.rs
@@ -240,6 +240,9 @@ pub mod pallet {
 		MissingCommitments,
 		/// Fee accumulation proof contains no address
 		IncompleteProof,
+		/// Withdrawal batch contains commitments delivered by more than one
+		/// relayer address.
+		MixedDeliveryAddressesInBatch,
 		/// Signature Decoding Error
 		SignatureDecodingError,
 		/// `(destination, set_id)` has already been claimed by some relayer.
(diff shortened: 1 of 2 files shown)


modules/pallets/relayer/src/lib.rs
```diff
@@ -713,14 +716,9 @@ where
...
-		// We expect the relayer used the same address for all deliveries in this batch
-		// That's the only behaviour supported by tesseract relayer
-		let mut delivery_address = Default::default();
-		for (address, fee) in result.clone().into_iter() {
-			delivery_address = address;
-			total_fee += fee;
-		}
+		// Every commitment in the batch must share a single delivery address.
+		ensure!(result.len() == 1, Error::<T>::MixedDeliveryAddressesInBatch);
+		let (delivery_address, total_fee) = result.into_iter().next().expect("len == 1; qed");
```
Wizdave97: Let's use map_err
```suggestion
		let (delivery_address, total_fee) = result.into_iter().next().ok_or_else(|| Error::<T>::MixedDeliveryAddressesInBatch)?;
```

Wizdave97 (approved):

PR #844, opened 2026-05-15
[evm, sdk, tesseract]: move HandlerV2 epoch tracking to the Host

seunlanlege:
## Summary

Moves authority-set epoch tracking off `HandlerV2` and into the `Host`, restoring `HandlerV2` to a stateless verifier/dispatcher. Supersedes #839, which scoped epoch storage per-host inside the handler instead.

- `IHost` gains `currentEpoch()`, `relayerOf(uint256)`, and `recordEpoch(uint256, address)`. `recordEpoch` is `restrict(_hostParams.handler)`-gated, matching the existing pattern used by `storeConsensusState` / `storeStateMachineCommitment`.
- `EvmHost` owns `_epochs` / `_currentEpoch` storage and emits `NewEpoch`. Per-host isolation falls out naturally from each host having its own storage — no nested mappings needed.
- `HandlerV2` loses its mappings, views and event. `handleConsensus` now ends with `host.recordEpoch(nextAuthoritySetId - 1, _msgSender())`, guarded against `nextAuthoritySetId == 0`.
- TS ABIs regenerated from `forge build` output. `EvmChain.currentEpoch()` reads the host directly (no longer goes via `hostParams.handler`).
- `tesseract-evm` imports `NewEpoch` from `evm_host` instead of `handler_v2`; the corresponding re-export in `evm/abi/src/generated/handler.rs` is trimmed.

## Test plan
- [x] `forge test` — all 266 EVM tests pass (12 `HandlerV2Test` cases assert against the host, with proof `nextAuthoritySetId` values bumped to match `epoch = nextAuthoritySetId - 1`).
- [x] `cargo check --workspace` clean.
- [x] `pnpm tsc --noEmit` clean in `sdk/packages/sdk`.
- [x] Manual relayer sanity check on testnet after merge.


Wizdave97 (commented):
We need tom modify pallet-ismp-relayer  to verify epoch claims from the host and tesseract to prove from the host

Wizdave97 (approved):

PR #849, opened 2026-05-15
[runtime]: enforce epoch state proof on every verifier state update

dharjeezy:
This PR ensures that every `VerifierStateUpdate` carries an `EpochProof` verified against the block's `state_root`, and the verifier decides whether to rotate based on the proven value.

### DIFF
--- a/modules/consensus/pharos/verifier/src/lib.rs
+++ b/modules/consensus/pharos/verifier/src/lib.rs
@@ -22,6 +22,7 @@ extern crate alloc;
 pub mod error;
 pub mod state_proof;
 
+use core::cmp::Ordering;
 use error::Error;
 use geth_primitives::Header;
 use ismp::messaging::Keccak256;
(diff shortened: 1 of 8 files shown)


modules/consensus/pharos/verifier/src/lib.rs
```diff
@@ -71,37 +64,75 @@ pub fn verify_pharos_block<C: Config, H: Keccak256 + Send + Sync>(
...
+				.validator_set_proof
+				.ok_or(Error::MissingValidatorSetProof { block_number: update_block_number })?;
+
+			let new_validator_set = state_proof::verify_validator_set_proof::<H>(
+				update.header.state_root,
+				&validator_set_proof,
+				observed_epoch,
+			)?;
+
+			verify_block_signature(&new_validator_set, &update.block_proof, computed_hash)?;
```
Wizdave97: This should verify the proof with the trusted validator set
```suggestion
			verify_block_signature(&trusted_state.current_validator_set, &update.block_proof, computed_hash)?;
```

seunlanlege (approved):

PR #869, opened 2026-05-17
[runtime]: Fix accumulate fees regression

Wizdave97:
Fee accumulation read the source-chain request commitment via `commitment_state_trie_key`, which addresses offset 1 of the EVM `FeeMetadata` struct (the `sender` field) so membership proofs hit an always-non-zero slot. `accumulate` then decoded that value as the relayer fee, but the fee lives at offset 0.

For EVM source chains the relayer now derives the `RequestCommitments` storage key at offset 0 directly in `accumulate`. `commitment_state_trie_key` is left untouched at offset 1 for membership verification, and Substrate sources continue to use it as-is since their commitment key already addresses the full metadata blob.

### DIFF
--- a/modules/pallets/relayer/src/accumulate.rs
+++ b/modules/pallets/relayer/src/accumulate.rs
@@ -73,28 +73,11 @@ where
 		let dest_sm = validate_state_machine(&host, withdrawal_proof.dest_proof.height)
 			.map_err(|_| Error::<T>::ProofValidationError)?;
 		let state_machine = withdrawal_proof.source_proof.height.id.state_id;
-		// Fee accumulation must read the relayer fee from the source-chain request commitment.
-		// On EVM chains the `RequestCommitments` mapping stores a `FeeMetadata { fee, sender }`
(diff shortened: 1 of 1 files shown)


seunlanlege:
What happens if the attached fee is zero?

modules/pallets/relayer/src/accumulate.rs
```diff
@@ -71,9 +72,30 @@ where
...
+		let source_keys = if state_machine.is_evm() {
+			withdrawal_proof
+				.commitments
+				.iter()
+				.map(|commitment| {
+					let slot = derive_unhashed_map_key_with_offset::<<T as Config>::IsmpHost>(
+						commitment.0.to_vec(),
+						REQUEST_COMMITMENTS_SLOT,
+						0,
+					);
```
seunlanlege: Would be cool to get a named helper method

Wizdave97:
> What happens if the attached fee is zero?

It is skipped in `validate_results`, it would be seen as a non-membership proof.

seunlanlege (approved):

PR #872, opened 2026-05-18
[runtime]: Tighten fee refund execution

dharjeezy:
The refund wrapper previously looked the fee up in storage, but the ISMP framework removes that entry just before invoking the callback, so the lookup never returned anything. `IsmpHost` now exposes an `on_request_timeout` hook that the timeout handler calls after the module callback returns `Ok`, passing back the encoded metadata that `delete_request_commitment` already produced. The pallet implements the hook to refund the payer, and the `RefundingModule` wrapper is gone.


seunlanlege:
Not a fan of this unfortunately. We need to do refunds somewhere else then.

Maybe a new host method.

seunlanlege (approved):

PR #877, opened 2026-05-18
[evm]: gate UniV3/V4 wrapper receive() to legitimate refund senders only

royvardhan:
## Summary

- **UniV4UniswapV2Wrapper.receive()** now reverts unless `msg.sender == _params.universalRouter`. The Universal Router is the only legitimate `receive()` caller (it refunds unused ETH to the wrapper after `execute()`).
- **UniV3UniswapV2Wrapper.receive()** now reverts unless `msg.sender == _params.WETH` — defense-in-depth against the same preload class. WETH is the only legitimate `receive()` caller (during `IWETH.withdraw(refund)`).

## Why

An unprivileged griefer could preload `UniV4UniswapV2Wrapper` via its open `receive()`, causing `swapETHForExactTokens` to read `refundETH = address(this).balance` as `preloaded + (msg.value - actual_spent)` and (a) revert the call via underflow when `preloaded > actual_spent`, or (b) under-report `amounts[0]` to the V2-style caller (`IntentGatewayV2.placeOrder`). When IG holds non-zero native escrow, the under-report path causes IG to over-refund the caller from its own balance, draining `(preloaded)` from gateway → admin EOA per call and blocking later native-escrow refunds with `InsufficientNativeToken`.

### DIFF
--- a/evm/script/DeployUniV4Wrapper.sol
+++ b/evm/script/DeployUniV4Wrapper.sol
@@ -18,8 +18,6 @@ contract DeployScript is BaseScript {
         uint24 defaultFee = uint24(config.get("DEFAULT_FEE").toUint256());
         int24 defaultTickSpacing = int24(config.get("DEFAULT_TICK_SPACING").toInt256());
         address weth = config.get("WETH").toAddress();
-        address intentGateway = config.get("INTENT_GATEWAY_V2").toAddress();
-        address ismpHost = config.get("HOST").toAddress();
 
         UniV4UniswapV2Wrapper wrapper = new UniV4UniswapV2Wrapper{salt: salt}(admin);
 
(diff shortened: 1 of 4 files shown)


Wizdave97 (approved):

MathCarv:
Hi @Wizdave97 ,

I noticed that the fix for this report has now been merged in PR #877:

The PR summary appears to confirm the same impact demonstrated in my supplemental PoC: preloading `UniV4UniswapV2Wrapper` could corrupt `amounts[0]`, cause `IntentGatewayV2.placeOrder` to over-refund from its own native balance, and leave later native-escrow refunds blocked with `InsufficientNativeToken`.

Since the issue has been triaged as valid and the fix has now been merged, could you please confirm the next steps regarding final bounty assessment and expected payment timeline?

Best,  
@ peixinn

PR #881, opened 2026-05-19
[runtime]: Bind source chain to relayer redirect signature

dharjeezy:
The accumulate flow verified a beneficiary redirect signature against the bare keccak of the beneficiary address, while the withdrawal flow already signs the beneficiary alongside the destination chain and a per relayer nonce. This change folds the source state machine into the prehash through a new `beneficiary_message` helper exported from `pallet-ismp-relayer`, and updates the off chain signer in `transaction-fees` to compute the same payload.

### DIFF
--- a/tesseract/messaging/primitives/src/lib.rs
+++ b/tesseract/messaging/primitives/src/lib.rs
@@ -586,6 +586,11 @@ pub trait HyperbridgeClaim {
 	) -> anyhow::Result<Vec<WithdrawFundsResult>>;
 	/// Check if this commitment has been claimed
 	async fn check_claimed(&self, commitment: H256) -> anyhow::Result<bool>;
+	/// Current relayer signature nonce stored on Hyperbridge for this `(address, chain)`
+	/// pair. Used to pin a beneficiary redirect signature to one accumulate call.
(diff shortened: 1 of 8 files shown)


Wizdave97 (commented):
Let's also use the relayer nonce in the message

tesseract/messaging/primitives/src/lib.rs
```diff
@@ -586,6 +586,11 @@ pub trait HyperbridgeClaim {
 	) -> anyhow::Result<Vec<WithdrawFundsResult>>;
 	/// Check if this commitment has been claimed
 	async fn check_claimed(&self, commitment: H256) -> anyhow::Result<bool>;
+	/// Current relayer signature nonce stored on Hyperbridge for this `(address, chain)`
+	/// pair. Used to pin a beneficiary redirect signature to one accumulate call.
+	async fn relayer_nonce(&self, _address: Vec<u8>, _chain: StateMachine) -> anyhow::Result<u64> {
+		Ok(0)
```
Wizdave97: Don't add a default implementation

Wizdave97 (approved):

PR #889, opened 2026-05-20
[evm]: refund wrapper unspent ETH to host; drop deployer field

royvardhan:
Both `UniV3UniswapV2Wrapper` and `UniV4UniswapV2Wrapper` previously routed unspent ETH from the swap path to a `_deployer` address (the admin EOA). This PR moves the refund destination to the `msg.sender`

### DIFF
--- a/evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
+++ b/evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
@@ -78,8 +78,8 @@ contract UniV3UniswapV2Wrapper {
      */
     error InvalidWethAddress();
 
-    constructor() {
-        _deployer = msg.sender;
+    constructor(address deployer) {
+        _deployer = deployer;
     }
 
     /**
--- a/evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
+++ b/evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
@@ -41,8 +41,6 @@ contract UniV3UniswapV2Wrapper {
         address quoter;
         /// @dev The fees that helps point to the specific pool.
         uint24 maxFee;
-        /// @dev IsmpHost address that receives unspent ETH from swaps.
-        address host;
     }
 
     /**
@@ -146,7 +144,7 @@ contract UniV3UniswapV2Wrapper {
             uint256 refund = msg.value - spent;
             IWETH(weth).withdraw(refund);
 
-            (bool success,) = _params.host.call{value: refund}("");
+            (bool success,) = msg.sender.call{value: refund}("");
             if (!success) revert RefundFailed();
         }
 
--- a/evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
+++ b/evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
@@ -56,6 +56,11 @@ contract UniV3UniswapV2Wrapper {
     bool private _initialized;
 
     /**
+     * @dev The address that deployed the wrapper. Only this address may call `init()`.
+     */
+    address private immutable _deployer;
+
+    /**
      * @dev Error indicating that a deposit operation has failed.
      */
     error DepositFailed();
@@ -75,13 +80,17 @@ contract UniV3UniswapV2Wrapper {
      */
     error InvalidWethAddress();
 
+    constructor() {
+        _deployer = msg.sender;
+    }
+
     /**
      * @notice Initializes the Uniswap V3 to V2 wrapper module
      * @dev Can only be called once
      * @param params Initialization parameters.
      */
     function init(Params memory params) public {
-        if (_initialized) revert Unauthorized();
+        if (_initialized || msg.sender != _deployer) revert Unauthorized();
         // approve the swap router to spend WETH
         IERC20(params.WETH).approve(params.swapRouter, type(uint256).max);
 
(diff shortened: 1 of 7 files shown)


evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
```diff
@@ -80,17 +75,13 @@ contract UniV3UniswapV2Wrapper {
...
-        _deployer = deployer;
-    }
-
     /**
      * @notice Initializes the Uniswap V3 to V2 wrapper module
      * @dev Can only be called once
      * @param params Initialization parameters.
      */
     function init(Params memory params) public {
-        if (_initialized || msg.sender != _deployer) revert Unauthorized();
```
seunlanlege, line 93: so anyone can initialize?

royvardhan: Yes

royvardhan: Better to call in an atomic txn

evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
```diff
@@ -146,7 +146,7 @@ contract UniV3UniswapV2Wrapper {
             uint256 refund = msg.value - spent;
             IWETH(weth).withdraw(refund);
 
-            (bool success,) = _deployer.call{value: refund}("");
+            (bool success,) = _params.host.call{value: refund}("");
```
seunlanlege: refund to msg.sender

evm/src/utils/uniswapv2/UniV4UniswapV2Wrapper.sol
```diff
@@ -84,7 +85,7 @@ contract UniV4UniswapV2Wrapper {
         uint256 refundETH = address(this).balance;
 
         if (refundETH > 0) {
-            (bool success,) = _deployer.call{value: refundETH}("");
+            (bool success,) = _params.host.call{value: refundETH}("");
```
seunlanlege: same

evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
```diff
@@ -80,8 +80,8 @@ contract UniV3UniswapV2Wrapper {
      */
     error InvalidWethAddress();
 
-    constructor(address deployer) {
-        _deployer = deployer;
+    constructor() {
+        _deployer = msg.sender;
```
seunlanlege: msg.sender is the deterministic create2 deployer. better to make this explicit with a param

evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol
royvardhan: When we always make a hard path that we are refunding any excess eth to our decided address, be it deployer or host. We are saving ourselves from attack vector that we don't know yet. 

If we refund to the msg.sender, then sure in the IG or Host context it will refund to IG or Host. 

But when a malicious actor calls the swap function after preloading, it will automatically send ALL the eth to the msg.sender. 

For this the below attack vector, the malicious actor was causing DOS for some victims but ultimately the funds were safe because everything went to the deployer.

```
We need to look at two report simultaneously. 233 and 8. 

Trace:
  - Attacker preloads $1000 eth to the wrapper before the placeOrder call. 
  - Attacker places order with order.fees as $1000 and msg.value as $1000 eth.
  - IG calls wrapper {value: $1000}. Wrapper now has $999 preload + $1000 = $1999.
  - Wrapper forwards $1000 to Universal Router. Router consumes all $1000. Wrapper balance = $999 (the preload).
  - Wrapper line 84: refundETH = address(this).balance = $999. Sweeps that $999 to _deployer.
  - Wrapper line 92: amounts[0] = msg.value − refundETH = $1000 − $999 = $1.

 That amounts[0] = $1 is the lie — actual spent was $1000, but the wrapper subtracts the preloaded amount from msg.value and reports the difference. Back in IG:

  - Line 344: msgValue -= amounts[0] → local msgValue = $1000 − $1 = $999.
  - Line 353-354: msgValue > 0 is true, so IG sends $999 to msg.sender from its own balance.

The $999 the attacker receives at line 354 doesn't come from the swap — it comes from IG's standing native balance. IG thinks "I only spent $1 on the swap, the other $999 of the user's msg.value must be sitting unused in my balance, refund it." But $999 of the user's msg.value is actually feeToken now (in IG's hands as escrow), and $1 of it is genuinely unused. The wrapper's preload-laundering tricked IG into refunding $999 from somewhere else — namely, native balance backing prior orders.
```

seunlanlege: seems like we need to record a balance delta instead of thinking 

```
refundETH = address(this).balance
```

royvardhan: Okay, so I forgot that we have now gated the receive function on the wrappers. `msg.sender` is good to go.

Wizdave97 (approved):

evm/src/utils/uniswapv2/UniV4UniswapV2Wrapper.sol
```diff
@@ -84,7 +84,7 @@ contract UniV4UniswapV2Wrapper {
         uint256 refundETH = address(this).balance;
```
seunlanlege: This is still not correct though

seunlanlege (changes requested):

evm/src/utils/uniswapv2/UniV4UniswapV2Wrapper.sol
royvardhan: fixed

seunlanlege (approved):

PR #897, opened 2026-05-21
[evm]: IntentGatewayV2 fixes

seunlanlege:
## Summary

EVM fixes to the IntentGatewayV2 / intents-v2 stack, plus the test and deploy-script changes needed to match them.

- Renamed `NewDeployment` struct → `Deployment` (field `stateMachineId` → `chain`, raw bytes hashed internally).
- `setParams(Params)` replaced by one-time `init(Params, Deployment[])` for seeding initial cross-chain peers.
- `_instance` no longer falls back to `address(this)`; unregistered chains now revert `UnknownInstance()`.
- `DestinationFee` now carries raw `chain` bytes; events reshaped (`OrderPlaced` uses `string` source/destination, `DeploymentAdded`, `DestinationProtocolFeeUpdated(string,uint256)`).

## Tests

Updated the foundry suites (`IntentGatewayV2Test`, `IntentGatewayV2SameChainTest`, `IntrinsicIntentsReentrancyTest`, `SolverAccountTest`) to the new types/auth: migrated `setParams` → `init`, seeded peer deployments where cross-chain resolution is exercised, and corrected stale event signatures.

`forge test`: **257 passed, 0 failed, 2 skipped** (21 suites).

## Note

The SDK interface `sdk/packages/core/contracts/apps/IntentGatewayV2.sol` still declares the pre-rename event signatures and drifts from the `IntentsBase` implementation events — not a compile error, but worth realigning separately.

### DIFF
--- a/evm/src/apps/intentsv2/IntentsBase.sol
+++ b/evm/src/apps/intentsv2/IntentsBase.sol
@@ -23,7 +23,7 @@ import {
     SweepDust,
     WithdrawalRequest,
     SelectOptions,
-    NewDeployment
+    Deployment
 } from "@hyperbridge/core/apps/IntentGatewayV2.sol";
 
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
@@ -183,6 +183,11 @@ abstract contract IntentsBase is EIP712 {
      */
     error UnknownOrder();
 
+    /*
+     * @dev Thrown when the cross-chain peer is unknown.
+    */
+    error UnknownInstance();
+
     /**
      * @dev Emitted when a new intent order is placed and input tokens are escrowed.
      * @param user The order creator's address encoded as bytes32.
@@ -199,8 +204,8 @@ abstract contract IntentsBase is EIP712 {
      */
     event OrderPlaced(
         bytes32 user,
-        bytes source,
-        bytes destination,
+        string source,
+        string destination,
         uint256 deadline,
         uint256 nonce,
         uint256 fees,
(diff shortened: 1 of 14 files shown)


evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -506,15 +512,15 @@ abstract contract IntentsBase is EIP712 {
         _params = update.params;
 
         for (uint256 i; i < update.destinationFees.length;) {
-            bytes32 stateMachineId = update.destinationFees[i].stateMachineId;
+            bytes memory chain = update.destinationFees[i].chain;
```
Wizdave97, line 515: Pallet Intent coprocessor still sends keccak hash of the chain, it's types have not been updated.

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -199,8 +204,8 @@ abstract contract IntentsBase is EIP712 {
      */
     event OrderPlaced(
         bytes32 user,
-        bytes source,
-        bytes destination,
+        string source,
```
royvardhan, line 207: Hmmm, this can potentially yield different offchain commitment, resulting in a mismatch for orders filled on the dest chain. Worth updating the sdk.

seunlanlege: done

evm/src/apps/intentsv2/IntentsBase.sol
seunlanlege: done

Wizdave97 (approved):

royvardhan (approved):

PR #919, opened 2026-05-27
[runtime]: Filter BEEFY consensus and fund_message, and require bonded collators on carryover

dharjeezy:
The nexus and gargantua base call filters now turn away `Ismp::handle_unsigned` batches that carry a BEEFY consensus update and disable `Ismp::fund_message` outright, resolving each consensus message's state id back to its client through the host so the check stays right as more states are bound to BEEFY. 

Separately, when the nexus collator manager fills empty slots from the previous session it now only carries over collators whose stash is still bonded, so one that has unbonded or deregistered no longer keeps its slot. The filter is covered by a simnode test and the carryover by unit tests.


modules/pallets/collator-manager/src/lib.rs
```diff
@@ -424,10 +424,21 @@ pub mod pallet {
 					.map(|(_, c)| c.clone().into()),
 			);
 
-			// fill remaining slots with the best of the previous set.
+			// fill remaining slots with the best of the previous set, but only carry over
+			// collators whose stash is still bonded. One that has since unbonded or
+			// deregistered shouldn't keep its slot just because it was active last session.
```
seunlanlege: Tbh i don't think we need to do this filling remaining slots

just remove this line:

https://github.com/polytope-labs/hyperbridge/blob/9b8a6bc945e617a559f54dd5b35847b674cf5802/modules/pallets/collator-manager/src/lib.rs#L407

from

https://github.com/polytope-labs/hyperbridge/blob/9b8a6bc945e617a559f54dd5b35847b674cf5802/modules/pallets/collator-manager/src/lib.rs#L402-L410

seunlanlege (approved):

modules/pallets/collator-manager/src/lib.rs
Wizdave97: I think we should remove this, if they don't have a balance this would break block production, we should just prioritize collators with reputation tokens
```suggestion
```

modules/pallets/collator-manager/src/migrations.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Moves every collator bond from the old pallet lock to a collator-selection reserve.
+///
+/// Collator-selection now holds the candidacy bond as a plain reserve, so for each bonded
+/// account we drop the `collbond` lock and reserve the same amount, then clear the legacy
+/// ledger. Draining the ledger makes it safe to run the migration more than once.
+pub struct MigrateBondsToReserves<T>(PhantomData<T>);
+
+impl<T: Config> OnRuntimeUpgrade for MigrateBondsToReserves<T> {
+	fn on_runtime_upgrade() -> Weight {
```
Wizdave97: This seems like it would run everytime, there's no storage versioning

parachain/runtimes/gargantua/src/lib.rs
```diff
@@ -571,7 +575,17 @@ impl pallet_collator_selection::Config for Runtime {
 	type KickThreshold = Period;
 	type ValidatorId = <Self as frame_system::Config>::AccountId;
 	type ValidatorIdOf = pallet_collator_selection::IdentityCollator;
-	type ValidatorRegistration = Session;
+	type ValidatorRegistration = CollatorManager;
+	type WeightInfo = ();
+}
+
+impl pallet_collator_manager::Config for Runtime {
```
Wizdave97, line 588: This would not work for gargantua, we use invulnerables, if you want to add this to gargantua, merge invulnerables to the validator set

parachain/runtimes/nexus/src/lib.rs
```diff
@@ -173,6 +173,7 @@ pub type Migrations = (
 	pallet_mmr_tree::migrations::ResetMmrTree<Runtime>,
```
Wizdave97, line 173: You can remove the old migrations

modules/pallets/collator-manager/src/migrations.rs
Wizdave97: Also kill the BondedStorage when it's done.

parachain/runtimes/nexus/src/lib.rs
```diff
@@ -733,6 +736,46 @@ impl Contains<RuntimeCall> for ReputationCallFilter {
 	}
 }
 
+/// Switches off two ismp calls on nexus. `fund_message` is disabled outright,
+/// and `handle_unsigned` is turned away whenever its batch carries a consensus
+/// update for the BEEFY client.
+///
+/// A consensus message only names the state it updates, so we ask the host which
```
Wizdave97: Add a comment explaining why we filter out the beefy consensus updates and fund_message.

parachain/runtimes/nexus/src/lib.rs
```diff
@@ -733,6 +736,46 @@ impl Contains<RuntimeCall> for ReputationCallFilter {
...
+					_ => false,
+				})
+			},
+			_ => true,
+		}
+	}
+}
+
+/// Blocks collator-selection's `leave_intent`. Collators leave through collator-manager's
+/// `unbond`, which enforces the seven day unbonding delay, so the instant exit is closed off.
```
Wizdave97: Same here, add a comment why we filter out `leave_intent`

Wizdave97: Also updte the docs with the unbonding flow

parachain/runtimes/nexus/src/lib.rs
```diff
@@ -733,6 +736,51 @@ impl Contains<RuntimeCall> for ReputationCallFilter {
 	}
 }
 
+/// Nexus routes all BEEFY consensus updates through `pallet-beefy-consensus-proofs`, which
+/// requires each proof to pass SP1 zkVM verification before it can advance the BEEFY state.
+/// Allowing raw updates through `handle_unsigned` would bypass that requirement entirely, so
+/// any batch that carries a BEEFY consensus message is rejected here. `fund_message` is also
+/// disabled because nexus uses the bandwidth model for request fees; per-message top-ups have
```
Wizdave97:
```suggestion
/// disabled because it will change the child trie root allowing beefy proofs that have no economic value
```

parachain/runtimes/nexus/src/lib.rs
```diff
@@ -733,6 +736,51 @@ impl Contains<RuntimeCall> for ReputationCallFilter {
...
+/// no role in that accounting.
```
Wizdave97:
```suggestion
```

Wizdave97 (approved):

PR #938, opened 2026-06-01
[tesseract]: claim oldest rewards first in outbound claims task

dharjeezy:
Merged result is now sorted by `delivery_height` ascending before processing, so the relayer always works through its oldest pending rewards first rather than in an arbitrary hash or commitment-key order.

### DIFF
--- a/tesseract/relayer/Cargo.toml
+++ b/tesseract/relayer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tesseract"
-version = "2.4.0"
+version = "2.5.0"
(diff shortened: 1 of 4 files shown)


tesseract/relayer/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tesseract"
-version = "2.4.0"
+version = "2.5.0"
```
seunlanlege: patch

Wizdave97 (approved):

PR #941, opened 2026-06-01
[runtime]: filter unwanted calls in call decompressor

dharjeezy:
The call decompressor bypassed the runtime's `BaseCallFilter` by invoking pallet internals directly rather than going through normal dispatch. This wires the filter into both `validate_unsigned` and `decode_and_execute` so any call the runtime would block is consistently rejected at the txpool and execution layers. The `BridgeDrop` pallet has also been removed entirely from both runtimes and the workspace.


Wizdave97 (commented):
Bump runtime spec

Wizdave97 (approved):

PR #943, opened 2026-06-02
[simplex]: delegation balance checks + fx pricing fixes

royvardhan:
Two independent simplex fixes:

**Delegation** (`DelegationService`, paymaster)
Pre-check balances before attempting EIP-7702 delegation and log the specific shortfall instead of a generic failure:
- USDC balance on the bundler/paymaster path (paymaster can't sponsor gas without it)
- Native ETH on the direct-tx fallback
- Fixed the misleading "no paymaster available" log when the paymaster *is* configured but USDC is short. Shared the balance check via `getUsdcBalanceStatus`.

**FX pricing** (`fx.ts`, `simplex.ts`)
Fixes orders being wrongly rejected during profitability evaluation:
- **Leg-index desync:** the valuation loop indexed `fillerOutputs[i]` against `order.inputs[i]`/`pairs[i]`, but skipped legs aren't pushed to `fillerOutputs` — so on a multi-leg order with a skipped leg it applied the wrong bid/ask curve. Now realigns via the original leg index.

### DIFF
--- a/sdk/packages/simplex/src/bin/simplex.ts
+++ b/sdk/packages/simplex/src/bin/simplex.ts
@@ -587,12 +587,13 @@ function validateConfig(config: FillerTomlConfig): void {
 				)
 			}
 
-			const hasStaticCurves = bidLen >= 2 && askLen >= 2
+			// A single point is a valid flat curve — FillerPricePolicy returns that price at every size.
+			const hasStaticCurves = bidLen >= 1 && askLen >= 1
 			const hasUniswapV4Positions = (strategy.vault?.uniswapV4?.positions?.length ?? 0) > 0
 
 			if (!hasStaticCurves && !hasUniswapV4Positions) {
 				throw new Error(
-					"hyperfx: provide bid+ask price curves (≥2 points each) or configure [strategies.vault.uniswapV4].positions for pool-based pricing",
+					"hyperfx: provide bid+ask price curves (≥1 point each) or configure [strategies.vault.uniswapV4].positions for pool-based pricing",
 				)
 			}
 
@@ -603,9 +604,6 @@ function validateConfig(config: FillerTomlConfig): void {
 			}
 
 			if (bidLen > 0) {
-				if (!hasStaticCurves) {
-					throw new Error("hyperfx: bid and ask price curves must each have at least 2 points when provided")
-				}
 				for (const point of strategy.bidPriceCurve!) {
 					if (point.amount === undefined || point.price === undefined) {
 						throw new Error("Each FX bidPriceCurve point must have 'amount' and 'price'")
--- a/sdk/packages/simplex/src/bin/simplex.ts
+++ b/sdk/packages/simplex/src/bin/simplex.ts
@@ -9,7 +9,6 @@ import { BasicFiller } from "@/strategies/basic"
 import { FXFiller } from "@/strategies/fx"
 import type { FundingVenue, UniswapV4PositionConfig } from "@/funding/types"
 import { UniswapV4FundingPlanner } from "@/funding/uniswapV4/UniswapV4FundingPlanner"
-import Decimal from "decimal.js"
 import { ConfirmationPolicy, FillerBpsPolicy, FillerPricePolicy } from "@/config/interpolated-curve"
 import { ChainConfig, FillerConfig, HexString } from "@hyperbridge/sdk"
 import {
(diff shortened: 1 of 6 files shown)


sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -616,6 +617,27 @@ function validateConfig(config: FillerTomlConfig): void {
...
+
+				// Prices are quoted as "exotic per USD" (an inverted price), so a profitable
+				// spread requires bid ≥ ask at every order size: the filler must hand out fewer
+				// exotic when selling (ask) than it demands when buying (bid). bid < ask is an
+				// inverted spread that loses money on every fill — reject it up front instead of
+				// silently filling at a loss. Both curves are piecewise-linear, so the bid−ask
+				// difference is minimised at a breakpoint; checking the union of points suffices.
+				const bidPolicy = new FillerPricePolicy({ points: strategy.bidPriceCurve! })
+				const askPolicy = new FillerPricePolicy({ points: strategy.askPriceCurve! })
+				const breakpoints = [...strategy.bidPriceCurve!, ...strategy.askPriceCurve!].map((p) => new Decimal(p.amount))
```
Wizdave97: I think the bid and ask policy should be independent of each other, we should not mix them in any way.

We should only ensure that each curve is configured correctly independently.

Mixing them is what is causing profitability calculation issues.

Wizdave97: Let each curve operate independently, we should consult bid curve exclusively for bids and ask curve exclusively for asks

Wizdave97: It's up to the operator to configure profitable prices that makes the bid and asks profitable.

royvardhan: Done in ef9711b5 — curves are fully decoupled now:

- Removed this `bid ≥ ask` config guard entirely. Each curve is only validated independently (≥2 well-formed points); their relationship is left to the operator.
- Removed the cross-curve profitability gauge in `calculateProfitability` (the `output USD ≥ input USD` reject). Output is still generated from a single curve per side (ask for sells, bid for buys), given to the user directly.
- The bid decision now gates only on `order.fees` covering estimated execution cost; FX margin is trusted to the curves.
- The bid/ask spread value is still computed but is **report-only** (`fxMarginUsd` in the log + added to the returned profit so the bid fires when fee profit is 0). It never rejects an order. Happy to drop even the report-only calc if you'd prefer zero cross-curve math in the file.

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -484,43 +491,37 @@ export class FXFiller implements FillerStrategy {
...
 			}
-
-			this.recordOrderOutcome(anyLegClamped, order.id)
-
-			const { totalCostInSourceFeeToken } = await this.contractService.estimateGasFillPost(order)
-			const feeProfit = order.fees > totalCostInSourceFeeToken ? order.fees - totalCostInSourceFeeToken : 0n
+			const feeProfit = order.fees - totalCostInSourceFeeToken
 			const feeProfitParsed = parseFloat(formatUnits(feeProfit, feeTokenDecimals))
-			const totalProfit = feeProfitParsed + spreadProfitUsd.toNumber()
+			const totalProfit = feeProfitParsed + fxMarginUsd.toNumber()
```
Wizdave97: let's not add the fxMargin to the reported profit for an order, we can report it seperately as it's own value, it's a theoretical profit value.

royvardhan: So you mean to disable this logging.

seunlanlege:
Meanwhile, i think we can allow the price curve take only 1 point & duplicate it for the purposes of creating a curve

sdk/packages/simplex/src/strategies/fx.ts
Wizdave97: I mean dont add `fxMarginUsd` to `feeProfitParsed`

royvardhan: Roger

royvardhan: `fxMarginUsd` is no longer summed into the returned profit. `totalProfit` is now fee profit only (`order.fees − gas`); `fxMarginUsd` stays in the log as its own theoretical value. FX bids are therefore gated purely on fee profit.

royvardhan:
@seunlanlege single-point price curves are now allowed. No duplication needed: `FillerPricePolicy` already returns the lone point's price at every order size, so one point behaves as a flat curve. Just relaxed the CLI validation to accept ≥1 point.

Wizdave97 (approved):

PR #944, opened 2026-06-02
[runtime]:  Anchor non-existence sibling proofs at the deepest ancestor

dharjeezy:
When the terminal internal node is all-zero, sibling evidence is now sourced from the deepest ancestor with non-empty non-path slots rather than the (empty) terminal, matching the actual proof format returned by the Pharos RPC. This closes the case where a real membership proof could be reused as a fake non-existence proof by moving a child hash to a sibling slot and appending an all-zero terminal.

### DIFF
--- a/modules/consensus/pharos/primitives/src/spv.rs
+++ b/modules/consensus/pharos/primitives/src/spv.rs
@@ -262,9 +262,10 @@ pub fn verify_membership_proof(
 
 /// Verify that a key does NOT exist in the trie (non-inclusion proof).
 ///
-/// Case 1: Proof ends at a leaf with a different key_hash (path collision).
-/// Case 2: Proof ends at an internal node where the target nibble slot is empty.
-///         Sibling proofs pin the non-empty slots to the same root, preventing forgery.
+/// Case 1: proof ends at a leaf with a different key hash (path collision).
+/// Case 2: proof ends at an internal node where the queried slot is empty. When the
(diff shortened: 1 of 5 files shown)


Wizdave97:
Prover/liveness completeness (flag this). The new branch makes honest all-zero-terminal non-existence proofs that pass through a branching ancestor now require anchor sibling proofs — previously they needed none. The off-chain Pharos proof generator must populate SiblingLeftmostLeafProof for the anchor's non-path slots, not just the terminal's. If the prover side isn't updated to match, legitimate GET-timeout / non-membership proofs could fail to verify. Notably, the tests only cover the no-branching honest case (test_non_existence_case2_empty_slot_all_zero_terminal) and the rejection case — there's no positive test for an honest all-zero proof that requires anchor siblings. I'd recommend confirming the prover change exists and adding that positive test.
Minor footgun: the two branches use different parent_end conventions — the terminal-anchor case excludes the anchor from parent_nodes (expecting the prover to re-supply it as sib.proof_path[0]), while the ancestor-anchor case includes it. Both are correct, but the asymmetry is easy to break in a future edit.

Wizdave97:
@dharjeezy  is this validated against the rpc response for honest zero terminal nodes?

dharjeezy:
> @dharjeezy is this validated against the rpc response for honest zero terminal nodes?

Yes it is

Wizdave97 (approved):

PR #945, opened 2026-06-02
[runtime]:  replace panicking unwraps with proper error propagation across runtime pallets and modules

dharjeezy:
Converts all .expect() calls in production pallet and ISMP module code to return Result or propagate errors via ?, using existing error variants where available.

### DIFF
--- a/modules/ismp/core/src/handlers/timeout.rs
+++ b/modules/ismp/core/src/handlers/timeout.rs
@@ -125,8 +125,8 @@ where
 						// Module callback failed; restore commitment so the request
 						// can be retried.
 						host.store_request_commitment(&request, meta)?;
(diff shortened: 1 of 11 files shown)


modules/ismp/core/src/handlers/timeout.rs
```diff
@@ -125,8 +125,8 @@ where
 						// Module callback failed; restore commitment so the request
 						// can be retried.
 						host.store_request_commitment(&request, meta)?;
-						if host.host_state_machine() != post.source && signer.is_some() {
-							host.store_request_receipt(&request, &signer.expect("Infaliible"))?;
```
Wizdave97, line 129: let's use ok_or_else instead of changing this logic here.

Wizdave97 (approved):

PR #948, opened 2026-06-03
[simplex]: add solver allowlist for order user addresses

### LINKED ISSUE #947: [simplex]: solver allowlist for order user addresses
## Background

Simplex currently considers every order surfaced by `EventMonitor` and tries to fill it if any configured strategy returns `canFill === true` and the order is profitable. The pipeline lives in `IntentFiller.handleNewOrder` (`sdk/packages/simplex/src/core/filler.ts`) and never inspects `order.user`.

A solver operator that has bilateral arrangements with specific aggregators / wallets — or that needs to restrict served traffic for compliance / KYC reasons — has no way today to tell the filler \"only quote for these users\". They have to either run a fork or post-filter externally.

## Proposed change

Add an optional user allowlist to the filler config. When set, `IntentFiller.handleNewOrder` short-circuits any order whose `order.user` is not on the list, before bidding or filling work begins. Empty / absent allowlist preserves today's \"accept all\" behaviour.

### Config

Extend `filler-config-example.toml` and `FillerConfig` (`sdk/packages/simplex/src/services/FillerConfigService.ts`) with an `[allowlist]` section, e.g.:

```toml
[allowlist]
# Global allowlist - any order whose `order.user` is in this list is eligible
users = [
    \"0xAggregatorA...\",
    \"0xAggregatorB...\",
]

# Optional per-source-chain overrides; merged on top of `users`
[allowlist.bySource]
\"EVM-1\"  = [\"0xMainnetOnlyUser...\"]
\"EVM-56\" = [\"0xBscOnlyUser...\"]
```

Comparisons should be case-insensitive on the hex address.

### Enforcement

In `IntentFiller.handleNewOrder`, before `verifyOrderOnSource` runs, call something like `configService.isUserAllowed(order.user, order.source)` and `return` early with a debug log when it returns false. This keeps the check cheap and avoids spending RPC budget on orders we'd never bid on.

### Surface

- `FillerConfigService.getAllowlist()` / `isUserAllowed(user, chain)` — new accessors.
- `IntentFiller.handleNewOrder` — early reject + log line (`orderId`, `user`, `source`).
- `filler-config-example.toml` — documented `[allowlist]` block with explicit \"omit to allow all\" comment.

## Motivation

- Solvers running on top of partner aggregators want to commit liquidity to that orderflow without serving the public mempool.
- Cuts wasted work (canFill, profitability, USD pricing, gas estimation) for orders the solver was never going to fill.
- Provides a hook compliance / KYC operators can wire allowlists into without forking the filler.

## Notes / future extension

A symmetrical `blocklist` could live in the same config block — out of scope here, but the accessor shape (`isUserAllowed`) should leave room for it.

royvardhan:
Restricts which order `user` addresses the filler processes, configured via a new optional `[allowlist]` section. Omitting it preserves accept-all behavior.

- Global `users` list merged with per-source overrides in `[allowlist.bySource]` (keyed by state machine id, e.g. `EVM-1`); case-insensitive.
- A present-but-empty merged set for a chain rejects every order; a fully-absent `[allowlist]` accepts all.
- Early rejection in `IntentFiller.handleNewOrder` before `verifyOrderOnSource`.

Closes #947

### DIFF
--- a/sdk/packages/simplex/src/services/FillerConfigService.ts
+++ b/sdk/packages/simplex/src/services/FillerConfigService.ts
@@ -108,6 +108,13 @@ export interface OverfillProtectionConfig {
 	maxConsecutiveClamps?: number
 }
 
+export interface AllowlistConfig {
+	/** Hex addresses eligible across all chains. */
+	users?: string[]
+	/** Per-source-chain overrides, keyed by state machine id (e.g. "EVM-1"). Merged with the global list. */
+	bySource?: Record<string, string[]>
+}
+
 export interface FillerConfig {
 	maxConcurrentOrders: number
 	logging?: LogLevel
@@ -131,6 +138,12 @@ export interface FillerConfig {
 	 * (maxOverfillBps=100, maxConsecutiveClamps=3).
 	 */
 	overfillProtection?: OverfillProtectionConfig
+	/**
+	 * Restricts which order `user` addresses the filler processes. When omitted, all
+	 * users are accepted. When present, only listed users (global ∪ per-source) are
+	 * accepted; a chain whose merged set is empty rejects every order.
+	 */
+	allowlist?: AllowlistConfig
 }
 
 /**
@@ -142,6 +155,10 @@ export class FillerConfigService {
 	private rpcOverrides: Map<number, string[]> = new Map()
 	private bundlerUrls: Map<number, string> = new Map()
 	private fillerConfig?: FillerConfig
+	/** Lowercased global allowlist users, or undefined when no allowlist is configured. */
+	private allowlistGlobal?: Set<string>
+	/** Lowercased per-source allowlist users, keyed by state machine id. */
+	private allowlistBySource: Map<string, Set<string>> = new Map()
 
 	constructor(chainConfigs: ResolvedChainConfig[], fillerConfig?: FillerConfig) {
 		chainConfigs.forEach((config) => {
@@ -157,6 +174,31 @@ export class FillerConfigService {
 
 		this.chainConfigService = new ChainConfigService({})
 		this.fillerConfig = fillerConfig
+
+		const allowlist = fillerConfig?.allowlist
+		if (allowlist) {
+			this.allowlistGlobal = new Set((allowlist.users ?? []).map((u) => u.toLowerCase()))
+			for (const [chain, users] of Object.entries(allowlist.bySource ?? {})) {
+				this.allowlistBySource.set(chain, new Set(users.map((u) => u.toLowerCase())))
+			}
+		}
+	}
+
+	getAllowlist(): AllowlistConfig | undefined {
+		return this.fillerConfig?.allowlist
+	}
+
+	/**
+	 * Whether an order from `user` on `chain` (a source state machine id, e.g. "EVM-1")
+	 * may be processed. Returns true when no allowlist is configured. Otherwise the user
+	 * must appear in the global list or the per-source override for that chain; an empty
+	 * merged set rejects every order on that chain.
+	 */
+	isUserAllowed(user: string, chain: string): boolean {
+		if (!this.allowlistGlobal) return true
+		const normalized = user.toLowerCase()
+		if (this.allowlistGlobal.has(normalized)) return true
+		return this.allowlistBySource.get(chain)?.has(normalized) ?? false
 	}
 
 	getChainConfig(chain: string): ChainConfig {
--- a/sdk/packages/simplex/src/services/FillerConfigService.ts
+++ b/sdk/packages/simplex/src/services/FillerConfigService.ts
@@ -155,10 +155,10 @@ export class FillerConfigService {
 	private rpcOverrides: Map<number, string[]> = new Map()
 	private bundlerUrls: Map<number, string> = new Map()
 	private fillerConfig?: FillerConfig
-	/** Lowercased global allowlist users, or undefined when no allowlist is configured. */
+	/** Lowercased global allowlist users, or undefined when no global list is configured. */
 	private allowlistGlobal?: Set<string>
-	/** Lowercased per-source allowlist users, keyed by state machine id. */
-	private allowlistBySource: Map<string, Set<string>> = new Map()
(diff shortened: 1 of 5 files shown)


sdk/packages/simplex/src/services/FillerConfigService.ts
```diff
@@ -157,6 +174,31 @@ export class FillerConfigService {
...
+	}
+
+	/**
+	 * Whether an order from `user` on `chain` (a source state machine id, e.g. "EVM-1")
+	 * may be processed. Returns true when no allowlist is configured. Otherwise the user
+	 * must appear in the global list or the per-source override for that chain; an empty
+	 * merged set rejects every order on that chain.
+	 */
+	isUserAllowed(user: string, chain: string): boolean {
+		if (!this.allowlistGlobal) return true
```
Wizdave97: Shouldn't this be?
```suggestion
		if (!this.allowlistGlobal &&  !allowlistBySource) return true
```

royvardhan: `allowlistGlobal` was set whenever `[allowlist]` existed, so `!this.allowlistGlobal` already covered it — just implicit. Latest commit makes both sets optional and checks both.

Wizdave97 (approved):

Wizdave97:
@royvardhan 

## Review of PR #948 — Solver allowlist for order user addresses

I reviewed the diff against the runtime data it operates on. There is **one critical correctness bug** that makes the feature reject every EVM order whenever an allowlist is configured.

### 🔴 Critical: `order.user` is `bytes32`, but the allowlist stores 20-byte addresses — they never match

The production call site passes the raw order user:

```ts
// filler.ts:369
if (!this.configService.isUserAllowed(order.user, order.source)) { ... return }
```

`order.user` originates from the `OrderPlaced` event, whose `user` field is **`bytes32`** ([[IntentGatewayV2.ts](https://claude.ai/epitaxy/sdk/packages/simplex/src/config/abis/IntentGatewayV2.ts)](sdk/packages/simplex/src/config/abis/IntentGatewayV2.ts) → `name: "user", type: "bytes32"`), and it is stored verbatim in [[event-monitor.ts:52](https://claude.ai/epitaxy/sdk/packages/simplex/src/core/event-monitor.ts)](sdk/packages/simplex/src/core/event-monitor.ts) (`user: decodedLog.args.user`). So at runtime `order.user` is the **66-char, left-padded** form, e.g.:

```
0x000000000000000000000000aaaa...aaaa   (32 bytes)
```

But the allowlist is configured and validated as **20-byte addresses** (`isAddress(user)` in [[simplex.ts:545](https://claude.ai/epitaxy/sdk/packages/simplex/src/bin/simplex.ts)](sdk/packages/simplex/src/bin/simplex.ts), and the TOML example uses `0x1111…1111` = 42 chars). The set is built from those raw 42-char strings:

```ts
// FillerConfigService.ts
this.allowlistGlobal = new Set(allowlist.users.map((u) => u.toLowerCase()))
...
isUserAllowed(user, chain) {
  const normalized = user.toLowerCase()          // 66-char bytes32
  if (this.allowlistGlobal?.has(normalized)) ...  // set holds 42-char addrs → never hits
  return this.allowlistBySource?.get(chain)?.has(normalized) ?? false
}
```

`"0x000…000aaaa…aaaa".toLowerCase()` (66 chars) will never equal `"0xaaaa…aaaa"` (42 chars) in the `Set`. **Result: as soon as any allowlist is configured, `isUserAllowed` returns `false` for every EVM order, and the filler silently skips all of them.** The feature does the opposite of "restrict to listed users" — it blocks everyone.

Notably, the codebase already imports the needed converter in the same file — `bytes32ToBytes20` is used at [[filler.ts:313](https://claude.ai/epitaxy/sdk/packages/simplex/src/core/filler.ts)](sdk/packages/simplex/src/core/filler.ts) for token addresses — but it is not applied to `user` on the allowlist path.

**Why the tests don't catch it:** every assertion in `FillerConfigService.allowlist.test.ts` calls `isUserAllowed(USER_A, …)` with a **20-byte** address (`0x1111…1111`), i.e. the config format, not the bytes32 form that `filler.ts` actually passes. The unit tests validate the comparison against the wrong input shape, so they pass green while the integration is broken. There's no test that feeds a `bytes20ToBytes32(...)` value through `isUserAllowed`.

**Fix:** normalize both sides to one canonical form. Simplest is to store the config addresses in the same bytes32 form the orders use, via the SDK's `normalizeAddressForEvmBytes32`:

```ts
import { normalizeAddressForEvmBytes32 } from "@hyperbridge/sdk"

this.allowlistGlobal = new Set(
  allowlist.users.map((u) => normalizeAddressForEvmBytes32(u).toLowerCase()),
)
// same for bySource
```

(or convert `order.user` with `bytes32ToBytes20(...)` inside `isUserAllowed`). Then add a test that passes a `bytes20ToBytes32(addr)` value to lock in the real call shape.

### Things that are correct
- **`source` keying is fine.** `order.source` is `normalizeStateMachineId(...)`, which yields the human-readable `"EVM-1"` form ([[utils.ts:275](https://claude.ai/epitaxy/sdk/packages/simplex/src/utils.ts)](sdk/packages/simplex/src/utils.ts)) — matching the `bySource` keys. No bug there.
- **Independently-optional global/per-source sets** behave as documented; the `{ users: [] }` "present-but-empty rejects all" semantics are intentional and called out in the example TOML and doc comments.
- **Config validation** (`isAddress`, `bySource` array check) is reasonable, and case-insensitive matching is handled correctly *within* each format.

### Recommendation
Block on the critical bug — the feature is non-functional as written (it rejects all EVM orders when enabled). Once `order.user`'s bytes32 form is normalized against the config addresses and a test exercises that exact shape, the rest of the change is sound.

One caveat on scope: I reviewed the logic against the data formats. I did not run the simplex test suite here (it's the source-linked SDK workspace); the analysis is from reading the ABI, event reconstruction, and the conversion helpers, all of which agree that `order.user` is the 32-byte form.

royvardhan:
Can we enable LLM reviews in the pull request pipeline?

Wizdave97 (approved):

PR #949, opened 2026-06-03
[sdk]: yield executable Bid objects from the intents SDK

### LINKED ISSUE #946: [sdk]: yield executable bid objects from the intents SDK
## Background

Today the intents SDK encapsulates the entire bid lifecycle inside the SDK. After a consumer calls `IntentGateway.execute()`, the executor polls the coprocessor for bids and then internally:

1. Decodes & validates each `FillerBid`, sorts them by output value (single-token amount, normalised stables USD, or DEX-quoted USD for mixed baskets — see `BidManager.validateAndSortBids`).
2. Iterates the sorted list, simulating each one via `eth_call` until one passes.
3. Signs the EIP-712 `SelectSolver` message with the session key.
4. Appends that signature to the solver's UserOp signature and submits to the bundler via `eth_sendUserOperation`.

All of this happens in `BidManager.selectBid` (`sdk/packages/sdk/src/protocols/intents/BidManager.ts`), driven by the loop in `OrderExecutor.executionStream` (`sdk/packages/sdk/src/protocols/intents/OrderExecutor.ts`). The consumer only sees opaque `BIDS_RECEIVED` / `BID_SELECTED` status updates — they can't inspect bids in a structured way, can't apply their own ranking (preferred solvers, reputation, KYT, etc.), and can't atomically batch the bid execution with other actions.

## Proposed change

Lift bid selection out of the SDK and yield first-class `Bid` objects to the consumer. Each bid should expose:

- `execute(): Promise<SelectBidResult>` — signs the `SelectSolver` message with the session key, packs the final UserOp signature, and submits to the bundler. Returns the same `SelectBidResult` shape we have today (userOpHash, solver address, txnHash, fillStatus, filledAssets).
- `simulate(): Promise<void>` — runs the existing batched `select` + `fillOrder` `eth_call` simulation so consumers can pre-check a bid before showing it to a user.
- Metadata accessors / fields:
  - `solverAddress: HexString`
  - `outputs: TokenInfo[]` — decoded `FillOptions.outputs`
  - `relayerFee`, `nativeDispatchFee` from the decoded fill options
  - `outputUsdValue(): Promise<Decimal | null>` — reuse the existing pricing helpers in `BidManager`
  - `userOp: PackedUserOperation` — raw access for advanced consumers

Generator semantics:

- `AWAITING_BIDS` → `NEW_BID { bid: Bid }` / `BIDS_RECEIVED { bids: Bid[] }` — consumer picks a bid and calls `bid.execute()` when ready.
- Existing fill-tracking statuses (`FILLED`, `PARTIAL_FILL`, `EXPIRED`, `FAILED`) stay the same and are still emitted when partial fills require continued bidding.

For consumers that want the current behaviour, keep a thin helper (e.g. `IntentGateway.selectAndExecuteBest(order, bids)`) that wraps the existing sort-and-execute path on top of the new `Bid` API — same code, but optional.

## Motivation

- Aggregators / wallets want to display bid metadata to end users and let *them* choose, or apply business rules (preferred solver list, reputation scores, regional restrictions).
- Custom auction logic — e.g. accept the first bid that beats a threshold rather than always picking the highest.
- Atomicity — `bid.execute()` exposed as a primitive means consumers can wrap it inside their own batched txs / paymaster flows.

## Affected files

- `sdk/packages/sdk/src/protocols/intents/BidManager.ts` — refactor `selectBid` into per-bid `execute()` plus helpers (`sortBids`, `simulate`, `submit`); keep current sort/price helpers as public utilities.
- `sdk/packages/sdk/src/protocols/intents/OrderExecutor.ts` — stop calling `selectBid` directly; surface `Bid` instances in the executor stream and let consumers drive selection.
- `sdk/packages/sdk/src/protocols/intents/IntentGateway.ts` — update `execute` / `resume` generator yield types; optionally add the convenience helper described above.
- `sdk/packages/sdk/src/types/index.ts` — new `Bid` interface, updated `IntentOrderStatusUpdate` shape.

royvardhan:
Closes #946.

Lifts bid selection out of the SDK and yields first-class `Bid` objects (`execute()`, `simulate()`, `outputUsdValue()`, `solverAddress`, `outputs`, `relayerFee`, `nativeDispatchFee`, `userOp`) so consumers can apply their own ranking and atomically batch bid execution.

`execute()`/`resume()` are now bidirectional: they yield `Bid[]` and the consumer feeds the executed `SelectBidResult` back via `gen.next(result)`.

The historical auto-select path is preserved on top of the new API:
- `selectBid(order, bids)` / `selectAndExecuteBest(order, bids)` — one-shot sort + simulate + sign + submit.
- `executeBest()` / `resumeBest()` — batteries-included streams that auto-select each round.

## TODO
- [x] E2E test — verified live on testnet (BSC Chapel → Polygon Amoy):
  - Autopilot (`executeBest`) path — `basic.testnet.test.ts`
  - Consumer-driven path (`execute` → inspect `Bid` metadata → `bid.execute()` → feed result back) — `fx.testnet.test.ts`


Wizdave97:
## Review of PR #949 — "yield executable Bid objects from the intents SDK"

This is a well-structured refactor: it extracts a first-class `BidImpl` (`Bid.ts`) out of `BidManager`, converts the executor into a **bidirectional generator** so consumers own `bid.execute()`, and adds `buildBids` / `sortBids` / `selectAndExecuteBest` plus the `executeBest` / `resumeBest` autopilot wrappers. The mechanical extraction (sorting helpers, simulate logic, fill tracking) is faithful, and the generator feedback-threading (`AWAITING_PLACE_ORDER` ↔ signed tx, `BIDS_RECEIVED` ↔ `SelectBidResult`) is wired correctly through `executeBest → execute → driveExecution → executeOrder → executionStream`.

But there are a few correctness/behavior issues, one of which I'd consider blocking.

### 🔴 1. `executeBest` drops the auction's retry-on-failure resilience (regression vs. stated intent)

The old auto-select path ran selection **inside** `executionStream`, wrapped in try/catch:

```ts
yield BIDS_RECEIVED
try { submitResult = await this.submitBid(...) }   // selectBid → may throw "No bids passed simulation"
catch (err) {
  yield { status: "FAILED", ... }
  // increment bidFailCounts for top bid (MAX_BID_ATTEMPTS = 2)
  await sleep(pollIntervalMs); continue              // keep polling for better bids until deadline
}
```

In the new design, `executeBest` calls the throwing `selectAndExecuteBest` **outside** the generator with no guard ([[IntentGateway.ts](https://claude.ai/epitaxy/sdk/packages/sdk/src/protocols/intents/IntentGateway.ts)](sdk/packages/sdk/src/protocols/intents/IntentGateway.ts)):

```ts
if (value.status === "BIDS_RECEIVED") {
  yield value
  input = await this.selectAndExecuteBest(order, value.bids)   // throws → propagates, unhandled
}
```

`selectAndExecuteBest` throws `"No valid bids found"` / `"No bids passed simulation"` (and bundler errors bubble out of `bid.execute()`). So if the **first** `BIDS_RECEIVED` round's bids all fail simulation, `executeBest` aborts the whole order — instead of looping back to poll for fresh bids until the deadline, as the old code did. The `executeBest` docstring explicitly claims it "preserves the SDK's historical 'pick the best and go' behaviour" / "the SDK's historical behaviour," but it does not.

The executor itself already supports the correct continuation contract — feeding back `undefined` after `BIDS_RECEIVED` means "didn't execute, keep polling" ([[OrderExecutor.ts:399](https://claude.ai/epitaxy/sdk/packages/sdk/src/protocols/intents/OrderExecutor.ts)](sdk/packages/sdk/src/protocols/intents/OrderExecutor.ts)). So the faithful fix is for `executeBest`/`resumeBest` to catch and continue:

```ts
if (value.status === "BIDS_RECEIVED") {
  yield value
  try { input = await this.selectAndExecuteBest(order, value.bids) }
  catch (err) { /* log/observe */ input = undefined }   // keep polling until deadline
}
```

Relatedly, the per-bid `MAX_BID_ATTEMPTS = 2` retry tracking (`bidFailCounts`) is removed entirely. That may be an intentional simplification, but combined with the above it means a transient failure on the top bid is no longer retried at all in autopilot mode. Worth a deliberate call-out in the PR description.

### 🟠 2. Generator teardown is skipped when `executeBest` throws

Because the throw in #1 propagates out while `executeOrder` and `executionStream` are suspended at a `yield`, their `finally` teardown blocks (the explicit `executionStream.return()` / `deadlineTimeout.return()` at [[OrderExecutor.ts](https://claude.ai/epitaxy/sdk/packages/sdk/src/protocols/intents/OrderExecutor.ts)](sdk/packages/sdk/src/protocols/intents/OrderExecutor.ts)) never run. The streams are paused-at-yield (not actively polling, so no busy leak), but the deadline `setTimeout` stays armed until it fires, and the explicit cleanup the code goes out of its way to do is bypassed. Fixing #1 (catch instead of throw) resolves this too.

### 🟡 3. Deadline can pre-empt an already-executed bid mid-handshake

`executeOrder` now races a fresh `executionStream.next(input)` against a single pre-created `deadlinePromise`:

```ts
const winner = await Promise.race([
  executionStream.next(input).then(r => ({ from: "exec", r })),
  deadlinePromise.then(r => ({ from: "deadline", r })),
])
if (winner.from === "deadline") { ...; return }
```

If the block deadline elapses while the consumer is busy in `bid.execute()` (between the `BIDS_RECEIVED` yield and feeding the result back), the next loop iteration races a freshly-started `executionStream.next(result)` against an **already-resolved** `deadlinePromise` — the resolved promise wins. The executor then yields `EXPIRED` and returns *without* processing `usedUserOps`/`BID_SELECTED`/fill, even though the consumer already submitted the UserOp to the bundler. The old `mergeRace` design had a similar race (submission was a raced step), but the consumer-driven window is now wider. Low severity / inherent to racing a deadline against a multi-step handshake, but worth a comment so it's a known edge rather than a surprise.

### 🟡 4. Test coverage: the rewritten control flow has no executable tests

Every touched test is `describe.skip(...)` integration tests (just renamed `execute`→`executeBest`). The new, subtle pieces — the bidirectional handshake, `driveExecution` feedback threading, the manual deadline race, and `BidImpl.simulate/execute` — have no unit coverage. Given this is primarily a control-flow rewrite, a couple of fake-generator unit tests (e.g. feeding back `undefined` continues polling; feeding back a `SelectBidResult` yields `BID_SELECTED`; deadline pre-empts) would meaningfully de-risk it.

### Things that look correct
- `buildBids` decode-and-drop semantics, the `NEW_BID` (now one `Bid` per bid) vs `BIDS_RECEIVED` (all fresh bids) split, and `auctionSeenBids` (dedups notifications only, not selection) are all consistent.
- `BidImpl`'s lazy, cached `signSelection()` correctly signs at most once across `simulate()`/`execute()`; domain separator and native-value (`address(0)` outputs + dispatch fee) computation are carried over unchanged.
- The `IntentOrderStatusUpdate` shape change for `NEW_BID`/`BIDS_RECEIVED` (now `Bid` objects) is an intentional breaking API change; types and the skipped tests are updated consistently.
- Cross-chain still idles until the deadline after `BID_SELECTED` (the docstring's "(terminates)" was already aspirational under `mergeRace`); not a regression.

### Recommendation
Address **#1** before merge — as written, the advertised drop-in `executeBest` is materially less resilient than the `execute` path it replaces, and the skipped tests won't catch it. #2 falls out of the same fix. #3 and #4 are worth a comment / follow-up. The core extraction is sound; the gap is specifically in the autopilot error-handling contract.

Wizdave97 (approved):

seunlanlege:
No sdk docs for this change

royvardhan:
> No sdk docs for this change

Sorry I was going to add but then forgot.

royvardhan:
> > No sdk docs for this change
> 
> Sorry I was going to add but then forgot.

@seunlanlege Added them [in](https://github.com/polytope-labs/hyperbridge/pull/958/changes/bb3003e6fd21be8e2395676c7387ce694ad71abc)

PR #950, opened 2026-06-04
[simplex]: shut down filler on delegation failure + raise set-code gas limit

royvardhan:
Addresses two issues raised by David on the filler's EIP-7702 delegation:

- **Fail fast.** The filler now shuts down when delegation fails on **all** solver-selection chains. If only some chains fail, it logs a warning and continues on the chains that succeeded.
- **Gas limit error.** Raised the flat set-code (0x04) gas limit from 350K → 650K, and fixed the Turnkey signer path which never applied the limit (it let viem estimate ~94k, which L2s reject).


seunlanlege:
does this shutdown when any chain delegation fails or when all fails?

royvardhan:
> does this shutdown when any chain delegation fails or when all fails?

When any fail, so that on a restart, it skips the delegated ones and directly try failed ones.

seunlanlege:
> > does this shutdown when any chain delegation fails or when all fails?
> 
> 
> 
> When any fail, so that on a restart, it skips the delegated ones and directly try failed ones. 

Should be when all fails

royvardhan:
> > > does this shutdown when any chain delegation fails or when all fails?
> > 
> > 
> > When any fail, so that on a restart, it skips the delegated ones and directly try failed ones.
> 
> Should be when all fails

Then if it continues to run, it will not be able to fill an order for the destination where it failed to delegate?

seunlanlege:
> Then if it continues to run, it will not be able to fill an order for the destination where it failed to delegate?

That's fine

Wizdave97 (approved):

Issue #939, opened 2026-06-01
[simplex] Aave Strategy


Wizdave97:
We want to have an aave funding strategy where filler can pull liquidity from an existing aave position.
The strategy should be configurable as one sided meaing Filler should be able to pull USDC, USDT from aave, but other token can continue to be pulled from a wallet or any other funding venue.

- [ ] Aave V3(priority for the first implementation)
- [ ] Aave V4(Can be implemented seperately)

seunlanlege:
Since we opted for the ERC-4626 vault interface, this is now closed

PR #958, opened 2026-06-09
[simplex]: ERC-4626 vault funding venue

royvardhan:
Implements #939 — yield-venue funding via ERC-4626.

Adds a filler-wide ERC-4626 venue for sourcing and managing stablecoin liquidity. Any compliant vault (Aave stataToken, Morpho, an issuer yield vault) is a config entry — no protocol-specific code:

- **Sourcing:** when wallet balance is short on a fill, withdraw the deficit from the solver's vault position (`vault.withdraw`), prepended atomically before `fillOrder` in the same ERC-7821 batch; capped by `vault.maxWithdraw`. Wired into both the stable and FX strategies.
- **Sweeping:** a timer deposits wallet balance above a configured `threshold` into the vault (exact `approve + deposit`, batched in one ERC-7821 call), keeping the threshold liquid for fills.
- **Shutdown:** all vault positions are redeemed back to the underlying asset.
- **Config:** top-level `[erc4626].vaults = [{ chain, vault, threshold? }]`; the underlying asset is resolved on-chain from the vault.
- Renamed the `basic` strategy to `stable`.


seunlanlege:
haven't looked at the code yet, but this should support hydrating free USDC/USDT balances at startup to aave & everytime its cumulative stablecoin free balance crosses some minimum threshhold.

docs/content/developers/evm/intent-gateway/simplex.mdx
```diff
@@ -222,20 +226,102 @@ No manual steps are required — delegation is idempotent and skipped if the EOA
...
+    { chain = "EVM-42161", vault = "0xa6D12574eFB239FC1D2099732bd8b5dC6306897F", threshold = "3000", minSweep = "10" },
+]
+```
+
+| Field | Required | Description |
+|---|---|---|
+| `chain` | yes | State-machine ID the vault lives on (e.g. `"EVM-8453"`) |
+| `vault` | yes | ERC-4626 vault address |
+| `threshold` | no | Wallet balance (human units) to keep liquid; excess is swept in. Omit to disable sweeping (withdraw-only). |
+| `minSweep` | no | Minimum excess (human units) worth sweeping; smaller excesses are skipped. Defaults to a small floor. |
```
seunlanlege: Tbh not sure i understand what minsweep does here

royvardhan: Let's say the Filler holds $5000 and the threshold is set to $2000. Then if there is no `minSweep`, it will send even a cent that crosses the threshold. minSweep can be $10-100, which says if the balance is threshold+minSweep, then only sweep the excess.

seunlanlege: instead of calling it minsweep, lets call it minBalance. Which is to say, sweep everything - minBalance

seunlanlege: also minBalance, would prevent fills from exceeding minBalance

seunlanlege:
also do we check maxDeposit/maxWithdrawal before interacting with the vault?

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -193,6 +236,8 @@ interface FillerTomlConfig {
 	chains: UserProvidedChainConfig[]
 	rebalancing?: RebalancingConfig
 	binance?: BinanceConfig
+	/** Filler-wide vault config: stablecoin sourcing for fills + threshold sweeping. */
+	erc4626?: VaultTomlConfig
```
seunlanlege: **Config key mismatch.** The code reads this section as `[erc4626]` (`config.erc4626`), but `filler-config-example.toml` and the simplex docs (`simplex.mdx`) both tell users to write `[vault]`. A user following the docs gets the section **silently ignored** — `config.erc4626` is `undefined`, the venue never initialises, and no error is raised.

Please make the key consistent across code, example, and docs. `[vault]` reads better than `[erc4626]`, but note it overlaps conceptually with the per-strategy `[strategies.vault.uniswapV4]` (the docs already need a callout to distinguish them), so pick deliberately.

sdk/packages/simplex/src/funding/vault/VaultFundingPlanner.ts
```diff
@@ -0,0 +1,344 @@
...
+				target: vault.vault,
+				value: 0n,
+				data: encodeFunctionData({
+					abi: ERC4626_ABI,
+					functionName: "withdraw",
+					args: [amount, solver, solver],
+				}) as HexString,
+			}
+
+			state.consume(vault.asset, amount)
```
seunlanlege, line 162: **Reservation may leak on un-filled orders.** `consume()` reserves capacity here, but `planWithdrawalForToken` runs inside `StableFiller.calculateProfitability` — i.e. during *bid evaluation*, for orders this filler may never win.

Per the reconcile logic in `VaultLiquidityState.refresh`, a reservation is only ever released when the on-chain position actually drops. A lost/abandoned bid never withdraws, so its reservation is never freed: `consumed` grows monotonically until `remaining = maxWithdrawable − consumed` floors at `0` and the venue stops sourcing entirely until restart.

Consider reserving only when a bid is actually selected/submitted, or giving reservations a TTL / per-round reset. (Worth confirming how often `calculateProfitability` runs vs. fills won — if it's ~1:1 in practice this is minor.)

sdk/packages/simplex/src/funding/vault/VaultLiquidityState.ts
```diff
@@ -0,0 +1,191 @@
...
+					args: [this.solver],
+				}) as Promise<bigint>,
+			])
+
+			// Reconcile consumed against realised position decrease.
+			const prevPosition = this.lastPositionAssets.get(key) ?? positionAssets
+			const decrease = prevPosition > positionAssets ? prevPosition - positionAssets : 0n
+			const prevConsumed = this.consumed.get(key) ?? 0n
+			const newConsumed = prevConsumed > decrease ? prevConsumed - decrease : 0n
+			this.consumed.set(key, newConsumed)
```
seunlanlege: This reconciliation only decrements `consumed` on a *realised* position decrease, which assumes every `consume()` eventually produces an on-chain withdrawal. In an auction most evaluated orders aren't won, so those reservations never reconcile — `consumed` leaks upward and `remaining` drifts to `0`, silently disabling sourcing over a long-running session. See the paired note at the `consume()` call site in `VaultFundingPlanner.planWithdrawalForToken`.

sdk/packages/simplex/src/funding/vault/VaultFundingPlanner.ts
```diff
@@ -0,0 +1,344 @@
...
+				if (vault.thresholdScaled === null) continue // sweeping disabled for this vault
+
+				const walletBalance = (await publicClient.readContract({
+					abi: ERC20_ABI,
+					address: vault.asset,
+					functionName: "balanceOf",
+					args: [solver],
+				})) as bigint
+
+				const excess = walletBalance > vault.thresholdScaled ? walletBalance - vault.thresholdScaled : 0n
```
seunlanlege: **Check `maxDeposit` before depositing (requested addition).** Size the deposit against `min(excess, maxDeposit(solver))` before building the `approve + deposit` pair. ERC-4626 requires `deposit` to revert when `assets > maxDeposit(receiver)`, and for the headline Aave `stataToken` venue `maxDeposit` drops to the remaining supply-cap headroom (and `0` when the cap is hit or the market is paused). Without the clamp the sweep tx reverts on every tick once the cap is reached — wasted gas, noisy failed-tx logs, and idle funds that silently never earn yield. Skip the vault when the clamped amount falls below `minSweepScaled`. This mirrors the `maxWithdraw` cap already used on the sourcing side.

_Minor, same path:_ `excess` is derived from a `balanceOf` read and deposited in a later tx — if a fill consumes wallet balance in between, `deposit(excess)` can exceed balance and revert. The batch reverts atomically (no stale allowance), but a small safety margin or a comment would help.

sdk/packages/simplex/src/config/abis/Erc4626.ts
```diff
@@ -0,0 +1,55 @@
...
+		outputs: [{ internalType: "address", name: "", type: "address" }],
+		stateMutability: "view",
+		type: "function",
+	},
+	{
+		inputs: [
+			{ internalType: "uint256", name: "assets", type: "uint256" },
+			{ internalType: "address", name: "receiver", type: "address" },
+		],
+		name: "deposit",
```
seunlanlege, line 14: For the `maxDeposit` clamp on the sweep path, `maxDeposit` needs adding to this ABI (alongside the existing `maxWithdraw`):

```ts
{
    inputs: [{ internalType: "address", name: "", type: "address" }],
    name: "maxDeposit",
    outputs: [{ internalType: "uint256", name: "", type: "uint256" }],
    stateMutability: "view",
    type: "function",
},
```

sdk/packages/simplex/src/funding/types.ts
```diff
@@ -117,10 +117,69 @@ export interface UniswapV4OutputFundingConfig {
...
+	 * Wallet balance of the underlying asset to keep liquid for direct fills, in
+	 * absolute human units (e.g. "3000"). Excess above it is swept into the
+	 * vault. Omit to disable sweeping for this vault (withdraw-only).
+	 */
+	threshold?: string
+	/**
+	 * Minimum excess (absolute human units) worth sweeping; smaller excesses are
+	 * skipped so gas never exceeds the value moved. Defaults to a small floor.
+	 */
+	minSweep?: string
```
seunlanlege: **`redeemOnShutdown` flag (requested addition).** Add an optional per-vault `redeemOnShutdown?: boolean` to `VaultConfig` so an operator can mark assets that should *not* be auto-redeemed back to the underlying on shutdown (e.g. a position you want to keep across restarts). Carry it through to `HydratedVault` (and the `hydrate()` mapping), then have `redeemChain` honour it.

Worth deciding the default explicitly: defaulting to `true` preserves today's behaviour (everything redeems); defaulting to `false` is safer (no surprise unwinds) but changes current semantics. Recommend documenting whichever you pick in `filler-config-example.toml`.

seunlanlege (commented):
Inline review of the ERC-4626 vault funding venue. Two pre-existing concerns plus the two requested additions (`maxDeposit` check, `redeemOnShutdown` flag) are left as inline comments at the relevant sections.

**Blocking-ish:** (1) the `[erc4626]` vs `[vault]` config-key mismatch silently disables the feature for anyone following the docs; (2) the reservation (`consumed`) accounting looks like it leaks for evaluated-but-not-won orders and can degrade sourcing to zero over time — worth confirming. Everything else is polish. Nice use of `maxWithdraw` as the authoritative sourcing cap, and good atomic `approve+deposit` batching.

sdk/packages/simplex/src/funding/types.ts
seunlanlege: Maybe we make the default `false`

royvardhan: The fillers will scratch their heads where the funds went after filler shutdown.

seunlanlege: So the cNGN yield vault only allows deposits every 24hrs. So if you redeem on shutdown, for instance just to change price, you miss your yields for that day.

royvardhan: Ohh then it makes sense.

sdk/packages/simplex/src/funding/vault/VaultLiquidityState.ts
```diff
@@ -0,0 +1,249 @@
...
+const DEFAULT_MIN_SWEEP = "10"
+
+/**
+ * Backstop expiry for a reservation whose bid never executes (lost auction,
+ * abandoned) — without it, `remaining` would drift to 0 and disable sourcing.
+ * Won bids release immediately via position-decrease reconciliation in
+ * {@link refresh}, so this only needs to exceed plan→execute latency (~15s
+ * auction + settlement); expiring too early risks an oversubscribed withdraw.
+ */
+const RESERVATION_TTL_MS = 30_000
```
seunlanlege: Reservation TTL is a clean fix for the leak — won bids release immediately via position-decrease reconciliation, and lost bids expire. One residual edge worth a thought (you already flag it in the comment): if plan→on-chain-execution exceeds 30s — a congested destination chain, slow bundler inclusion, or a cross-chain order whose dest-chain UserOp lands late — the reservation expires *before* the withdrawal executes. Two concurrent fills can then each plan against the full `maxWithdraw`, and whichever `withdraw` lands second reverts and rolls back its entire fill batch.

The `refresh()`-per-plan + per-call `maxWithdraw` cap keep this narrow, and over-conservatism is the safe direction, so 30s is a reasonable default. Might be worth deriving the TTL from `auctionTimeMs` (+ a settlement margin) or making it configurable, since cross-chain settlement can legitimately run longer than the fixed 30s. Non-blocking.

sdk/packages/simplex/src/funding/vault/VaultLiquidityState.ts
```diff
@@ -0,0 +1,249 @@
...
+		let toRelease = amount
+		while (toRelease > 0n && list.length > 0) {
+			const head = list[0]
+			if (head.amount <= toRelease) {
+				toRelease -= head.amount
+				list.shift()
+			} else {
+				head.amount -= toRelease
+				toRelease = 0n
+			}
```
seunlanlege, line 238: Minor: `releaseReserved` shifts strictly FIFO without pruning expired entries first, while `reservedFor` prunes them. In `refresh` you call `releaseReserved(key, decrease)` *before* `reservedFor(key)`, so a realised position drop can consume an already-expired head entry instead of the live reservation it actually corresponds to, leaving the live one counted for up to a TTL. It only ever *over*-reports reserved (under-reports `remaining`), so it's safe — just a touch more conservative than necessary. Pruning expired entries at the top of `releaseReserved` would tighten it. Non-blocking.

seunlanlege (commented):
Re-reviewed at `4f0421d` — thanks, all four items from the last pass are addressed:

- ✅ **Config key** — now `[vault]` consistently across code, `filler-config-example.toml`, and docs (planner `name` also renamed `"Vault"`).
- ✅ **Reservation leak** — replaced the monotonic `consumed` counter with TTL-bounded reservations; won bids release immediately on the realised position drop, lost bids expire. Good back-to-back-fill and TTL-expiry tests.
- ✅ **maxDeposit clamp** — added to the ABI and the sweep now deposits `min(excess, maxDeposit)`, skipping when headroom < `minSweep`. Covered by two new tests.
- ✅ **redeemOnShutdown** — per-vault flag, defaults `false` (positions kept across restarts), honoured in `redeemChain` and documented in the example config. Tested both ways.

Also a nice unprompted catch: ordering the vault venue **before** the V4 LP for the FX strategy so stablecoins source from the treasury before draining (and perturbing the price of) a concentrated-liquidity position.

Two small non-blocking notes inline (TTL vs. slow cross-chain settlement; a benign FIFO-vs-expiry ordering nit in `releaseReserved`). LGTM otherwise.

sdk/packages/simplex/src/funding/types.ts
```diff
@@ -117,10 +117,77 @@ export interface UniswapV4OutputFundingConfig {
...
```
seunlanlege: **Naming / config redesign suggestion** (non-blocking). The current `threshold` (retained float) + `minSweep` (dust floor) pair does two jobs with non-obvious names. Suggest replacing them with the standard high-water/low-water pattern:

- **`minBalance`** — the floor of the underlying to always keep in the wallet (gas/paymaster fees paid in USDC, working fill float). The sweep never drops below it.
- **`threshold`** — the high-water *trigger*: only when `walletBalance ≥ threshold` does a sweep fire, and it deposits everything down to `minBalance`.

```
on each sweep tick:
  if walletBalance < threshold:        skip            // hysteresis, no action below trigger
  sweepAmount = walletBalance - minBalance
  sweepAmount = min(sweepAmount, maxDeposit(solver))   // existing cap still applies
  deposit(sweepAmount)                                 // leaves exactly minBalance behind
```

Why it's simpler:
- **`minSweep` goes away entirely** — its only job was dust avoidance, and here the gap `threshold - minBalance` *is* the guaranteed minimum sweep size, set directly by the operator. No separate floor, no per-tick dust comparison.
- **Real hysteresis** — balance oscillates in `[minBalance, threshold]`: inflows/fills climb to `threshold`, a sweep drops to `minBalance`, repeat. No thrashing on tiny excesses.
- Both names say exactly what they do.

Mapping for the diff: today's `threshold` (retained float) becomes **`minBalance`**; the new `threshold` is a new high-water trigger; **`minSweep` is removed**.

Two specifics to fold in:
1. Enforce **`threshold > minBalance`** in `validateConfig` — otherwise a sweep triggers and immediately tries to deposit ≤ 0.
2. Make **`minBalance` required when `threshold` is set** rather than defaulting to 0 — since gas/paymaster is paid in this token, a silent 0 default would let a sweep empty the wallet of USDC. Omitting `threshold` keeps the existing withdraw-only (no sweeping) behaviour.

sdk/packages/simplex/src/funding/vault/VaultFundingPlanner.ts
```diff
@@ -0,0 +1,363 @@
...
+				if (excess < vault.minSweepScaled) continue
```
seunlanlege: Under the proposed `minBalance` + trigger model (see the `VaultConfig` comment), this becomes:

```ts
if (walletBalance < vault.thresholdScaled) continue   // below the high-water trigger
const sweepAmount = walletBalance - vault.minBalanceScaled
// ...then the existing maxDeposit clamp, depositing down to minBalance
```

The `excess < minSweepScaled` dust check on the next line is dropped — the `threshold > minBalance` gap is the implicit minimum sweep size.

seunlanlege (commented):
One follow-up design suggestion on the sweep config naming/semantics — non-blocking, details inline on `VaultConfig` and `sweepChain`. TL;DR: replace `threshold` (retained float) + `minSweep` (dust floor) with `minBalance` (floor to keep in wallet) + `threshold` (high-water trigger that sweeps down to `minBalance`). Drops `minSweep` entirely and gives clean hysteresis with two self-explanatory knobs.

sdk/packages/simplex/src/funding/vault/VaultLiquidityState.ts
royvardhan: Bumped the TTL from 30s to 60s and kept it a hardcoded constant. `auctionTimeMs` lives on the SDK's bid-consumer side (the user collecting bids), not the filler, so the venue can't read it to derive a TTL. We could add a configurable TTL, but that's another config knob, and I'd rather trim our configs than grow them for now. Won bids release immediately via the position-decrease reconciliation, so only never-executed bids ride the full minute. Happy to revisit if slow-settlement chains oversubscribe in practice.

sdk/packages/simplex/src/funding/vault/VaultLiquidityState.ts
royvardhan: Fixed. It prunes expired entries before the FIFO shift now, so a position drop only cancels a live reservation.

sdk/packages/simplex/src/funding/types.ts
royvardhan: Done. `minBalance` is the wallet floor and `threshold` is the high-water trigger that sweeps down to `minBalance`; `minSweep` is removed. `validateConfig` requires `minBalance` when `threshold` is set and enforces that `threshold` is greater than `minBalance`.

sdk/packages/simplex/src/funding/vault/VaultFundingPlanner.ts
royvardhan: Done. The sweep now skips when wallet balance is below `threshold`, then deposits `walletBalance - minBalance`, then applies the existing `maxDeposit` clamp (skipping when the clamped amount is zero or less). Dropped the per-tick dust check.

seunlanlege (approved):

seunlanlege:
lets try to get tests passing & we can merge

seunlanlege (commented):
## Final review — `2e773cfd`

Re-reviewed the PR end-to-end. Every item raised across the previous rounds is resolved, and I verified the assembled code (not just the deltas), the rename consistency, and config validation. **LGTM — ready to merge once CI (currently pending) is green.**

### Resolved
- ✅ **Config key** `[erc4626]` → `[vault]`, consistent across code / example / docs (repo-wide grep confirms no stale `erc4626`, `BasicFiller`, `minSweep`, or `type = "basic"` references).
- ✅ **Reservation leak** — TTL-bounded reservations (now 60s); won bids release immediately on the realised position drop, lost bids expire. `releaseReserved` now prunes expired entries before the FIFO shift, so a realised drop can't cancel a stale head and leave a live reservation counted.
- ✅ **maxDeposit clamp** on the sweep path; ABI updated; covered by tests (clamp + zero-headroom skip).
- ✅ **redeemOnShutdown** per-vault flag, defaults `false`, honoured in `redeemChain`, tested both ways.
- ✅ **Sweep config redesign** — `minSweep` removed; `minBalance` (wallet floor) + `threshold` (high-water trigger that deposits down to `minBalance`) give clean hysteresis. `validateConfig` enforces `minBalance` required when `threshold` is set and `threshold > minBalance`. Good coverage.
- ✅ Unprompted: FX strategy now sources the vault **before** the V4 LP, so stablecoins come from the treasury before draining (and skewing the price of) a concentrated-liquidity position.

### Verified sound
- `maxWithdraw` (not `previewRedeem`) as the authoritative sourcing cap → a `withdraw` can't revert and roll back the atomic ERC-7821 fill batch.
- Position-decrease reconciliation keys off a real `previewRedeem` drop; yield accrual only raises the position, so it won't trigger false reservation releases.
- Atomic `approve + deposit` / share-denominated `redeem`; per-chain shutdown failures are caught, not thrown.

### Residual non-blocking notes (fine to defer)
1. **TTL is a fixed 60s constant.** Reasonable, and the bump from 30s helps, but plan→execution on a congested/cross-chain destination can still exceed it, briefly allowing two fills to oversubscribe `maxWithdraw` (the second `withdraw` reverts and rolls back its fill). Deriving it from `auctionTimeMs` + a settlement margin would close that window; current behaviour is conservative-safe otherwise.
2. **Repeated `calculateProfitability` rechecks** on the same order each call `consume()`, stacking reservations within the TTL window → temporary over-reservation. Self-heals on win (release) or expiry; only worth touching if rechecks turn out to be frequent.
3. **Scope:** the `execute()`/`executeBest()`/`resume()` SDK docs overhaul is unrelated to vault funding and bloats the diff — purely cosmetic, just noting it for history/bisect.

Nice iteration on this — the funding-venue abstraction stays clean and the treasury logic is well tested.

seunlanlege:
Simplex test failing

PR #960, opened 2026-06-09
[sdk]: Add upgradeable variants of HyperFungibleToken and WrappedHyperFungibleToken

dharjeezy:
Adds upgradeable versions of the cross-chain fungible token contracts to the core SDK

### DIFF
--- a/sdk/packages/core/contracts/apps/HyperFungibleTokenUpgradeable.sol
+++ b/sdk/packages/core/contracts/apps/HyperFungibleTokenUpgradeable.sol
@@ -1,0 +1,363 @@
+// Copyright (C) Polytope Labs Ltd.
+// SPDX-License-Identifier: Apache-2.0
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+pragma solidity ^0.8.17;
+
+import {ERC20Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {PausableUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/PausableUpgradeable.sol";
+import {ERC165Upgradeable} from "@openzeppelin/contracts-upgradeable/utils/introspection/ERC165Upgradeable.sol";
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import {IHyperFungibleToken} from "../interfaces/IHyperFungibleToken.sol";
+import {PostRequest} from "../libraries/Message.sol";
+import {DispatchPost, IDispatcher} from "../interfaces/IDispatcher.sol";
+import {IncomingPostRequest, PostRequestTimeout} from "../interfaces/IApp.sol";
+import {ICallDispatcher} from "../interfaces/ICallDispatcher.sol";
+import {HyperApp} from "./HyperApp.sol";
+
+/**
+ * @title HyperFungibleTokenUpgradeable
+ * @author Polytope Labs (hello@polytope.technology)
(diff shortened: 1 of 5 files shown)


Wizdave97:
## Review: PR #960 — Upgradeable HyperFungible contracts

I compared the two new contracts against their non-upgradeable counterparts in the same directory:

- [[HyperFungibleTokenUpgradeable.sol](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/HyperFungibleTokenUpgradeable.sol)](sdk/packages/core/contracts/apps/HyperFungibleTokenUpgradeable.sol) vs [[HyperFungibleToken.sol](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/HyperFungibleToken.sol)](sdk/packages/core/contracts/apps/HyperFungibleToken.sol)
- [[WrappedHyperFungibleTokenUpgradeable.sol](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/WrappedHyperFungibleToken.sol)](sdk/packages/core/contracts/apps/[WrappedHyperFungibleToken.sol](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/WrappedHyperFungibleToken.sol)) vs [WrappedHyperFungibleToken.sol](sdk/packages/core/contracts/apps/WrappedHyperFungibleToken.sol)

The upgradeability conversion itself is done correctly. But the implementations are **not fully at parity** — there is one real behavioral regression.

### 🔴 Medium — `pause()` no longer freezes token transfers in `HyperFungibleTokenUpgradeable`

The non-upgradeable `HyperFungibleToken` overrides ERC20 movement to respect the pause switch:

```solidity
// HyperFungibleToken.sol:336-343
function transfer(address to, uint256 value) public override whenNotPaused returns (bool) {...}
function transferFrom(address from, address to, uint256 value) public override whenNotPaused returns (bool) {...}
```

`HyperFungibleTokenUpgradeable` **omits both overrides** (and does not override `_update` either). The result: when the owner calls `pause()` on the upgradeable token, only the cross-chain entrypoints (`send`, `onAccept`, `onPostRequestTimeout`) are blocked — **ordinary holder `transfer`/`transferFrom` still succeed.**

This is a meaningful weakening of the emergency lever. `HyperFungibleToken` is a mint/burn bridged asset; the non-upgradeable design intends `pause()` to freeze *all* token movement during an incident (e.g. an unauthorized-mint exploit), so fraudulently minted or compromised balances can't be moved or sold while paused. The upgradeable variant loses that — an attacker holding tokens can still freely transfer them while the contract is "paused." Given everything else is a faithful line-for-line port, this looks like an oversight rather than a deliberate design change.

**Fix:** re-add the two overrides (using `ERC20Upgradeable`), or implement the gate once via `_update`:

```solidity
function _update(address from, address to, uint256 value) internal override whenNotPaused {
    super._update(from, to, value);
}
```
(`_update` is the cleaner choice — it also covers `_mint`/`_burn`, though note that mint/burn already happen inside `whenNotPaused` paths, so gating `_update` would additionally block the refund mint on `onPostRequestTimeout`; if you want refunds to keep working while paused, stick with the `transfer`/`transferFrom` overrides to match the original exactly.)

### ✅ Wrapped variant — at parity
`WrappedHyperFungibleTokenUpgradeable` is a faithful port. It isn't an ERC20 itself (it locks/unlocks an underlying), so there are no transfer overrides to carry over. `send`/`onAccept`/`onPostRequestTimeout`, the WETH push-then-rewrap fallback, and `receive()` are all identical to the original.

### ✅ Upgradeability mechanics — correct
- `constructor() { _disableInitializers(); }` with `@custom:oz-upgrades-unsafe-allow constructor` — implementation can't be initialized directly. Correct.
- `initialize` (`initializer`) → `__X_init` (`onlyInitializing`) calls every parent init (`__ERC20_init`, `__ERC165_init`, `__Ownable_init`, `__Pausable_init`). No missed initializer.
- `HyperApp` base is **stateless** (no storage variables), so inheriting it under a proxy is safe.
- OZ `contracts-upgradeable@5.6.1` uses ERC-7201 namespaced storage for its bases, so the contract's own `_host` / `_dispatcher` / `_supportedChains` (and `_underlying` / `_isWeth`) slots can't collide with OZ base storage.
- `configure()`'s host-locking (`if (_host == address(0))`) matches the original.

### Minor notes (not blockers)
- **No storage gap.** These are the repo's first first-party upgradeable contracts and reserve no `__gap`. Appending new state in a future version is safe (custom storage is sequential, OZ bases are namespaced), but you cannot insert variables, and a future base contract with non-namespaced storage would be a hazard. Consider documenting the convention or adding a `uint256[N] __gap`.
- **Proxy init front-running.** `initialize` is publicly callable; deploy + initialize must be atomic (factory or proxy constructor) so an attacker can't initialize the proxy first and seize ownership. Standard upgradeable caveat — worth ensuring the deployment script does this.
- **No tests.** No test file references `HyperFungibleTokenUpgradeable`/`WrappedHyperFungibleTokenUpgradeable`. Given they're meant to be behind a proxy, an initialize-once + pause-behavior test would have caught the regression above.

### Bottom line
The Wrapped contract is safe and at parity. `HyperFungibleTokenUpgradeable` is **not** at parity: it silently drops pause enforcement on holder transfers. I'd treat that as a blocker — add the `whenNotPaused` gate before merging — then the pair is safe modulo the standard proxy-deployment and storage-gap hygiene notes.

Wizdave97 (approved):

PR #962, opened 2026-06-10
 [tesseract]: periodic outbound consensus claim task

dharjeezy:
The outbound consensus claim task has been refactored to work like the fee withdrawal mechanism. Instead of forwarding claims through a channel on each delivery, the delivery path now writes a row to the database and a separate periodic task wakes every 600 seconds, checks which claims have already been redeemed on Hyperbridge, and submits the rest in parallel. A fast-fail height check was also added so claims on chains where Hyperbridge has not yet committed to the delivery block are skipped immediately rather than burning the full 300 second wait on every cycle.

### DIFF
--- a/modules/pallets/relayer/src/outbound_consensus.rs
+++ b/modules/pallets/relayer/src/outbound_consensus.rs
@@ -161,8 +161,8 @@ where
 		// Ethereum trie stores. Returns `None` for an unset / zero-address
 		// slot, which we surface as `OutboundDeliveryNotProven` (logically
 		// equivalent to "no delivery proven yet").
-		let evm_address =
-			Self::decode_epochs_slot_address(&raw).ok_or(Error::<T>::OutboundDeliveryNotProven)?;
+		let evm_address = Self::decode_epochs_slot_address(destination, &raw)
+			.ok_or(Error::<T>::OutboundDeliveryNotProven)?;
 
 		// Replay protection comes from the `OutboundConsensusRotationsClaimed`
 		let msg = outbound_consensus_delivery_message(set_id, destination, payee);
@@ -197,20 +197,18 @@ where
 		Ok(())
 	}
 
-	/// Decode the EVM `address` value stored at
-	/// `EvmHost._epochs[set_id]`, as returned by
-	/// `EvmStateMachine::verify_state_proof`.
-	///
-	/// Standard EVM MPT trie returns RLP-encoded values; Pharos flat trie
-	/// returns the raw 32-byte zero-padded ABI value. We try RLP first and
-	/// fall back to extracting the last 20 bytes of a 32-byte padded value.
-	pub fn decode_epochs_slot_address(raw: &[u8]) -> Option<Address> {
+	/// Decode the `address` value from `EvmHost._epochs[set_id]` as returned
+	/// by `EvmStateMachine::verify_state_proof`. Standard EVM chains RLP-encode
+	/// the value; Pharos stores it as a raw 32-byte ABI-padded word.
+	pub fn decode_epochs_slot_address(
+		state_id: ismp::host::StateMachine,
+		raw: &[u8],
+	) -> Option<Address> {
 		use alloy_rlp::Decodable;
 		if let Ok(addr) = Address::decode(&mut &*raw) {
 			return if addr == Address::ZERO { None } else { Some(addr) };
 		}
-		// Pharos flat trie: raw 32-byte zero-padded ABI encoding
-		if raw.len() == 32 {
+		if state_id.is_pharos() && raw.len() == 32 {
 			let addr = Address::from_slice(&raw[12..]);
 			return if addr == Address::ZERO { None } else { Some(addr) };
 		}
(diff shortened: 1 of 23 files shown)


Wizdave97:
## Re-review of [[#962](https://github.com/polytope-labs/hyperbridge/pull/962)](https://github.com/polytope-labs/hyperbridge/pull/962) at new head `758736f0`


### What the new commits propagate or introduce

1. **The 300s-timeout / challenge-period hazard now exists in two places.** The request claim task's `process_claim` ([[outbound_request_claim.rs:246](https://claude.ai/epitaxy/tesseract/messaging/messaging/src/outbound_request_claim.rs:246)](tesseract/messaging/messaging/src/outbound_request_claim.rs:246)) copied the same `timeout(300s, wait_for_state_machine_update(...))` construction. As before, `wait_for_state_machine_update` unconditionally sleeps the **full challenge period** via `observe_challenge_period` even when the height is long-committed — so any destination whose challenge period on Hyperbridge is ≥ 300s has *both* reward types permanently unclaimable through the automatic path. The request task also serves **substrate destinations**, whose challenge periods may differ from the EVM chains'. This remains the most important thing to resolve before merge.

3. **New gap: the manual `claim-rewards` command doesn't cover request claims.** [[claim_rewards.rs](https://claude.ai/epitaxy/tesseract/relayer/src/claim_rewards.rs)](tesseract/relayer/src/claim_rewards.rs) only reads `list_pending_rotation_claims`. Now that request claims are also periodic-and-fast-fail, a Polkadot Hub request delivery whose height Hyperbridge never commits has **no recovery path at all** — the periodic task fast-fails every tick and the manual command ignores those rows. Either the subcommand should also drain `list_pending_request_claims` (it already advances the Hub's height, so the marginal work is small), or this asymmetry should be documented.

4. **Magic numbers now duplicated across both tasks.** `Duration::from_secs(600)` and `Duration::from_secs(300)` appear verbatim in both [[outbound_claim.rs:53](https://claude.ai/epitaxy/tesseract/messaging/messaging/src/outbound_claim.rs:53)](tesseract/messaging/messaging/src/outbound_claim.rs:53) and [[outbound_request_claim.rs:56](https://claude.ai/epitaxy/tesseract/messaging/messaging/src/outbound_request_claim.rs:56)](tesseract/messaging/messaging/src/outbound_request_claim.rs:56). Shared named consts (or `RelayerConfig` fields) would prevent the two tasks drifting apart.

5. Let's check the statemachine id explicitly before doing any fallback flat trie decoding for Pharos


- The **periodic task can't claim Polkadot Hub rewards on its own** — `advance_counterparty_to` is still only wired into the manual subcommand, and stuck Hub rows still log at `error!` level every 600s.
- `IsmpHost::advance_counterparty_to`'s **silent `Ok(())` default**.
- `is_polkadot_hub` **re-hardcodes chain IDs** that exist as constants in [[consensus.rs:49](https://claude.ai/epitaxy/modules/ismp/clients/parachain/client/src/consensus.rs:49)](modules/ismp/clients/parachain/client/src/consensus.rs:49).

modules/pallets/relayer/src/outbound_consensus.rs
```diff
@@ -200,14 +200,21 @@ where
...
-		let addr = Address::decode(&mut &*raw).ok()?;
-		if addr == Address::ZERO {
-			None
-		} else {
-			Some(addr)
+		if let Ok(addr) = Address::decode(&mut &*raw) {
+			return if addr == Address::ZERO { None } else { Some(addr) };
+		}
+		// Pharos flat trie: raw 32-byte zero-padded ABI encoding
+		if raw.len() == 32 {
```
Wizdave97: check state machine before this fallback

tesseract/messaging/evm/src/lib.rs
```diff
@@ -348,6 +369,10 @@ impl EvmClient {
 
 		let http_client = alloy::transports::http::reqwest::Client::builder()
 			.timeout(Duration::from_secs(180))
+			// Some cloud-fronted endpoints (e.g. Polkadot Hub's Cloudflare RPC)
+			// reject the ALPN h2 handshake. HTTP/1.1 works everywhere and is
+			// fine for JSON-RPC (single request/response, no server push needed).
+			.http1_only()
```
Wizdave97: Don't think we need this, polkadot hub endpoint works fine on our relayer

tesseract/messaging/messaging/src/outbound_claim.rs
```diff
@@ -49,90 +38,73 @@ use sp_core::Pair;
...
+							destination = %row.destination,
+							set_id = row.set_id,
+							"failed to delete already-claimed row",
 						);
 					}
 				}
 			}
 		}
 
+		advance_hub_heights(&unclaimed, &destinations, &consensus_hosts, &hb_provider).await;
```
Wizdave97: This should be done in `process_claim`, There should be an explicit check if the claim is for polkadot hub then explicitly advance it's consensus.

tesseract/messaging/messaging/src/outbound_request_claim.rs
```diff
@@ -108,77 +84,83 @@ pub async fn run(
...
-							"failed to delete already-claimed row; will retry next trigger",
+							commitment = %row.commitment(),
+							"failed to delete already-claimed row",
 						);
 					}
 				}
 			}
 		}
 
+		crate::outbound_claim::advance_hub_heights(
```
Wizdave97: Same

Wizdave97 (approved):

PR #966, opened 2026-06-12
[sdk}: intent quote service

### LINKED ISSUE #963: Create Methods in the SDK for users to quote amount
This method should also consider protocol fees logic

ddboy19912:
# feat(sdk): add partner-facing intent quote API (`quoteIntent`)

## Summary

Adds a strategy-based intent quoting API to the SDK so partners can price IntentGateway V2 orders before placing them. Uniswap V4 is the only strategy today, but the API is strategy-shaped (`strategy: "uniswap_v4"` is the default) so new quote sources can be added without breaking partner integrations.

```typescript
import { quoteIntent } from "@hyperbridge/sdk"

const quote = await quoteIntent({
  source: { chainId: "EVM-8453" },
  destination: { chainId: "EVM-8453" },
  tokenIn: { address: USDC, decimals: 6, symbol: "USDC" },
  tokenOut: { address: cNGN, decimals: 6, symbol: "cNGN" },
  amountIn: 100_000_000n,
})
// quote.amountIn / quote.amountOut → raw Uniswap V4 quoter amounts
```

## What's included

- **`IntentQuoteService` / `quoteIntent()`** (`protocols/intents/quote/`) — supports exact-input and exact-output quotes. Returns the raw amounts from the Uniswap V4 quoter with no slippage or fee adjustments; partners apply their own tolerance. `quoteMetadata` exposes the resolved `poolKey`, `quoterAddress`, and the source-chain `protocolFeeBps` (read from `IntentGatewayV2.params()`) so callers can account for the gateway's input fee when constructing the order.
- **Pool resolution** — pools are resolved from SDK chain config (`uniswapV4Pools`, seeded with USDC/cNGN on Base). Same-chain quotes match configured pools by token address; cross-chain quotes match `tokenIn` by symbol since its source-chain address is meaningless against destination config. For unconfigured pairs, callers can pass an explicit `uniswapV4.poolKey` override (with optional `quoterAddress` and, for cross-chain quotes, `currencyIn` — the destination-side input currency, validated against the pool to prevent silently quoting the wrong swap direction).
- **Config plumbing** — `uniswapV4Pools` chain config field, `ChainConfigService.getUniswapV4PoolConfigs()` and `getAssetAddress()`.
- **Errors** — `UnsupportedIntentQuoteStrategyError`, `UnsupportedIntentQuotePairError`, and a clear failure when the V4 quoter is not configured for a chain instead of an opaque `eth_call` failure.
- **Docs** — new `quoteIntent(params)` section in `intent-gateway.mdx`, including how to apply slippage and the protocol fee on top of the raw amounts.
- **Test** — live Base USDC → cNGN exact-input quote in `tests/sequential/intentGateway.test.ts`.

## Notes for reviewers

- Quoted amounts are intentionally unadjusted: applying slippage tolerance and accounting for `protocolFeeBps` is the caller's responsibility (documented with a snippet in the docs).
- The V4 quoter is called via `eth_call` with manual encode/decode since its quote functions are `nonpayable`.

## Test plan

- [ ] `pnpm test:intent-gateway` (live Base RPC) — quotes 1 USDC → cNGN through the configured V4 pool
- [ ] `tsc --noEmit` passes (two pre-existing errors in unrelated test files)

Closes #963

### DIFF
--- a/sdk/packages/sdk/src/protocols/intents/quote/uniswapV4.ts
+++ b/sdk/packages/sdk/src/protocols/intents/quote/uniswapV4.ts
@@ -18,7 +18,6 @@ import {
 type GatewayParamsObject = { protocolFeeBps?: bigint | number | string }
 type GatewayParams = GatewayParamsObject | readonly unknown[]
 export const UNISWAP_INTENT_QUOTE_CHAIN = Chains.BASE_MAINNET
-export const UNISWAP_INTENT_QUOTE_SLIPPAGE_BPS = 30n
 const BPS_DENOMINATOR = 10_000n
 
 interface ResolvedPoolConfig {
@@ -192,10 +191,10 @@ export class UniswapV4IntentQuoteStrategy implements IntentQuoteStrategyHandler 
 		poolConfig: ResolvedPoolConfig
 	}): Promise<QuoteIntentResult> {
 		const amountIn = args.params.amountIn!
-		const amountOut = applyUniswapIntentQuoteSlippage(
-			await this.readV4QuoteExactInput(args.client, args.poolConfig, amountIn),
-			"EXACT_INPUT",
-		)
+		// The gateway deducts its protocol fee from order inputs, so only the
+		// reduced amount reaches the swap. Quote against that net amount.
+		const swapAmountIn = deductProtocolFee(amountIn, args.protocolFeeBps)
+		const amountOut = await this.readV4QuoteExactInput(args.client, args.poolConfig, swapAmountIn)
 
 		return {
 			strategy: "uniswap_v4",
@@ -206,7 +205,6 @@ export class UniswapV4IntentQuoteStrategy implements IntentQuoteStrategyHandler 
 				quoteChain: UNISWAP_INTENT_QUOTE_CHAIN,
 				poolKey: args.poolConfig.poolKey,
 				quoterAddress: args.poolConfig.quoterAddress,
-				slippageBps: UNISWAP_INTENT_QUOTE_SLIPPAGE_BPS,
 				protocolFeeBps: args.protocolFeeBps,
 			},
 		}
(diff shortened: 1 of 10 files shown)


seunlanlege:
better to be a method on the existing intent class

Wizdave97 (approved):

seunlanlege (commented):
(Retracted — this review was posted in error from a stale file and did not apply to this PR. Its inline comments have been removed. See the follow-up review for the actual feedback.)

sdk/packages/sdk/src/protocols/intents/quote/uniswapV4.ts
```diff
@@ -0,0 +1,357 @@
...
+	type QuoteIntentParams,
+	type QuoteIntentResult,
+	UnsupportedIntentQuotePairError,
+	type UniswapV4PoolKey,
+} from "./types"
+
+type GatewayParamsObject = { protocolFeeBps?: bigint | number | string }
+type GatewayParams = GatewayParamsObject | readonly unknown[]
+export const UNISWAP_INTENT_QUOTE_CHAIN = Chains.BASE_MAINNET
+export const UNISWAP_INTENT_QUOTE_SLIPPAGE_BPS = 30n
```
seunlanlege: Let's not hardcode a 30 bps slippage buffer here. The amount returned from the Uniswap V4 quoter should be passed through as-is, and the only adjustment baked into the quote should be the one that's actually charged on-chain: `protocolFeeBps` (read from `IntentGatewayV2.params()`).

A fixed 30 bps is arbitrary: it's unrelated to the gateway's real economics, it silently diverges from the actual fee whenever `protocolFeeBps` changes, and it conflicts with the goal of returning unadjusted amounts so partners can apply their own tolerance. Deriving the buffer from `protocolFeeBps` keeps the quote tied to the real cost.

sdk/packages/sdk/src/protocols/intents/quote/uniswapV4.ts
```diff
@@ -0,0 +1,357 @@
...
+
+function getZeroForOne(tokenIn: HexString, poolKey: UniswapV4PoolKey): boolean {
+	return getAddress(tokenIn).toLowerCase() === getAddress(poolKey.currency0).toLowerCase()
+}
+
+function isGatewayParamsTuple(value: GatewayParams): value is readonly unknown[] {
+	return Array.isArray(value)
+}
+
+export function applyUniswapIntentQuoteSlippage(
```
seunlanlege: Suggest dropping `applyUniswapIntentQuoteSlippage` (and the call sites at L195 / L222). Return the raw quoter amount, and use `protocolFeeBps` for the buffer instead of the hardcoded 30 bps:

- EXACT_INPUT: reduce `amountOut` by `protocolFeeBps` rather than the fixed slippage.
- EXACT_OUTPUT: gross up `amountIn` by `protocolFeeBps`.

That way the adjustment reflects what the gateway actually deducts, and `quoteMetadata.protocolFeeBps` becomes the single source of truth instead of two separate numbers (`slippageBps` + `protocolFeeBps`) the caller has to reconcile.

seunlanlege (approved):

seunlanlege:
need to patch bump sdk

ddboy19912:
> need to patch bump sdk

It's already patched to `2.2.1`

PR #972, opened 2026-06-15
[indexer]: add chain field to DustCollected/DustSwept entities

### LINKED ISSUE #951: [indexer]: DustCollected/DustSwept entities miss chain field
## Bug

The indexer records `DustCollected` and `DustSwept` for IntentGatewayV3 events, but the entities don't carry a first-class `chain` field. The host chain is only baked into the entity `id` string (`\`${chainId}-${tokenAddress}\`` — see `ProtocolRevenueService.recordDustCollected` / `recordDustSwept` in `sdk/packages/indexer/src/services/protocol-revenue.service.ts`) and the handler never threads the resolved `chain` through to the service.

Effects:

- Cannot filter / group queries by chain — consumers have to substring-match on `id`.
- The schema docstring says `\"Unique identifier: chainId-tokenAddress\"`, but the actual schema (`sdk/packages/indexer/src/configs/schema.graphql`) has no `chain` column, no index on it, and no `@derivedFrom` link to a chain entity.
- The handlers (`dustCollected.event.handler.ts`, `dustSwept.event.handler.ts`) already resolve `chain = getHostStateMachine(chainId)` for the timestamp lookup but discard it before persisting.

## Fix

1. Add a `chain: String! @index` (state-machine ID, e.g. `EVM-1`) to both `DustCollected` and `DustSwept` in `sdk/packages/indexer/src/configs/schema.graphql`.
2. Plumb the resolved `chain` from the event handlers (`sdk/packages/indexer/src/handlers/events/intentGatewayV3/dustCollected.event.handler.ts`, `dustSwept.event.handler.ts`) into `ProtocolRevenueService.recordDustCollected` / `recordDustSwept`.
3. Populate `chain` on entity create, and use it (instead of the bare `chainId` global) to compose the `id`. Keep the id format stable (`${chain}-${tokenAddress}`) to keep the per-chain-per-token aggregation behaviour.

## Affected files

- `sdk/packages/indexer/src/configs/schema.graphql` — add `chain` field to `DustCollected` and `DustSwept`.
- `sdk/packages/indexer/src/services/protocol-revenue.service.ts` — accept and persist `chain`; build `id` from the passed-in chain.
- `sdk/packages/indexer/src/handlers/events/intentGatewayV3/dustCollected.event.handler.ts` — pass `chain` to the service.
- `sdk/packages/indexer/src/handlers/events/intentGatewayV3/dustSwept.event.handler.ts` — same.

### LINKED ISSUE #952: [indexer]: add cumulative dust collected/swept per chain
## Background

`DustCollected` and `DustSwept` (`sdk/packages/indexer/src/configs/schema.graphql`) currently aggregate per-chain *and* per-token (id = \`${chainId}-${tokenAddress}\` — see \`ProtocolRevenueService\` in \`sdk/packages/indexer/src/services/protocol-revenue.service.ts\`). There is no chain-wide rollup. Consumers that want \"how much dust has the protocol collected on Base / Arbitrum / Mainnet?\" have to fan-out per token and sum client-side.

## Proposed change

Add two new aggregate entities tracking the running total per chain, updated whenever the per-token entities are updated:

```graphql
type CumulativeDustCollectedPerChain @entity {
  \"\"\"State machine id, e.g. EVM-1\"\"\"
  id: ID!
  chain: String! @index
  \"\"\"Sum across all tokens, in USD scaled by 1e18 (mirrors CumulativeVolumeUSD)\"\"\"
  cumulativeUsd: BigInt!
  lastUpdated: Date! @index
}

type CumulativeDustSweptPerChain @entity {
  id: ID!
  chain: String! @index
  cumulativeUsd: BigInt!
  lastUpdated: Date! @index
}
```

Mirror the convention already used by \`CumulativeVolumeUSD\` (`schema.graphql:1444`): 1e18 fixed-point USD.

Wire it into \`ProtocolRevenueService.recordDustCollected\` / \`recordDustSwept\` so that each event:

1. Updates the per-chain-per-token \`DustCollected\` / \`DustSwept\` (existing).
2. Looks up token price via \`TokenPriceService.getPrice(symbol)\`, computes the USD delta for the *new* amount only, and adds it to the chain-wide cumulative entity (create on first sight).

Tokens with no price (`getPrice` returns 0) should be skipped from the USD rollup but still recorded in the per-token entity, with a debug log so missing prices are visible.

## Depends on

\#951 — the per-event \`chain\` field needs to be threaded through \`ProtocolRevenueService\` first; this issue then uses it as the rollup key.

## Affected files

- \`sdk/packages/indexer/src/configs/schema.graphql\` — two new aggregate entities.
- \`sdk/packages/indexer/src/services/protocol-revenue.service.ts\` — update aggregates after the per-token write, USD conversion via \`TokenPriceService\`.

royvardhan:
Closes #951
Closes #952

Adds a `chain` field to the dust entities and a per-chain USD rollup.

- Add `chain: String! @index` to `ProtocolDustCollected` / `ProtocolDustSwept` (renamed from `DustCollected` / `DustSwept`), threading the resolved host state machine from the event handlers and composing the ID as `${chain}-${tokenAddress}`
- Add `CumulativeDustCollectedPerChain` / `CumulativeDustSweptPerChain` entities tracking a running USD total per chain (1e18 fixed-point, mirroring `CumulativeVolumeUSD`)
- Roll the USD value of each new dust amount into the cumulative entity via `TokenPriceService`; tokens without price data are skipped from the rollup and logged

### DIFF
--- a/sdk/packages/indexer/src/services/protocol-revenue.service.ts
+++ b/sdk/packages/indexer/src/services/protocol-revenue.service.ts
@@ -11,8 +11,13 @@ export class ProtocolRevenueService {
 	/**
 	 * Get or create a DustCollected record
 	 */
-	static async recordDustCollected(tokenAddress: string, amount: bigint, timestamp: bigint): Promise<DustCollected> {
-		const id = `${chainId}-${tokenAddress.toLowerCase()}`
+	static async recordDustCollected(
+		chain: string,
+		tokenAddress: string,
+		amount: bigint,
+		timestamp: bigint,
+	): Promise<DustCollected> {
+		const id = `${chain}-${tokenAddress.toLowerCase()}`
 		let symbol = "eth"
 
 		// Get token symbol if not native token
@@ -35,6 +40,7 @@ export class ProtocolRevenueService {
 		if (!dustCollected) {
 			dustCollected = await DustCollected.create({
 				id,
+				chain,
 				tokenSymbol: symbol,
 				amount,
 				lastUpdated: timestampToDate(timestamp),
@@ -60,8 +66,13 @@ export class ProtocolRevenueService {
 	/**
 	 * Get or create a DustSwept record
 	 */
-	static async recordDustSwept(tokenAddress: string, amount: bigint, timestamp: bigint): Promise<DustSwept> {
-		const id = `${chainId}-${tokenAddress.toLowerCase()}`
+	static async recordDustSwept(
+		chain: string,
+		tokenAddress: string,
+		amount: bigint,
+		timestamp: bigint,
+	): Promise<DustSwept> {
+		const id = `${chain}-${tokenAddress.toLowerCase()}`
 		let symbol = "eth"
 
 		// Get token symbol if not native token
@@ -84,6 +95,7 @@ export class ProtocolRevenueService {
 		if (!dustSwept) {
 			dustSwept = await DustSwept.create({
 				id,
+				chain,
 				tokenSymbol: symbol,
 				amount,
 				lastUpdated: timestampToDate(timestamp),
--- a/sdk/packages/indexer/src/services/protocol-revenue.service.ts
+++ b/sdk/packages/indexer/src/services/protocol-revenue.service.ts
@@ -2,10 +2,16 @@ import Decimal from "decimal.js"
 import { ERC6160Ext20Abi__factory } from "@/configs/src/types/contracts"
 import { ProtocolDustCollected } from "@/configs/src/types/models/ProtocolDustCollected"
 import { ProtocolDustSwept } from "@/configs/src/types/models/ProtocolDustSwept"
+import { CumulativeDustCollectedPerChain } from "@/configs/src/types/models/CumulativeDustCollectedPerChain"
+import { CumulativeDustSweptPerChain } from "@/configs/src/types/models/CumulativeDustSweptPerChain"
 import { timestampToDate } from "@/utils/date.helpers"
 import PriceHelper from "@/utils/price.helpers"
 import { TokenPriceService } from "./token-price.service"
 import stringify from "safe-stable-stringify"
+
+/** Decimal places used for the BigInt fixed-point representation of USD amounts. */
+const USD_SCALE = 18
+const USD_SCALE_FACTOR = new Decimal(10).pow(USD_SCALE)
 
 export class ProtocolRevenueService {
 	/**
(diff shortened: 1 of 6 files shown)


royvardhan:
@Wizdave97 Do I need to change the schema names?

Wizdave97:
> @Wizdave97 Do I need to change the schema names?

Yes

sdk/packages/indexer/src/services/protocol-revenue.service.ts
```diff
@@ -105,4 +158,27 @@ export class ProtocolRevenueService {
...
+	 * Returns null when no price data is available (testnet, non-whitelisted, or
+	 * unavailable), so the caller can skip the USD rollup for that token.
+	 */
+	private static async computeDustUsdDelta(
+		chain: string,
+		symbol: string,
+		amount: bigint,
+		decimals: number,
+	): Promise<bigint | null> {
+		const price = await TokenPriceService.getPrice(symbol)
```
Wizdave97: Let's get the token price from the onchain dex, our price service no longer works.

sdk/packages/indexer/src/services/protocol-revenue.service.ts
```diff
@@ -95,6 +131,23 @@ export class ProtocolRevenueService {
 
 		await dustSwept.save()
 
+		const usdDelta = await this.computeDustUsdDelta(chain, symbol, amount, decimals)
```
Wizdave97: We should still record cumulative for the raw amount if we can't get it's usd value, usd amount is just a convenience.

sdk/packages/indexer/src/services/protocol-revenue.service.ts
royvardhan: Done

sdk/packages/indexer/src/services/protocol-revenue.service.ts
royvardhan: Raw is already saved on the per-token `ProtocolDust*` entities before the USD step, so it's never lost. Kept the per-chain cumulative USD-only since summing raw across tokens/decimals just drifts. Can change if you'd rather.

Wizdave97: Is ProtocolDust cumulative?

royvardhan: Yes

Wizdave97 (approved):

PR #978, opened 2026-06-19
[simplex]: one-sided LP for the FX strategy

### LINKED ISSUE #974: [simplex]: One sided LP
Simplex should support one-sided Liquidity Provisioning.

royvardhan:
Closes #974.

Adds one-sided liquidity provisioning to the HyperFX strategy, so the filler can fill in a single direction and keep accumulating one asset while only giving away the other. The mechanism follows the pricing mode, with no new config in the static-curve case:

- Static curves: omit one curve to restrict the direction.
  - Ask curve only (omit `bidPriceCurve`): only sells exotic — accumulates the stablecoin (e.g. take USDC, hand out cNGN).
  - Bid curve only (omit `askPriceCurve`): only buys exotic — accumulates the exotic token.
- Uniswap V4 pool pricing (no curves): set `side` on `[strategies.vault.uniswapV4]` — `"ask"` to only sell exotic, `"bid"` to only buy exotic.
- Both curves set, or pool pricing with no `side`: fills both directions (unchanged default).

```toml
[strategies.vault.uniswapV4]
side = "ask"  # only sell exotic; keep accumulating stablecoins
positions = [
    { chain = "EVM-8453", tokenId = "2087350" },
]
```

Enforcement runs in both `canFill` and `calculateProfitability`, so off-side orders are rejected consistently. The direction gate is deliberately kept out of `classifyAllPairs` — that result is cached and shared across strategies, whereas enablement is per-strategy, so the gate must run on every evaluation regardless of cache state. Mixed-direction orders are rejected as a whole, since the IntentGateway settles all legs atomically. `side` is rejected when combined with static curves (which already express one-sided by omission).

Validation relaxed from "set both curves or neither" to "a bid and/or ask curve, or Uniswap V4 positions". Config example and docs updated; unit tests cover curve-based (both/ask-only/bid-only), pool `side` (ask/bid/none), the side-plus-curve rejection, and a regression test that the gate still rejects when another strategy has cached the classification under the shared cache.

### DIFF
--- a/docs/content/developers/evm/intent-gateway/simplex.mdx
+++ b/docs/content/developers/evm/intent-gateway/simplex.mdx
@@ -88,7 +88,7 @@ With both curves set HyperFX fills in both directions. Omitting one curve restri
 - **Ask only** (omit `bidPriceCurve`) — the filler only *sells* exotic, filling orders where the user sends stable and receives exotic. It accumulates the stablecoin and gives away the exotic token (for example, take USDC, hand out cNGN).
 - **Bid only** (omit `askPriceCurve`) — the filler only *buys* exotic, filling orders where the user sends exotic and receives stable. It accumulates the exotic token and gives away the stablecoin.
 
-Orders in the disabled direction are rejected before bidding (an order with mixed-direction legs is rejected as a whole, since the IntentGateway settles all legs atomically). With Uniswap V4 pool pricing, the same applies: provide just one curve to price and enable a single side.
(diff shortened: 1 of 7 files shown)


Wizdave97 (commented):
it would be simpler to make the ask or bid curve optional, so filler skips orders wherfe the curve was not set instead of adding a new config

Wizdave97 (approved):

PR #979, opened 2026-06-19
[simplex]: Uniswap price guards for the FX strategy

### LINKED ISSUE #975: [simplex] Uniswap price guards
We should add a configurable reference price when using uniswap funding, if the quote from the uniswap pool ever goes a certain basis points above or below the reference price the filler should not fill orders.

royvardhan:
Closes #975.

Adds an optional price guard to the HyperFX strategy's Uniswap V4 positions. When pricing the exotic token from a Uniswap V4 pool, a manipulated, stale, or thin pool can return a bad quote. Setting `referencePrice` and `maxDeviationBps` on a position makes the filler reject orders whenever the live pool quote on that chain drifts beyond the band, before any bid is submitted.

- `referencePrice`: static expected price in exotic-per-USD (same units as the bid/ask curves).
- `maxDeviationBps`: tolerance in basis points; the two fields must be set together.
- Configured per position under `[strategies.vault.uniswapV4]`; omit both to leave a chain unguarded.

```toml
[strategies.vault.uniswapV4]
positions = [
    { chain = "EVM-8453", tokenId = "2087350", referencePrice = "1575", maxDeviationBps = 200 },
]
```

Enforcement lives in `calculateProfitability` right after venue prices are fetched — returning 0 rejects the whole order (legs settle atomically). Returning 0 rather than null is deliberate: null would silently fall back to the static/placeholder curve and still fill.

Config example and docs updated; unit test covers inside-band (incl. the edge), above, below, no-reference, and no-guard cases.

### DIFF
--- a/docs/content/developers/evm/intent-gateway/simplex.mdx
+++ b/docs/content/developers/evm/intent-gateway/simplex.mdx
@@ -141,9 +141,9 @@ Startup validation requires a pricing source: **either** full static `bidPriceCu
 
 #### Uniswap price guards
 
-Pool-based pricing trusts the live pool, which leaves the filler exposed to a manipulated, stale, or thin pool returning a bad quote. The optional **`[strategies.priceGuard]`** block bounds that risk: it sets a static reference price per chain and a tolerance in basis points. Whenever the pool quote drifts more than `maxDeviationBps` above or below the reference, the filler refuses to fill — the order is rejected before any bid is submitted.
+Pool-based pricing trusts the live pool, which leaves the filler exposed to a manipulated, stale, or thin pool returning a bad quote. To bound that risk, give a position a **`referencePrice`** and **`maxDeviationBps`**. Whenever the pool quote on that chain drifts more than `maxDeviationBps` above or below the reference, the filler refuses to fill — the order is rejected before any bid is submitted.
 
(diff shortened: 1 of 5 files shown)


docs/content/developers/evm/intent-gateway/simplex.mdx
```diff
@@ -139,6 +139,32 @@ positions = [
...
+[[strategies]]
+type = "hyperfx"
+maxOrderUsd = 5000
+
+[strategies.token1]
+"EVM-8453" = "0x46C85152bFe9f96829aA94755D9f915F9B10EF5F"
+
+[strategies.vault.uniswapV4]
+positions = [
+    { chain = "EVM-8453", tokenId = "2087350" },
```
Wizdave97: Why not just add the reference price to the positions instead of adding a new config?

docs/content/developers/evm/intent-gateway/simplex.mdx
```diff
@@ -139,6 +139,32 @@ positions = [
...
+]
+
+[strategies.priceGuard]
```
Wizdave97: We should add these config params to the `positions` object instead

docs/content/developers/evm/intent-gateway/simplex.mdx
royvardhan: Makes sense, not a big fan of adding new configs

Wizdave97 (approved):

PR #983, opened 2026-06-19
[runtime, indexer, simplex]: add phantom order support for passive price and liquidity indexing

### LINKED ISSUE #977: [runtime/simplex]: Intent Coprocessor Price and Liquidity indexing
For a price and liquidity indexing service, let's have the intent coprocessor periodically(every 1 hour) generate a phantom order(always set as expired) for supported token pairs.
The simplex filler should watch for this phantom order and submit a valid bid for the order.

The indexer should watch hyperbridge for these phantom bids, gather all bids for an order, simulate filling this order using state overrides to insert the phantom order into the chain state. 

Using only the bids that succeeded in simulations, it should calculate and store the median price, the best price, and the account balance of each filler for each token in the pair.

The phantom order should be a same chain swap on Base.
The pallet should only accept bids for a phantom order for at most 10-20 blocks after order creation.
The bid submission transaction should be signed.
A filler can only submit one bid for a phantom order.
All data should be stored offchain.

dharjeezy:
### What & why
Indexes **exchange rates** and **available liquidity** from the LPs running the simplex filler, without any real funds moving. The runtime periodically emits synthetic "phantom" orders; fillers bid on them as if they were real; we read those bids to learn each LP's price and on-chain liquidity for a token pair, and store a snapshot.

### How it works (end to end)
1. **Runtime (`pallet-intents-coprocessor`)** — governance sets a `PhantomOrderConfig` (destination chain, token pairs, `interval_blocks`). Each interval, `on_initialize` generates one phantom `Order` per pair:
   - same-chain, **zero output amount** (it only probes price), and a deadline set to the latest confirmed height so it reads as **already expired** — it can never be filled for real.
   - emits `PhantomOrderRegistered` and offchain-indexes the ABI-encoded order.
   - `on_finalize` emits `PhantomBidWindowExhausted` when a pair's bid window closes (the indexer's trigger to aggregate).
   - `place_bid` enforces phantom rules (one bid per filler, within the window); a governance guard requires `bid_window < interval_blocks`. Default window: 5 blocks (gargantua), 25 (nexus).

2. **Simplex filler** — subscribes to `PhantomOrderRegistered`, fetches the exact order from offchain storage, the **FX strategy quotes** the pair (e.g. USDC→cNGN) at its full policy price, builds a real `fillOrder` UserOp through the normal contract-interaction pipeline, and submits the bid (retracting the prior interval's bid in the same `utility.batch`).

3. **SDK (`aggregatePhantomBids`)** — the shared aggregation: pulls live bids via `intents_getBidsForOrder`, **simulates each fill via `eth_simulateV1`** against the destination chain (injects escrow/balance/allowance via state overrides, confirms an `OrderFilled` log), measures the solver's liquidity for the output token (native ERC-20 + ERC-4626 vault positions), and reduces the bids to a **liquidity-weighted median** plus the quote range. Lives in the SDK so the indexer and tests share one implementation.

4. **Indexer (SubQuery)** —
   - `handlePhantomOrderRegistered` → `PhantomOrder` entity.
   - `handlePhantomOrderPrices` (on `PhantomBidWindowExhausted`) → calls `aggregatePhantomBids` and persists the result. Gateway resolved via `INTENT_GATEWAY_V3_ADDRESSES`; token addresses normalized to 20-byte.

### Data model (GraphQL)
- **`PhantomOrder`** — the probe: chain, tokenA/tokenB, `standardAmount`, created-at block.
- **`PhantomOrderPriceSnapshot`** — per window close: tokenA/tokenB, `standardAmount`, `lowestPrice`/`medianPrice` (liquidity-weighted)/`highestPrice`, `bidCount`. The rate = `medianPrice / standardAmount` (apply token decimals).
- **`PhantomOrderLpBalance`** — per solver per snapshot: that LP's measured liquidity for the output token (keyed `{commitment}-{blockNumber}-{solver}` so it's a time series).

### Testing
- Pallet unit tests (hooks, bid rules, `window < interval` guard).
- SDK unit tests for the aggregation helpers (`weightedMedian`, `extractFillData`, `buildSimulationOrder`, slot derivation incl. the 20/32-byte normalization).
- **Full E2E** (`test:phantom-filler-e2e`): a locally-built **simnode** emits the order, **three real `IntentFiller`s** watch + quote USDC→cNGN + submit, and `aggregatePhantomBids` produces the snapshot — simulated against an **anvil fork of Base mainnet** (real gateway/tokens + `eth_simulateV1`). Wired into `test-sdk.yml`. Verified locally: quotes 1500/1510/1520 cNGN → weighted median 1510.

closes #977


modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +484,39 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Register a phantom order as the single active one. Replaces any previously
+		/// registered phantom order. Called by the intent coprocessor service.
+		#[pallet::call_index(7)]
+		#[pallet::weight(T::WeightInfo::register_phantom_order())]
+		pub fn register_phantom_order(
```
Wizdave97: This is not correct, the phantom order should be generated by the runtime in the pallet hooks

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +484,39 @@ pub mod pallet {
 
 			Ok(())
 		}
+
```
Wizdave97, line 536: There's no extrinsic to setup the phantom order details, like token pairs, chain for the order?

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderRegistered.handler.ts
```diff
@@ -0,0 +1,38 @@
+import { SubstrateEvent } from "@subql/types"
+import { wrap } from "@/utils/event.utils"
+import { getBlockTimestamp } from "@/utils/rpc.helpers"
+import { getHostStateMachine } from "@/utils/substrate.helpers"
+import { timestampToDate } from "@/utils/date.helpers"
+import { PhantomOrder } from "@/configs/src/types"
+
+export const handlePhantomOrderRegistered = wrap(async (event: SubstrateEvent): Promise<void> => {
```
Wizdave97, line 10: There's no need to index the phantom order

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomBidPlaced.handler.ts
```diff
@@ -0,0 +1,436 @@
...
+ * Simulates each ERC-20 `transferFrom(solver → beneficiary, amount)` via
+ * `eth_call` using state overrides that grant the solver infinite balance and
+ * allowance on the token contract (OZ ERC-20 storage layout: slot 0 = balances,
+ * slot 1 = allowances).
+ *
+ * Returns `true` when every simulated transfer succeeds, `false` when at least
+ * one reverts, and `null` when the simulation could not be run (e.g. RPC error,
+ * unsupported node).
+ */
+async function simulateTokenTransfers(
```
Wizdave97: This is not correct, simulate the bid as is, don't change the solver's balance, only state override needed is the block number so the order is not expired when simulating

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -461,11 +592,35 @@ pub mod pallet {
...
 		pub fn offchain_bid_key(commitment: &H256, filler: &T::AccountId) -> Vec<u8> {
 			offchain_bid_key_raw(commitment, &filler.encode())
 		}
 
+		fn compute_phantom_commitment(
+			block: BlockNumberFor<T>,
+			chain: &[u8],
+			pair: &PhantomTokenPair,
+		) -> H256 {
+			let mut preimage = Vec::new();
```
Wizdave97: The phantom order should be an actual intent gateway order.

sdk/packages/simplex/src/services/ContractInteractionService.ts
```diff
@@ -633,6 +633,58 @@ export class ContractInteractionService {
 		return { commitment, userOp: encodedUserOp }
 	}
 
+	/**
+	 * Builds a PackedUserOperation for a phantom (expired same-chain) order bid.
+	 * Uses zero relayer fees and default gas values — no estimation needed since
+	 * the order will never execute; the indexer only reads the proposed fill amounts.
+	 */
+	async preparePhantomBidUserOp(
```
Wizdave97, line 641: I don't see any event listener for phantom orders on hyperbridge

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -244,6 +295,52 @@ mod sol_types {
...
+		predispatch: sol_types::DispatchInfo { assets: vec![], call: Bytes::new() },
+		inputs: vec![sol_types::TokenInfo {
+			token: FixedBytes::from(token_a_bytes),
+			amount: AlloyU256::from(standard_amount),
+		}],
+		output: sol_types::PaymentInfo {
+			beneficiary: FixedBytes::from([0u8; 32]),
+			assets: vec![sol_types::TokenInfo {
+				token: FixedBytes::from(token_b_bytes),
+				amount: AlloyU256::from(min_output),
```
Wizdave97: The output should be zero

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -244,6 +295,52 @@ mod sol_types {
...
+	min_output: u128,
+) -> H256 {
+	use alloy_primitives::{Address, Bytes, FixedBytes, U256 as AlloyU256};
+
+	let mut token_a_bytes = [0u8; 32];
+	token_a_bytes[12..].copy_from_slice(token_a.as_bytes());
+	let mut token_b_bytes = [0u8; 32];
+	token_b_bytes[12..].copy_from_slice(token_b.as_bytes());
+
+	let order = sol_types::Order {
```
Wizdave97, line 354: This order is not added to offchain storage

parachain/runtimes/nexus/src/ismp.rs
```diff
@@ -297,12 +297,14 @@ impl pallet_bandwidth::Config for Runtime {
 
 parameter_types! {
 	pub const IntentsStorageDepositFee: Balance = EXISTENTIAL_DEPOSIT * 10;
+	pub const IntentsPhantomOrderBidWindow: u32 = 15;
```
Wizdave97: 1 hour in terms of blocks

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,224 @@
...
+			method: "POST",
+			headers: { accept: "application/json", "content-type": "application/json" },
+			body: JSON.stringify({
+				id: 1,
+				jsonrpc: "2.0",
+				method: "eth_call",
+				params: [
+					{ from: solver, to: solver, data: callData },
+					"latest",
+					{ [gatewayAddress]: { code: "0x" } },
```
Wizdave97: This is wrong, the bid should be validated against intent gatway's code.
the override required is inserting the phantom order commitment into the intent gateway's contract state

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -244,6 +295,52 @@ mod sol_types {
...
+		user: FixedBytes::from([0u8; 32]),
+		source: Bytes::copy_from_slice(chain),
+		destination: Bytes::copy_from_slice(chain),
+		deadline: AlloyU256::ZERO,
+		nonce: AlloyU256::from(block),
+		fees: AlloyU256::ZERO,
+		session: Address::ZERO,
```
Wizdave97: You can't use a zero session key, generate a static key with a known private key that would always be used for all phantom orders.
Please refer to the sdk for how bid simulation works.

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,224 @@
...
+				if (!fillData?.outputs.length) continue
+
+				const output = fillData.outputs[0]
+				const outputAmount = output.amount
+
+				const simOk = await simulateBid(evmUrl, solver, callData, gatewayAddress)
+				if (!simOk) continue
+
+				const tokenAddress = bytes32ToBytes20(output.token)
+				const balance = await getTokenBalance(evmUrl, tokenAddress, solver)
```
Wizdave97: You need to fetch the filler's balance from the Aave's USDC/USDT vault and cngn yield vault also

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,224 @@
...
+				if (balance === null || balance < outputAmount) continue
```
Wizdave97: Why are we not storing the LP's balance?

Wizdave97: This check is not necessary, if the simulation succeeded the filler must have some balance somewhere
```suggestion
```

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +512,97 @@ pub mod pallet {
...
+				},
+			};
+
+			if !should_generate {
+				return T::DbWeight::get().reads(2);
+			}
+
+			// Use pallet-ismp's time provider so the deadline is a real chain timestamp
+			// (non-zero). Simulation uses block 0 (timestamp=0) so the check still passes.
+			let deadline_secs = <T as pallet_ismp::Config>::TimestampProvider::now().as_secs();
```
Wizdave97: Deadline is in blocks, I meant fetch the latest state machine height for the order chain and use that as the deadline

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +512,97 @@ pub mod pallet {
...
+				CurrentPhantomOrder::<T>::put((commitment, info));
+				offchain_index::set(&offchain_phantom_key(&commitment), &order_bytes);
+				Self::deposit_event(Event::PhantomOrderRegistered {
+					commitment,
+					chain: chain_bytes.clone(),
+					created_at: n,
+					token_a: pair.token_a,
+					token_b: pair.token_b,
+					standard_amount: pair.standard_amount,
+					min_output: pair.min_output,
```
Wizdave97: Remove this min_output config, it's not useful

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,280 @@
...
+// gateway treats the phantom commitment as a registered order.
+async function simulateBid(
+	evmRpcUrl: string,
+	solver: string,
+	callData: HexString,
+	gatewayAddress: string,
+	commitment: string,
+): Promise<boolean> {
+	try {
+		const { slot1, slot2 } = requestCommitmentKey(commitment as HexString)
```
Wizdave97: Why are you using request commitment keys?
You are meant to override this storage in IntentGateway contract, please check how the contract works 
```
    mapping(bytes32 => mapping(address => uint256)) public _orders;
```

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,280 @@
...
+		const response = await fetchWithRetry(evmRpcUrl, {
+			method: "POST",
+			headers: { accept: "application/json", "content-type": "application/json" },
+			body: JSON.stringify({
+				id: 1,
+				jsonrpc: "2.0",
+				method: "eth_call",
+				params: [
+					{ from: solver, to: solver, data: callData },
```
Wizdave97: This is not correct, is the to not meant to be the intent gateway contract?

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,280 @@
...
+
+				const fillData = extractFillData(callData, gatewayAddress)
+				if (!fillData?.outputs.length) continue
+
+				const simOk = await simulateBid(evmUrl, solver, callData, gatewayAddress, phantom.id)
+				if (!simOk) continue
+
+				const output = fillData.outputs[0]
+				const tokenAddress = bytes32ToBytes20(output.token)
+				const totalBalance = await getTotalSolverBalance(evmUrl, phantom.chain, tokenAddress, solver)
```
Wizdave97: Get solver balance across all configured evm chains

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,280 @@
...
+
+				prices.push(output.amount)
+				if (totalBalance > bestLpBalance) bestLpBalance = totalBalance
```
Wizdave97: This is wrong, store the balance of each Lp that submitted a valid bid independently, also balances should be stored by token

sdk/packages/indexer/src/yield-vault-addresses.ts
```diff
@@ -0,0 +1,57 @@
+// ERC-4626 vault addresses per chain, keyed by underlying token address (lowercase).
+// Aave stata token addresses sourced from https://github.com/bgd-labs/aave-address-book
+// Values are arrays because multiple vaults may wrap the same underlying token.
+export const YIELD_VAULT_ADDRESSES: Record<string, Record<string, string[]>> = {
```
Wizdave97: Add this to the builld pipeline like other addresses, don't harcode it here

sdk/packages/indexer/src/yield-vault-addresses.ts
```diff
@@ -0,0 +1,57 @@
...
+	// Ethereum mainnet
+	"EVM-1": {
+		// USDC → stataUSDC
+		"0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48": ["0xD4fa2D31b7968E448877f69A96DE69f5de8cD23E"],
+		// USDT → stataUSDT
+		"0xdac17f958d2ee523a2206206994597c13d831ec7": ["0x7Bc3485026Ac48b6cf9BaF0A377477Fff5703Af8"],
+	},
+	// Optimism
+	"EVM-10": {
+		// USDC.e (bridged) → stataUSDC.e
```
Wizdave97: Only native USDC

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -729,4 +741,86 @@ export class IntentFiller {
...
+		const ZERO_BYTES32 = `0x${"00".repeat(32)}` as HexString
+		const ZERO_ADDRESS = `0x${"00".repeat(20)}` as HexString
+		const phantomOrder: Order = {
+			user: ZERO_BYTES32,
+			source: event.chain,
+			destination: event.chain,
+			deadline: 0n,
+			nonce: BigInt(event.createdAt),
+			fees: 0n,
+			session: ZERO_ADDRESS,
```
Wizdave97: This session address is not correct

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -729,4 +741,86 @@ export class IntentFiller {
...
+		const entryPointAddress = this.configService.getEntryPointAddress(`EVM-${getChainId(event.chain) ?? event.chain}`)
+		if (!entryPointAddress) {
+			this.logger.debug({ chain: event.chain }, "No entry point configured for phantom order chain, skipping")
+			return
+		}
+
+		// Reconstruct the Order with the same field values the pallet used when computing the commitment.
+		const ZERO_BYTES32 = `0x${"00".repeat(32)}` as HexString
+		const ZERO_ADDRESS = `0x${"00".repeat(20)}` as HexString
+		const phantomOrder: Order = {
```
Wizdave97: This order is wrong, fetch it from offchain storage

Wizdave97:
Understood — locking in those decisions. Here is the updated findings section. Note one renumbering: your "Mod 3 → retract previous phantom bid" actually resolves **Mod2 (deposit lockup)**, so I've applied it there and kept the original Mod3 (commitment-parity) separate.

---

## 🔴 Blocking (unchanged)

**B1** — Wire both handlers into the manifest (`handlePhantomOrderRegistered` as an `intentsCoprocessor.PhantomOrderRegistered` EventHandler; `handlePhantomOrderPrices` as a `substrate/BlockHandler` with `modulo: 10`).
**B2** — Re-type `handlePhantomOrderPrices` to `SubstrateBlock` and use `event.block.header.number/.hash`; drop the now-redundant internal `% 10n` guard.

---

## 🟠 Major — decisions locked

### M1 — Make the simulation real (solution **b**)
Change the simulated fill so it actually exercises the solver delivering `solverAmount` of the output token, instead of the current zero-output no-op.

Required changes in `simulateBid` / `extractFillData`:
1. **Rebuild the simulated order** with `output.assets[0].amount = solverAmount` (the solver's quoted amount) so `totalRequired > 0` and `_fillSameChain` actually transfers.
2. **Use a non-zero beneficiary.** Set `output.beneficiary = solver` so `safeTransferFrom(solver → solver)` still validates the solver's **balance and allowance** without tokens leaving (avoids the `address(0)` revert on USDC-class tokens).
3. **Recompute the commitment.** Because the contract derives `commitment = keccak256(abi.encode(order))` from the calldata, modifying the output amount/beneficiary changes the commitment. The state-diff escrow injection must therefore key on the **recomputed** commitment (`ordersStorageSlot(newCommitment, inputTokenBytes32)`), not `phantom.id`.
4. **Inject the input escrow** at that slot = the order's input amount, so `_withdraw` releases it to the solver and the fill completes.
5. **Override the solver's ERC-20 allowance** to the gateway (the extracted `fillCalldata` is only the inner `fillOrder` call — the batch's approval step is dropped), or simulate the full ERC-7821 batch. Without this the `transferFrom` reverts on lack of allowance and the sim fails for everyone.

After this, a passing simulation means the solver genuinely holds and has authorized `solverAmount` — so `bidCount`/prices reflect *fillable* bids. The `ordersStorageSlot`/block-0 machinery becomes load-bearing as originally intended.

### M1 companion — solver selection in the sim
Disable solver selection inside the `eth_call`: **state-override `_params.solverSelection` to `false`** (so `fillOrder` skips the transient `tload(commitment)` selection check), and **set the order's `session` key to the zero address**. This removes the `Unauthorized` revert path without needing to replicate the `select()` transient-storage flow. (This supersedes the earlier "verify solverSelection is off" minor note — it is now an explicit override.)

### M2 — Per-chain token mapping for liquidity
Replace the "same address on every chain" sum in `getTotalSolverBalanceAllChains`: **map the output token to its per-chain equivalent before summing** (resolve each chain's canonical address for that asset, e.g. via a token-equivalence map keyed off the destination-chain token, and skip chains with no mapping). Then **clarify the schema docs**:
- `PhantomOrderPriceSnapshot.lpBalance` — total liquidity of the highest-liquidity bidder, summed across chains using **per-chain equivalents** of the output token (raw ERC-20 + redeemable ERC-4626).
- `PhantomOrderLpBalance.balance` — same per-solver, same cross-chain-equivalent basis.

---

## 🟡 Moderate — decisions locked

### Mod2 — Free phantom bid deposits via retraction (your "Mod 3")
In the simplex filler, **submit a `retract_bid` for the previous phantom commitment whenever submitting a bid for a new phantom commitment.** Track the last phantom commitment the filler bid on (per chain), and in `handlePhantomOrder`, before/after `submitBid`, call `coprocessor.retractBid(previousCommitment)` to unreserve the prior `StorageDepositFee`. This bounds reserved deposits to ~one window's worth instead of growing unbounded across intervals.

### Mod1 — phantom bid rules apply to only one pair per interval (still open)
Unchanged recommendation: the hook overwrites `CurrentPhantomOrder` per iteration, so window/duplicate enforcement in `place_bid` covers only the last pair. Track the full set of active commitments if the rules must cover every pair.

### Mod3 — commitment-parity fragility (still open, separate from the retraction item)
Unchanged recommendation: have the filler submit under `event.commitment` (or `keccak256` of the exact stored bytes) rather than re-deriving via decode→re-encode.

---

## 🟢 Minor (still open)
- Confirm RPC support for the 4th `blockOverrides` arg to `eth_call` (now load-bearing again under M1).
- `medianOf` returns upper-middle for even counts — add a comment.
- Add indexer-handler tests (`ordersStorageSlot`, `extractFillData`, snapshot aggregation, and the new allowance/escrow overrides under M1).

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +511,110 @@ pub mod pallet {
...
+
+			if !should_generate {
+				return T::DbWeight::get().reads(2);
+			}
+
+			// Use the latest confirmed EVM block for the destination chain as the deadline
+			// base. The gateway checks `order.deadline < block.number`, so we add the
+			// current interval as a buffer to keep the order valid for one full window.
+			let chain_bytes = config.chain.to_string().into_bytes();
+			let opt_latest_height = LatestStateMachineHeight::<T>::iter()
```
Wizdave97: Why iterating, the config should hold type `StateMachineId` that holds the consensus state id

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +511,110 @@ pub mod pallet {
...
+				.max();
+			let Some(latest_evm_height) = opt_latest_height else {
+				log::warn!(
+					target: "pallet-intents",
+					"No confirmed state machine height for {:?}, skipping phantom order generation",
+					config.chain,
+				);
+				return T::DbWeight::get().reads(3);
+			};
+			let deadline = latest_evm_height.saturating_add(config.interval_blocks as u64);
```
Wizdave97: This is wrong, use the latest_state machine height as is

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +511,110 @@ pub mod pallet {
...
+			// current interval as a buffer to keep the order valid for one full window.
+			let chain_bytes = config.chain.to_string().into_bytes();
+			let opt_latest_height = LatestStateMachineHeight::<T>::iter()
+				.filter_map(|(id, height)| {
+					if id.state_id == config.chain { Some(height) } else { None }
+				})
+				.max();
+			let Some(latest_evm_height) = opt_latest_height else {
+				log::warn!(
+					target: "pallet-intents",
```
Wizdave97: log target should be defined as a pallet constant

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -135,6 +137,31 @@ pub struct GatewayInfo {
...
+	pub token_a: H160,
+	pub token_b: H160,
+	pub standard_amount: u128,
+}
+
+/// Governance-settable configuration for autonomous phantom order generation.
+/// Stored in `PhantomOrderConfig`; the pallet hook reads it every block.
+#[derive(Clone, Debug, Encode, Decode, DecodeWithMemTracking, TypeInfo, PartialEq, Eq)]
+pub struct PhantomOrderConfiguration {
+	pub chain: StateMachine,
```
Wizdave97:
```suggestion
	pub chain: StateMachineId,
```

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -244,6 +294,59 @@ mod sol_types {
 	}
 }
 
+/// Fixed session address for all phantom orders, derived from private key 0x01.
+/// The indexer uses the matching private key when signing simulated bids.
+pub const PHANTOM_SESSION_ADDRESS: alloy_primitives::Address =
```
Wizdave97: No longer necessary, disable solver selection when simulating the bid

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,146 @@ type BandwidthTier @entity {
...
+	"""
+	Hyperbridge block number at which this phantom order was registered (emitted via PhantomOrderRegistered).
+	All phantom orders sharing the same createdAtBlock belong to the same generation interval.
+	"""
+	createdAtBlock: BigInt! @index
+
+	"""
+	Timestamp of the Hyperbridge block at which this phantom order was registered.
+	"""
+	blockTimestamp: Date
```
Wizdave97:
```suggestion
	blockTimestamp: Date @index
```

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,146 @@ type BandwidthTier @entity {
...
+	"""
+	Hyperbridge block number at which this snapshot was taken.
+	"""
+	blockNumber: BigInt! @index
+
+	"""
+	Median output amount across all valid bids, in the smallest unit of the output token (tokenB).
+	Null if no valid bids were found.
+	"""
+	medianPrice: BigInt
```
Wizdave97:
```suggestion
	medianPrice: BigInt @index
```

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,146 @@ type BandwidthTier @entity {
...
+
+	"""
+	Lowest output amount across all valid bids — the least the user would receive (most favorable for the solver).
+	Null if no valid bids were found.
+	"""
+	lowestPrice: BigInt
```
Wizdave97:
```suggestion
	lowestPrice: BigInt @index
```

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,146 @@ type BandwidthTier @entity {
...
+
+	"""
+	Highest output amount across all valid bids — the most the user would receive (most favorable for the user).
+	Null if no valid bids were found.
+	"""
+	highestPrice: BigInt
```
Wizdave97:
```suggestion
	highestPrice: BigInt @index
```

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,146 @@ type BandwidthTier @entity {
...
+	"""
+	Number of valid bids that contributed to this snapshot (passed simulation).
+	"""
+	bidCount: Int!
+
+	"""
+	Total liquidity available from the winning solver across raw ERC-20 balance and ERC-4626
+	vault positions (e.g. Aave stataTokens, yield vaults) for the output token.
+	"""
+	lpBalance: BigInt
```
Wizdave97:
```suggestion
```

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,146 @@ type BandwidthTier @entity {
...
+	"""
+	EVM address of the solver whose balance is recorded.
+	"""
+	solver: String! @index
+
+	"""
+	Total solver balance for the output token across all configured EVM chains,
+	including raw ERC-20 balance and redeemable ERC-4626 vault positions.
+	"""
+	balance: BigInt!
```
Wizdave97, line 2359: Balances should be mapped by token, you can't just add balances of different tokens into one value

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,374 @@
...
+			headers: { accept: "application/json", "content-type": "application/json" },
+			body: JSON.stringify({
+				id: 1,
+				jsonrpc: "2.0",
+				method: "eth_call",
+				params: [{ from: solver, to: gatewayAddress, data: callData }, "latest", stateDiff],
+			}),
+		})
+		const result = await response.json()
+		return !result.error
```
Wizdave97: Inspect the logs returned for OrderFilled Event, that's how we know it succeeded

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,374 @@
...
+		})
+		const result = await response.json()
+		if (result.error || !result.result || result.result === "0x") return 0n
+		return BigInt(result.result)
+	} catch {
+		return 0n
+	}
+}
+
+async function getTotalSolverBalance(
```
Wizdave97: You can't add balances of distinct tokens into a single value

sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -0,0 +1,374 @@
...
+	const blockNumber = event.block.header.number.toBigInt()
+	const blockHash = event.block.header.hash.toString()
+
+	const activeCommitment = await getActiveCommitment(blockHash)
+	if (!activeCommitment) return
+
+	const anchor = await PhantomOrder.get(activeCommitment)
+	if (!anchor) return
+
+	const phantomOrders = await PhantomOrder.getByCreatedAtBlock(anchor.createdAtBlock, { limit: 100 })
```
Wizdave97: Process only the active phantom order

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +511,110 @@ pub mod pallet {
...
+		}
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T>
+	where
+		T::AccountId: From<[u8; 32]>,
+	{
+		fn on_initialize(n: BlockNumberFor<T>) -> Weight {
+			let Some(config) = PhantomOrderConfig::<T>::get() else {
```
Wizdave97, line 647: Let's add a storage item to mark when the bid window for phantom orders has been exhausted so the indexer can use that as the pointer to do it's aggregation

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -135,6 +137,37 @@ pub struct GatewayInfo {
 	pub params: IntentGatewayParams,
 }
 
+/// Upper bound on the token pairs a single config may probe, and therefore on the number
+/// of phantom orders active at once.
+pub const MAX_PHANTOM_TOKEN_PAIRS: u32 = 10;
```
Wizdave97:
```suggestion
pub const MAX_PHANTOM_TOKEN_PAIRS: u32 = 64;
```

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -729,4 +743,84 @@ export class IntentFiller {
...
+			const { userOp } = await this.contractService.preparePhantomBidUserOp(
+				phantomOrder,
+				entryPointAddress,
+				solverAccountAddress,
+				fillerOutputs,
+			)
+
+			const prevCommitment = this.lastPhantomCommitmentByChain.get(event.chain)
+			if (prevCommitment && prevCommitment !== event.commitment) {
+				// Retract the previous bid so its deposit is reclaimed before the new interval starts.
```
Wizdave97: Use a utility batchAll to submit the retraction and bid submission in one transaction

sdk/packages/indexer/src/yield-vault-addresses.ts
Wizdave97: Let's not commit this file

sdk/packages/indexer/src/token-slot-overrides.ts
```diff
@@ -0,0 +1,29 @@
+// Auto-generated, DO NOT EDIT
+// Per-token ERC-20 storage slot overrides for tokens that don't follow the standard OZ layout
+// (slot 0 = _balances, slot 1 = _allowances).
+// To add or update entries, edit the "tokenSlots" field in the relevant chain entry
+// in src/configs/config-mainnet.json (or config-testnet.json) and re-run codegen.
+export const TOKEN_SLOT_OVERRIDES: Record<string, { balanceSlot: bigint; allowanceSlot: bigint }> = {
+	// USDC (Circle FiatToken) — balances at slot 9, allowed at slot 10
```
Wizdave97: Don't commit this file

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +528,123 @@ pub mod pallet {
...
+				let mut exhausted: BoundedVec<H256, ConstU32<MAX_PHANTOM_TOKEN_PAIRS>> =
+					BoundedVec::new();
+				for (commitment, info) in active.iter() {
+					Self::deposit_event(Event::PhantomBidWindowExhausted {
+						commitment: *commitment,
+						created_at: info.created_at_block,
+					});
+					let _ = exhausted.try_push(*commitment);
+				}
+				ExhaustedPhantomOrder::<T>::put(exhausted);
```
Wizdave97: There's no need for this storage item 
```suggestion
```

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +528,123 @@ pub mod pallet {
...
+					target: LOG_TARGET,
+					"No confirmed state machine height for {:?}, skipping phantom order generation",
+					config.chain,
+				);
+				return T::DbWeight::get().reads(3);
+			};
+
+			// The previous batch has seen every bid it will get. Signal each commitment so the
+			// indexer can aggregate, and record them as the exhausted pointer before replacing.
+			if let Some(active) = active {
```
Wizdave97: This should happen before should_generate block above, if not its useless

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -443,6 +528,123 @@ pub mod pallet {
...
+	where
+		T::AccountId: From<[u8; 32]>,
+	{
+		fn on_initialize(n: BlockNumberFor<T>) -> Weight {
+			let Some(config) = PhantomOrderConfig::<T>::get() else {
+				return Weight::zero();
+			};
+
+			let active = CurrentPhantomOrder::<T>::get();
+			let should_generate = match active.as_ref().and_then(|orders| orders.first()) {
```
Wizdave97: Let's just use a distinct storage value to keep track of the last time phantom orders were generated

Wizdave97 (commented):
.

Wizdave97 (dismissed):

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,145 @@ type BandwidthTier @entity {
...
+type PhantomOrderPriceSnapshot @entity {
+	"""
+	Composite identifier: {commitment}-{blockNumber}.
+	"""
+	id: ID!
+
+	"""
+	The phantom order commitment this snapshot belongs to. References PhantomOrder.id.
+	"""
+	commitment: String! @index
```
Wizdave97, line 2258: Where's the token information for this snapshot? This does not tell use which tokens own this exchange rate
```suggestion
```

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2192,3 +2192,145 @@ type BandwidthTier @entity {
...
+Per-solver, per-token liquidity balance recorded at a phantom order price snapshot.
+One entity is written for every valid bidder whose fill simulation passes. The balance covers a
+single output token on the order's destination chain, as raw ERC-20 plus redeemable ERC-4626
+vault positions. Balances of different tokens are never combined into one value.
+"""
+type PhantomOrderLpBalance @entity {
+	"""
+	Composite identifier: {commitment}-{blockNumber}-{solverAddress}.
+	"""
+	id: ID!
```
Wizdave97, line 2328: Commitment  is not useful in the Ids here, Id should be token-solver address

seunlanlege (approved):

PR #984, opened 2026-06-22
[simplex]: fill from funding venues first, reserving the wallet minBalance

royvardhan:
Fills now reserve the vault's `minBalance` floor in the wallet and never spend below it, keeping it liquid for the gas/paymaster pull. Output is sourced from the free wallet balance (above the reserve) first, then any remaining shortfall is withdrawn from the configured funding venue (the ERC-4626 vault).

Previously the fill spent the full wallet balance, ignoring `minBalance`. The Circle paymaster pulls USDC during `validatePaymasterUserOp`, which runs before the fill executes, so on a small fill the paymaster took its gas out of the same wallet USDC the fill was about to pay out, and `fillOrder` reverted with `ERC20: transfer amount exceeds balance` (basescan tx `0xd5c8c0…fa94f9`: output 797,833, balance 772,247, short by the paymaster's exact 25,586 pull). Reserving `minBalance` keeps the paymaster's slice liquid.

Applies to both the FX and stable strategies. A new `walletReserveForToken` on the funding-venue interface exposes the floor; the vault venue returns its `minBalance`, Uniswap V4 returns zero.

Note for operators: set `minBalance` on the USDC vault to comfortably exceed the worst-case paymaster prefund per chain.

Also skips the fee-token (USDC) approval on same-chain fills. Same-chain orders release escrow locally with no Hyperbridge dispatch, so the gateway never pulls the fee token; only cross-chain fills, which dispatch a RedeemEscrow message paid in the fee token, still approve it. Output-token approvals are unchanged.

### DIFF
--- a/sdk/packages/simplex/src/strategies/fx.ts
+++ b/sdk/packages/simplex/src/strategies/fx.ts
@@ -455,10 +455,10 @@ export class FXFiller implements FillerStrategy {
 					)
 				}
 
-				// Source the output from funding venues first (the vault), leaving the wallet
-				// untouched. Only what the venues can't cover is drawn from the wallet, and
-				// never below the configured minBalance reserve — kept liquid for the
-				// gas/paymaster pull during validatePaymasterUserOp.
+				// Spend the free wallet balance first, down to the configured minBalance
+				// reserve — kept liquid for the gas/paymaster pull during
+				// validatePaymasterUserOp — then source any remaining shortfall from the
+				// funding venues (the vault).
 				const tokenAddress = bytes32ToBytes20(output.token).toLowerCase()
 				const balance = await this.getAndCacheBalance(tokenAddress, walletAddress, destClient, balanceCache)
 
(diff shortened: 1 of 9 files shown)


sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -455,22 +455,34 @@ export class FXFiller implements FillerStrategy {
 					)
 				}
 
-				// Cap by wallet balance on the destination chain, optionally topped up via LP removal.
+				// Source the output from funding venues first (the vault), leaving the wallet
+				// untouched. Only what the venues can't cover is drawn from the wallet, and
+				// never below the configured minBalance reserve — kept liquid for the
+				// gas/paymaster pull during validatePaymasterUserOp.
```
seunlanlege: I think we should actually consume free balances first accounting for the minimum balance, before sourcing from the vault.

The vault is a yield-bearing position, & some vaults may not be available for deposits all the time. So better to let go of non-yield bearing free balances before the vault balance

royvardhan: Done

seunlanlege (approved):
Don't forget to bump simplex version

PR #988, opened 2026-06-24
[evm/runtime]: make IntentGatewayV2 upgradeable via cross-chain governance

### LINKED ISSUE #981: [evm/runtime]: make IntentGatewayV2 upgradeable via cross-chain governance
## Motivation

`IntentGatewayV2` is currently deployed as a non-upgradeable contract (`new IntentGatewayV2{salt}(admin)` in `DeployIntentGateway.s.sol`). Any logic change requires a fresh deployment and re-registration of the gateway address across all chains — disruptive given cross-chain auth keys off the gateway address.

`onAccept` already implements cross-chain governance: `RequestKind.UpdateParams`/`NewDeployment`/`SweepDust` are authorized by `source == IDispatcher(host()).hyperbridge()`. We can make the gateway upgradeable by the **same authority** (the Hyperbridge coprocessor) by adding one more `RequestKind` — no new trust assumptions, no new message plumbing.

### EVM (`evm/`)

1. **Wrap the gateway in an `ERC1967Proxy`.** The implementation stays a plain contract — **no `UUPSUpgradeable`**. The upgrade entry point is `onAccept`, not the standard UUPS `upgradeToAndCall`, so `UUPSUpgradeable`/`_authorizeUpgrade`/`proxiableUUID` buy nothing here (the `proxiableUUID` rollback check verifies the wrong invariant — it can't certify that a new impl keeps the `onAccept` upgrade branch). Use `ERC1967Utils.upgradeToAndCall` directly for the slot write.

2. **Add `RequestKind.UpgradeContract`** (append last, index `5`) in `IntentsBase.sol`, and handle it inside the existing `source == hyperbridge` block of `ExtrinsicIntents.onAccept`:
   ```solidity
   } else if (kind == RequestKind.UpgradeContract) {
       (address newImpl, bytes memory initData) =
           abi.decode(incoming.request.body[1:], (address, bytes));
       ERC1967Utils.upgradeToAndCall(newImpl, initData); // reverts if newImpl has no code
   }
   ```
   Keep this branch in the inherited base so it can't be accidentally dropped on a future upgrade.

3. **Constructor → initializer, and remove `_admin`.** Behind a proxy the constructor runs against the implementation's storage, so anything it sets is unreachable at runtime. Replace the constructor + `init()` with `initialize(params, deployments)`, run **atomically via the proxy's `initData`** (one tx → no front-run window). `_admin` exists only to gate the one-time `init()`; with atomic init + a one-shot guard it is vestigial — there is no other consumer, and every config path already goes through `onAccept` governance. **Remove `_admin` and the `admin` constructor param**, and gate `initialize` with an explicit `_initialized` flag instead. Removing `_admin` only shifts the slots of variables declared *after* it (`_filled` is declared before it and stays at slot 2), so it is safe for this genesis deployment — but it must be done now, since the append-only rule forbids removing it in any later upgrade. Also call `_disableInitializers()` in the impl constructor to lock the raw implementation. `EIP712` is proxy-safe (OZ recomputes the domain separator when `address(this)` differs from the cached address).

   ```solidity
   // implementation — plain logic contract; constructor only locks the raw impl
   constructor() EIP712("IntentGateway", "2") {
       _disableInitializers();
   }

   function initialize(Params memory p, Deployment[] memory deployments) external {
       if (_initialized) revert Unauthorized();   // reinit guard (replaces the _admin gate)
       _initialized = true;
       for (uint256 i; i < deployments.length; i++) _addDeployment(deployments[i]);
       _validateParams(p);
       _params = p;
   }

   // deployment — both via CREATE2 with the same salt; init runs atomically in the
   // proxy constructor, so there is no uninitialized window to front-run.
   IntentGatewayV2 impl = new IntentGatewayV2{salt: salt}();
   bytes memory initData = abi.encodeCall(IntentGatewayV2.initialize, (params, deployments));
   ERC1967Proxy proxy   = new ERC1967Proxy{salt: salt}(address(impl), initData);
   // `address(proxy)` is the canonical gateway address used everywhere below.
   ```

4. **Deploy script (`DeployIntentGateway.s.sol`).** Deploy impl and proxy *both* via CREATE2 with the same salt; pass `initData = abi.encodeCall(IntentGatewayV2.initialize, (params, deployments))` to the proxy constructor so initialization is atomic. Point all `Deployment[]` entries, `SolverAccount`, and the `INTENT_GATEWAY_V2` config key at the **proxy** address.

### Coprocessor (`modules/pallets/intents-coprocessor/`)

5. Add a `RequestKind::UpgradeContract { new_impl: H160, init_data: Vec<u8> }` variant in `types.rs` and extend `encode_body()` to emit discriminator byte `5` + `abi.encode(newImpl, initData)`. Keep the byte value in sync with the EVM enum (`IntentGatewayRequestKind` currently: `UpdateParams = 2`, `SweepDust = 3`).

6. Add an `upgrade_gateway(origin, state_machine, new_impl, init_data)` extrinsic gated by `GovernanceOrigin`, mirroring `update_params`: build `RequestKind::UpgradeContract`, `encode_body()`, and `Self::dispatch(state_machine, gateway, body)`.

## Hard constraints

- **Proxy address must be identical on every chain.** Cross-chain auth (`_authenticate` → `_instance`) requires the registered gateway address to match across chains. `ERC1967Proxy` init code embeds the impl address, so the **impl must also be deployed deterministically** (same salt + byte-identical bytecode → same impl address → same proxy address). No chain-specific immutables baked into the impl.
- **`_filled` must remain at storage slot 2.** `FILLED_SLOT_BIG_ENDIAN_BYTES` and cross-chain cancel storage proofs depend on it. EIP-1967's hashed implementation slot avoids collision, so the proxy migration preserves it — but **all future upgrades may only append storage variables, never reorder.**
- **Bricking risk.** The upgrade authority is solely the Hyperbridge coprocessor (no EOA escape hatch). A future impl that drops the `UpgradeContract` branch loses upgradability permanently — keep it in the base contract and regression-test its presence.

## Acceptance criteria

- [ ] `IntentGatewayV2` deployed behind `ERC1967Proxy`; proxy address deterministic and identical across all target chains.
- [ ] `RequestKind.UpgradeContract` handled in `onAccept`, authorized by `source == hyperbridge`; rejects non-coprocessor sources.
- [ ] `_admin` removed (constructor param dropped); `initialize` gated by an explicit `_initialized` reinit guard and run atomically via the proxy `initData`; impl constructor calls `_disableInitializers()`.
- [ ] Public/EOA upgrade paths are closed — upgrades only via `onAccept`.
- [ ] `intents-coprocessor` exposes a `GovernanceOrigin`-gated extrinsic that dispatches the upgrade message; wire-byte parity with the EVM enum.
- [ ] Foundry tests: successful governance upgrade preserves state (`_filled`, `_orders`, `_nonce`); `_filled` stays at slot 2; unauthorized-source and no-code-impl upgrades revert.

royvardhan:
Makes `IntentGatewayV2` upgradeable by the Hyperbridge coprocessor — the same authority that already governs `UpdateParams`/`SweepDust` — through a new `RequestKind.UpgradeContract` handled in `onAccept`. The gateway runs behind an `ERC1967Proxy`, initialized atomically at deployment.

### EVM
- Wrap the gateway in `ERC1967Proxy`. The implementation stays a plain contract (no UUPS); upgrades go through `ERC1967Utils.upgradeToAndCall` in `onAccept`, gated by `source == hyperbridge`. The branch lives in the inherited base so it can't be dropped by a future upgrade.
- Replace the constructor `init` with an `initializer`-guarded `initialize`; `_disableInitializers()` locks the raw implementation. The init gate is an immutable `_owner` (replaces storage `_admin`, zero slots). `_filled` stays at storage slot 2.
- Initialize atomically through the proxy's `initData`. `_instance()` resolves to `address(this)`, so cross-chain peers share the gateway's address rather than a stored registry, and the self-referential deployments array is dropped. `NewDeployment` remains as a governance override.
- Append a `_paused` storage flag (slot 13, after all proof-sensitive slots); the pause logic is deferred.

### Coprocessor
- `RequestKind::UpgradeContract { new_impl, init_data }` and a `GovernanceOrigin`-gated `upgrade_gateway` extrinsic mirroring `update_params`. Wire byte `5` matches the EVM enum; the payload is encoded to match the Solidity `abi.decode(body[1:], (address, bytes))`.

### Determinism
Atomic init makes the proxy address depend on the encoded `Params`. Those are byte-identical across chains (`host`, `dispatcher`, `solverSelection` uniform), so the proxy lands at the same address everywhere — which is exactly what lets `_instance()` resolve peers to `address(this)`. This relies on `ADMIN` and `Params` staying uniform across chains, the discipline the deployment already follows.

Closes #981.

### DIFF
--- a/evm/script/DeployIntentGateway.s.sol
+++ b/evm/script/DeployIntentGateway.s.sol
@@ -24,7 +24,7 @@ contract DeployScript is BaseScript {
         // init data. The address depends on (impl address, salt, params, peer chain ids), all of
         // which are identical across chains, keeping the proxy address identical everywhere.
         address priceOracle = address(0);
-        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}(admin);
+        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}();
         bytes[] memory peerChains;
         if (config.get("is_mainnet").toBool()) {
             peerChains = new bytes[](9);
--- a/evm/script/DeployIntentGateway.s.sol
+++ b/evm/script/DeployIntentGateway.s.sol
@@ -4,13 +4,12 @@ pragma solidity ^0.8.17;
 import "forge-std/Script.sol";
 import "stringutils/strings.sol";
 
-import {IntentGatewayV2, Params, Deployment} from "../src/apps/IntentGatewayV2.sol";
+import {IntentGatewayV2, Params} from "../src/apps/IntentGatewayV2.sol";
 import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 import {BaseScript} from "./BaseScript.sol";
 import {CallDispatcher} from "../src/utils/CallDispatcher.sol";
 import {SolverAccount} from "../src/apps/intentsv2/SolverAccount.sol";
 import {VWAPOracle} from "../src/utils/VWAPOracle.sol";
-import {StateMachine} from "@hyperbridge/core/libraries/StateMachine.sol";
 
 contract DeployScript is BaseScript {
     using strings for *;
@@ -18,86 +17,39 @@ contract DeployScript is BaseScript {
     /// @notice Main deployment logic - called by BaseScript's run() functions
     /// @dev This function is called within a broadcast context
     function deploy() internal override {
-        // Deploy the implementation and proxy both via CREATE2 with the same salt. The proxy
-        // is created with empty init data so its address depends only on (impl address, salt) —
-        // identical across chains. Initialization runs as a separate, deployer-gated call below.
-        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}(admin);
-        console.log("IntentGateway implementation deployed at:", address(implementation));
-
-        ERC1967Proxy proxy = new ERC1967Proxy{salt: salt}(address(implementation), "");
-        IntentGatewayV2 intentGateway = IntentGatewayV2(payable(address(proxy)));
-        console.log("IntentGateway proxy deployed at:", address(intentGateway));
-
-        SolverAccount solverAccount = new SolverAccount{salt: salt}(address(intentGateway));
-        console.log("SolverAccount deployed at:", address(solverAccount));
-
+        // Deploy implementation and proxy via CREATE2 with the same salt. The proxy is initialized
+        // atomically through its init data, so its address depends on (impl address, salt, params).
+        // Params are identical across chains, keeping the proxy address identical everywhere — which
+        // lets cross-chain peers resolve to `address(this)` rather than a stored registry.
         address priceOracle = address(0);
         // address priceOracle = deployPriceOracle(address(intentGateway));
 
-        Deployment[] memory deployments;
-        if (config.get("is_mainnet").toBool()) {
-            deployments = new Deployment[](9);
-            deployments[0] = Deployment({
-                chain: StateMachine.evm(1), // ethereum
-                gateway: address(intentGateway)
-            });
-            deployments[1] = Deployment({
-                chain: StateMachine.evm(10), // optimism
-                gateway: address(intentGateway)
-            });
-            deployments[2] = Deployment({
-                chain: StateMachine.evm(42161), // arbitrum
-                gateway: address(intentGateway)
-            });
-            deployments[3] = Deployment({
-                chain: StateMachine.evm(8453), // base
-                gateway: address(intentGateway)
-            });
-            deployments[4] = Deployment({
-                chain: StateMachine.evm(56), // bsc
-                gateway: address(intentGateway)
-            });
-            deployments[5] = Deployment({
-                chain: StateMachine.evm(100), // gnosis
-                gateway: address(intentGateway)
-            });
-            deployments[6] = Deployment({
-                chain: StateMachine.evm(137), // polygon
-                gateway: address(intentGateway)
-            });
-            deployments[7] = Deployment({
-                chain: StateMachine.evm(420420419), // polkadot
-                gateway: address(intentGateway)
-            });
-            deployments[8] = Deployment({
-                chain: StateMachine.evm(1868), // soneium
-                gateway: address(intentGateway)
-            });
-        } else {
-            deployments = new Deployment[](2);
-            deployments[0] = Deployment({
-                chain: StateMachine.evm(97), // bsc testnet (chapel)
-                gateway: address(intentGateway)
-            });
-            deployments[1] = Deployment({
-                chain: StateMachine.evm(80002), // polygon amoy
-                gateway: address(intentGateway)
-            });
-        }
+        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}(admin);
 
-        intentGateway.initialize(
-            Params({
-                host: HOST_ADDRESS,
-                dispatcher: config.get("CALL_DISPATCHER").toAddress(),
-                solverSelection: config.get("7702").toBool(),
-                surplusShareBps: 5_000, // 50%
-                protocolFeeBps: 30, // 0.3%
-                priceOracle: address(priceOracle)
-            }),
-            deployments
+        bytes memory initData = abi.encodeCall(
+            IntentGatewayV2.initialize,
+            (
+                Params({
+                    host: HOST_ADDRESS,
+                    dispatcher: config.get("CALL_DISPATCHER").toAddress(),
+                    solverSelection: config.get("7702").toBool(),
+                    surplusShareBps: 5_000, // 50%
+                    protocolFeeBps: 30, // 0.3%
+                    priceOracle: priceOracle
+                })
+            )
         );
+        ERC1967Proxy proxy = new ERC1967Proxy{salt: salt}(address(implementation), initData);
+        IntentGatewayV2 intentGateway = IntentGatewayV2(payable(address(proxy)));
+
+        SolverAccount solverAccount = new SolverAccount{salt: salt}(address(intentGateway));
 
         vm.stopBroadcast();
+
+        console.log("IntentGateway implementation deployed at:", address(implementation));
+        console.log("IntentGateway proxy deployed at:", address(intentGateway));
+        console.log("SolverAccount deployed at:", address(solverAccount));
+
         config.set("INTENT_GATEWAY_V2", address(intentGateway));
         config.set("SOLVER_ACCOUNT", address(solverAccount));
         config.set("PRICE_ORACLE", address(priceOracle));
--- a/evm/script/DeployIntentGateway.s.sol
+++ b/evm/script/DeployIntentGateway.s.sol
@@ -24,7 +24,7 @@ contract DeployScript is BaseScript {
         // init data. The address depends on (impl address, salt, params, peer chain ids), all of
         // which are identical across chains, keeping the proxy address identical everywhere.
         address priceOracle = address(0);
-        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}();
+        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}(admin);
         bytes[] memory peerChains;
         if (config.get("is_mainnet").toBool()) {
             peerChains = new bytes[](9);
--- a/evm/script/DeployIntentGateway.s.sol
+++ b/evm/script/DeployIntentGateway.s.sol
@@ -10,6 +10,7 @@ import {BaseScript} from "./BaseScript.sol";
 import {CallDispatcher} from "../src/utils/CallDispatcher.sol";
 import {SolverAccount} from "../src/apps/intentsv2/SolverAccount.sol";
 import {VWAPOracle} from "../src/utils/VWAPOracle.sol";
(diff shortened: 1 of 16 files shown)


evm/script/DeployIntentGateway.s.sol
```diff
@@ -17,8 +18,15 @@ contract DeployScript is BaseScript {
...
-        console.log("IntentGateway deployed at:", address(intentGateway));
+        // Deploy the implementation and proxy both via CREATE2 with the same salt. The proxy
+        // is created with empty init data so its address depends only on (impl address, salt) —
+        // identical across chains. Initialization runs as a separate, deployer-gated call below.
+        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}(admin);
+        console.log("IntentGateway implementation deployed at:", address(implementation));
+
+        ERC1967Proxy proxy = new ERC1967Proxy{salt: salt}(address(implementation), "");
+        IntentGatewayV2 intentGateway = IntentGatewayV2(payable(address(proxy)));
+        console.log("IntentGateway proxy deployed at:", address(intentGateway));
```
seunlanlege: console after stop broadcast

evm/src/apps/IntentGatewayV2.sol
```diff
@@ -57,16 +58,24 @@ import {
...
     /**
-     * @dev Initializes the EIP-712 domain with name "IntentGateway" and version "2".
-     * Sets the initial admin who has one-time authority to call `setParams`.
-     * @param admin The address that will have permission to set initial parameters.
+     * @dev Deployer authorized to call `initialize` once on the proxy. An immutable (zero storage
+     * slots), so it must be byte-identical across chains or the deterministic proxy address diverges.
      */
-    constructor(address admin) EIP712("IntentGateway", "2") {
-        _admin = admin;
+    address private immutable _deployer;
```
seunlanlege: tbh lets just make admin immutable, maybe call it owner

evm/script/DeployIntentGateway.s.sol
```diff
@@ -17,8 +18,15 @@ contract DeployScript is BaseScript {
...
     function deploy() internal override {
-        IntentGatewayV2 intentGateway = new IntentGatewayV2{salt: salt}(admin);
-        console.log("IntentGateway deployed at:", address(intentGateway));
+        // Deploy the implementation and proxy both via CREATE2 with the same salt. The proxy
+        // is created with empty init data so its address depends only on (impl address, salt) —
+        // identical across chains. Initialization runs as a separate, deployer-gated call below.
+        IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}(admin);
+        console.log("IntentGateway implementation deployed at:", address(implementation));
+
+        ERC1967Proxy proxy = new ERC1967Proxy{salt: salt}(address(implementation), "");
```
seunlanlege: i think we can initialize here actually

evm/script/DeployIntentGateway.s.sol
seunlanlege: all intent gateways will have the same deployments array, so their proxy addresses will be the same

evm/script/DeployIntentGateway.s.sol
```diff
@@ -4,100 +4,52 @@ pragma solidity ^0.8.17;
...
+        // lets cross-chain peers resolve to `address(this)` rather than a stored registry.
+        address priceOracle = address(0);
+        // address priceOracle = deployPriceOracle(address(intentGateway));
+
         IntentGatewayV2 implementation = new IntentGatewayV2{salt: salt}(admin);
-        console.log("IntentGateway implementation deployed at:", address(implementation));
 
-        ERC1967Proxy proxy = new ERC1967Proxy{salt: salt}(address(implementation), "");
+        bytes memory initData = abi.encodeCall(
+            IntentGatewayV2.initialize,
```
seunlanlege, line 47: isn't this supposed to take deployments as well?

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -321,8 +316,7 @@ abstract contract IntentsBase is EIP712 {
      */
     function _instance(bytes calldata stateMachineId) internal view returns (address) {
         address gateway = _instances[keccak256(stateMachineId)];
-        if (gateway == address(0)) revert UnknownInstance();
-        return gateway;
+        return gateway == address(0) ? address(this) : gateway;
```
seunlanlege: nahhh please don't change this

evm/script/DeployIntentGateway.s.sol
seunlanlege: I just meant we can set the same deployments array for all intent gateways so they have the same address, like what was previously there

seunlanlege (approved):

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -359,6 +367,66 @@ impl RequestKind {
...
 			},
+			RequestKind::UpgradeContract { new_impl, init_data } => {
+				use alloy_primitives::{Address, Bytes};
+				// Mirrors the Solidity `abi.decode(body[1:], (address, bytes))` in
+				// `ExtrinsicIntents.onAccept`. `abi_encode_params` emits the two values as bare
+				// ABI parameters (no outer tuple wrapper), matching `abi.encode(newImpl, initData)`.
+				let payload = (Address::from_slice(&new_impl.0), Bytes::from(init_data.clone()));
+
+				let mut body = vec![IntentGatewayRequestKind::UpgradeContract as u8];
+				body.extend_from_slice(&payload.abi_encode_params());
```
Wizdave97, line 378:
```suggestion
				body.extend_from_slice(&payload.abi_encode());
```

Wizdave97 (approved):

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -359,6 +367,66 @@ impl RequestKind {
...
 			},
+			RequestKind::UpgradeContract { new_impl, init_data } => {
+				use alloy_primitives::{Address, Bytes};
+				// Mirrors the Solidity `abi.decode(body[1:], (address, bytes))` in
+				// `ExtrinsicIntents.onAccept`. `abi_encode_params` emits the two values as bare
+				// ABI parameters (no outer tuple wrapper), matching `abi.encode(newImpl, initData)`.
+				let payload = (Address::from_slice(&new_impl.0), Bytes::from(init_data.clone()));
+
+				let mut body = vec![IntentGatewayRequestKind::UpgradeContract as u8];
+				body.extend_from_slice(&payload.abi_encode());
```
Wizdave97:
```suggestion
				body.extend_from_slice(&payload.abi_encode_params());
```

Wizdave97:
@royvardhan 

🔴 2. Atomic CREATE2 proxy deployment will revert — initialize owner gate is incompatible with the deploy flow
initialize requires msg.sender == _owner (IntentGatewayV2.sol), where _owner is the immutable admin. The production script deploys atomically:

ERC1967Proxy proxy = new ERC1967Proxy{salt: salt}(address(implementation), initData); // initData = initialize(...)
(DeployIntentGateway.s.sol:60)

Under forge script --broadcast, salted creation is routed through the deterministic CREATE2 factory (0x4e59…4956C). During the proxy constructor, ERC1967Utils.upgradeToAndCall delegatecalls initialize, and delegatecall preserves msg.sender = the CREATE2 factory — not admin. So msg.sender != _owner ⇒ revert, and the whole deployment fails (unless ADMIN is literally set to the factory address, which would make the gate meaningless).

This is masked because the tests don't exercise the atomic path — _deployGatewayProxy uses empty init data and calls initialize separately from address(this) (= the owner), with a comment claiming it "mirrors production" (IntentGatewayV2Test.sol:99-106). Production is atomic; the comment is wrong and the real path is untested.

Fix (recommended): drop the msg.sender == _owner check entirely. With atomic init, the initData is bound into the CREATE2 initcode hash, so the canonical address can only be produced by deploying the exact correct params — the owner gate adds nothing and breaks the flow. (initializer still guarantees single-init.) Then add a test that deploys through new ERC1967Proxy{salt}(impl, initData) so the atomic path is actually covered. If you instead keep separate (non-atomic) init, the owner gate is fine and front-running is harmless (front-runner isn't the owner), but the script must be changed to empty initData + a separate initialize tx.

seunlanlege:
Yeah we can just drop the owner gating

evm/src/apps/IntentGatewayV2.sol
```diff
@@ -57,16 +58,15 @@ import {
...
     /**
-     * @dev Initializes the EIP-712 domain with name "IntentGateway" and version "2".
-     * Sets the initial admin who has one-time authority to call `setParams`.
-     * @param admin The address that will have permission to set initial parameters.
+     * @dev Initializes the EIP-712 domain with name "IntentGateway" and version "2",
+     * and locks this raw implementation so it can never be initialized directly.
      */
-    constructor(address admin) EIP712("IntentGateway", "2") {
-        _admin = admin;
+    constructor() EIP712("IntentGateway", "2") {
```
seunlanlege: We still need the admin to invoke pauses in future upgrades.

seunlanlege (dismissed):

PR #994, opened 2026-06-25
[evm,sdk]: fix reverting quote() on HyperApp and HyperFungibleToken

### LINKED ISSUE #987: [evm]: Fix reverting `quote` function on IHyperApp and HyperfungibleToken
## Root cause: `quote()` is `view`, but the host's fee oracle is now a Uniswap **V3** quoter that cannot run inside a `STATICCALL`

The revert is not a bug in your deployment config — it's a mutability mismatch baked into the contract, triggered by the recent migration of the mainnet fee-token swap oracle to deep-liquidity V3/V4 pools.

### The call chain

`quote(SendParams)` on your WHFT → [`[HyperApp.quote(DispatchPost)](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/HyperApp.sol:73)`](sdk/packages/core/contracts/apps/HyperApp.sol:73):

```solidity
function quote(DispatchPost memory request) public view returns (uint256) {
    address _uniswap = IDispatcher(_host).uniswapV2Router();
    ...
    return IUniswapV2Router02(_uniswap).getAmountsIn(request.fee, path)[0];
}
```

Two things collide here:

1. **`quote()` is declared `view`**, and the local interface declares [`[getAmountsIn(...) external pure](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/HyperApp.sol:34)`](sdk/packages/core/contracts/apps/HyperApp.sol:34). Because the compiler trusts that `pure` declaration, it emits a **`STATICCALL`** to the router.

2. On Ethereum mainnet your host's `uniswapV2Router()` no longer points at a real Uniswap V2 router. After [[PR #916 ("migrate mainnet fee-token swap wrappers to deep-liquidity v3/v4 pools")](https://github.com/polytope-labs/hyperbridge/pull/916)](https://github.com/polytope-labs/hyperbridge/pull/916), it points at [`[UniV3UniswapV2Wrapper](https://claude.ai/epitaxy/evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol:164)`](evm/src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol:164), whose `getAmountsIn` calls Uniswap's `QuoterV2.quoteExactOutputSingle`.

`QuoterV2` is **not** `view`. Per Uniswap's own note in [[IQuoterV2.sol](https://claude.ai/epitaxy/evm/node_modules/.pnpm/@uniswap+v3-periphery@1.4.4/node_modules/@uniswap/v3-periphery/contracts/interfaces/IQuoterV2.sol:8)](evm/node_modules/.pnpm/@uniswap+v3-periphery@1.4.4/node_modules/@uniswap/v3-periphery/contracts/interfaces/IQuoterV2.sol:8):

> These functions are not marked view because they rely on calling non-view functions and reverting to compute the result… they should not be called on-chain.

It works by executing a real `pool.swap()` (which does `SSTORE`s) and reverting to read the result back. Inside the `STATICCALL` forced by the `view` `quote()`, that first state write is illegal → the static context reverts → `getAmountsIn` fails → `quote()` reverts. Your SDK surfaces that as *"Unable to calculate nativeFee."* (that exact string is client-side — it isn't anywhere in the contracts).

This is independent of your params (`relayerFee`, `dest`, `to`, etc.). Any call to `quote()` against a V3/V4-wrapper host reverts.

### Why `send()` itself is not the broken part

`send()` is `payable`/non-`view`. When the host does the same fee conversion during an actual dispatch, it runs as a normal `CALL` (not `STATICCALL`), so the V3 quoter's swap-and-revert trick works. The practical breakage is that the **SDK can't price the transfer** — it calls `quote()` to size `msg.value`, and that read reverts before `send()` is ever attempted.

### Fix (requires an implementation upgrade)

Drop the `view`/`pure` mutability so the compiler emits a regular `CALL`, then have the SDK read it via `eth_call` (non-view functions simulate fine through `eth_call`, which is how every V3-quoter consumer does it):

- In [[HyperApp.sol](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/HyperApp.sol)](sdk/packages/core/contracts/apps/HyperApp.sol): remove `pure` from the `IUniswapV2Router02.getAmountsIn` declaration, and remove `view` from both `quote(DispatchPost)` and `quote(DispatchGet)`.
- Propagate: [`[WrappedHyperFungibleTokenUpgradeable.quote](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/WrappedHyperFungibleTokenUpgradeable.sol:249)`](sdk/packages/core/contracts/apps/WrappedHyperFungibleTokenUpgradeable.sol:249) (and the `HyperFungibleToken*` variants) can no longer be `view` — drop it there too.
- Your WHFT is a proxy (`0x6e35…`), so this ships as an implementation upgrade — no redeployment of the proxy/state needed.

### Immediate workarounds (no upgrade)

- **Pay in the fee token instead of native:** call `send()` with `msg.value == 0`. That takes the [`[dispatchWithFeeToken](https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/WrappedHyperFungibleTokenUpgradeable.sol:308)`](sdk/packages/core/contracts/apps/WrappedHyperFungibleTokenUpgradeable.sol:308) path, which never touches the native-fee quoter. The sender just needs to approve `relayerFee` of the host's fee token to the WHFT. This sidesteps `quote()` entirely.
- If you must pay native, you can query the host's `uniswapV2Router()` wrapper's `getAmountsIn` directly via `eth_call` (it's non-view, so call it as a transaction simulation, not `staticCall`) to get the native fee, then call `send{value: nativeFee + params.amount}(...)`.

Want me to draft the contract diff (removing the `view`/`pure` modifiers across HyperApp + the HFT/WHFT `quote` overrides) and check the SDK's `quote` call site uses `eth_call` rather than a static read?
<h3>Solidity picks STATICCALL based on the <em>callee's</em> declared mutability, not the caller's</h3>
<p>The <code>STATICCALL</code> opcode is emitted whenever you call a function that is <strong>declared</strong> <code>view</code> or <code>pure</code> — regardless of whether the calling function is <code>view</code>. In <a href="https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/HyperApp.sol:34">HyperApp.sol:34</a> the local interface declares:</p>
<pre><code class="language-solidity">function getAmountsIn(uint256, address[] calldata) external pure returns (uint256[] memory);
</code></pre>
<p>So even if you make <code>quote()</code> non-view, this call still compiles to a <code>STATICCALL</code> because <code>getAmountsIn</code> is declared <code>pure</code> here. The V3/V4 wrapper would still revert.</p>
<h3>To actually fix it on-chain you must change two things</h3>
<ol>
<li>Remove <code>view</code> from <a href="https://claude.ai/epitaxy/sdk/packages/core/contracts/apps/HyperApp.sol:73"><code>quote(DispatchPost)</code> / <code>quote(DispatchGet)</code></a> (and the propagated <code>quote()</code> overrides in HFT/WHFT).</li>
<li>Remove <code>pure</code> from the <code>IUniswapV2Router02.getAmountsIn</code> declaration (make it plain <code>external ... returns</code>, i.e. nonpayable) — this is what flips the opcode from <code>STATICCALL</code> to <code>CALL</code>.</li>
</ol>
<h3>Then both router kinds work</h3>

Router | getAmountsIn actual mutability | Under a CALL
-- | -- | --
Vanilla Uniswap V2 | view | ✅ works — view runs fine under CALL, just doesn't mutate
V3/V4 quoter wrapper | non-view (QuoterV2 swap-and-revert) | ✅ works — CALL permits the transient state writes

<p>A <code>CALL</code> to a view-implemented function is always valid; the EVM only enforces the read-only constraint when <em>you</em> invoke via <code>STATICCALL</code>. So declaring the interface method nonpayable doesn't hurt vanilla V2 — it just stops forcing the static context that the V3/V4 path can't survive.</p>

### Trade-offs of the on-chain route (vs. the SDK fix already applied)

- `quote()` would no longer be callable from another contract's `view` function, and tooling can't treat it as a free read — though off-chain `eth_call`/`readContract` still works on non-view functions.
- It's an implementation upgrade across HyperApp + every app that inherits it (HFT, WHFT, etc.), and it changes the published ABI's `stateMutability`.

That's exactly why the SDK-side fix is the lighter lever: it sidesteps the `STATICCALL` entirely by calling `getAmountsIn` top-level, with zero contract changes. But to answer your direct question: yes, vanilla V2 survives removing `view` — you just have to remove the `pure` on the interface method too, or the change has no effect.

royvardhan:
`HyperApp.quote()` was `view` and called `getAmountsIn` through an interface that declared it `pure`, so the compiler emitted a `STATICCALL`. On mainnet the host's fee oracle is now a Uniswap V3 quoter wrapper whose `getAmountsIn` prices by executing a swap and reverting to read the result — a state write that a `STATICCALL` rejects, so `quote()` reverts.

Dropped `view`/`pure` so the call runs as a `CALL`, and propagated the change to the `HyperFungibleToken`/`WrappedHyperFungibleToken` `quote` overrides. The SDK now reads `quote` via `simulateContract` (eth_call) since it is no longer a view function. Added a mainnet-fork test that prices a send through the real `UniV3UniswapV2Wrapper`.

Ships as an implementation upgrade behind the existing WHFT proxy; no proxy/state redeployment.

Closes #987

### DIFF
--- a/evm/tests/foundry/HyperFungibleTokenTest.sol
+++ b/evm/tests/foundry/HyperFungibleTokenTest.sol
@@ -16,6 +16,9 @@ import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol
 
 import {HyperFungibleToken} from "@hyperbridge/core/apps/HyperFungibleToken.sol";
 import {WrappedHyperFungibleToken} from "@hyperbridge/core/apps/WrappedHyperFungibleToken.sol";
+import {MainnetForkBaseTest} from "./MainnetForkBaseTest.sol";
+import {UniV3UniswapV2Wrapper} from "../../src/utils/uniswapv2/UniV3UniswapV2Wrapper.sol";
+import {HostParams} from "../../src/core/EvmHost.sol";
 
 // Concrete HyperFungibleToken for testing
 contract TestHFT is HyperFungibleToken {
@@ -722,4 +725,49 @@ contract WrappedHyperFungibleTokenTest is BaseTest {
     }
 
     receive() external payable {}
+}
+
+/// @notice Forks mainnet and prices a cross-chain send through `HyperApp.quote()` against a host
+/// whose fee oracle is the real `UniV3UniswapV2Wrapper`.
+///
+/// The V3 quoter prices `getAmountsIn` by executing a swap and reverting to read the result, so it
+/// writes state. `quote()` must therefore reach it via a `CALL`; a `view` `quote()` emits a
+/// `STATICCALL`, which rejects the state write and reverts before the fee can be priced.
+contract HyperFungibleTokenQuoteForkTest is MainnetForkBaseTest {
+    function testQuotePricesNativeFeeThroughV3Wrapper() public {
+        UniV3UniswapV2Wrapper wrapper = new UniV3UniswapV2Wrapper(address(this));
+        wrapper.init(
+            UniV3UniswapV2Wrapper.Params({
+                WETH: _uniswapV2Router.WETH(),
+                swapRouter: 0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45,
+                quoter: 0x61fFE014bA17989E743c5F6cB21bF9697530B21e,
+                maxFee: 500
+            })
+        );
+
+        // Point the host's fee oracle at the V3 wrapper, as it is configured on mainnet.
+        HostParams memory params = host.hostParams();
+        params.uniswapV2 = address(wrapper);
+        vm.prank(address(manager));
+        host.updateHostParams(params);
+
+        bytes memory destChain = StateMachine.evm(42161); // Arbitrum
+        bytes memory remoteContract = abi.encodePacked(address(0xBEEF));
+
+        TestHFT hft = new TestHFT();
+        hft.configure(HyperFungibleToken.ConfigOptions({host: address(host), dispatcher: address(dispatcher)}));
+        hft.addChain(destChain, remoteContract);
+
+        HyperFungibleToken.SendParams memory sendParams = HyperFungibleToken.SendParams({
+            dest: destChain,
+            to: remoteContract,
+            amount: 1 ether,
+            timeout: 3600,
+            relayerFee: 10 * 1e18, // 10 DAI
+            data: ""
+        });
+
+        uint256 nativeFee = hft.quote(sendParams);
+        assertGt(nativeFee, 0, "quote should price the relayer fee in native through the V3 wrapper");
+    }
 }
--- a/sdk/packages/core/contracts/apps/HyperApp.sol
+++ b/sdk/packages/core/contracts/apps/HyperApp.sol
@@ -31,7 +31,7 @@ import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol
 interface IUniswapV2Router02 {
     function WETH() external pure returns (address);
 
-    function getAmountsIn(uint256, address[] calldata) external pure returns (uint256[] memory);
+    function getAmountsIn(uint256, address[] calldata) external returns (uint256[] memory);
 }
 
 /**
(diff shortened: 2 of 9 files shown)


Wizdave97 (approved):
Bump the core package's patch version

PR #1000, opened 2026-06-27
[sdk]: skip timeout inference for terminal post requests

ddboy19912:
Fixes SDK status snapshots so completed post requests no longer get timeout events appended.

The change skips timeout inference when the indexed request statuses already contain a terminal status (`DESTINATION` or `TIMED_OUT`). This prevents stale or missing destination receipt lookups from causing completed Bifrost messages to appear timed out.

### DIFF
--- a/sdk/packages/sdk/src/client/PostRequestClient.ts
+++ b/sdk/packages/sdk/src/client/PostRequestClient.ts
@@ -264,12 +264,7 @@ export class PostRequestClient {
 	 * accompanying timeout-proof calldata.
 	 */
 	async addTimeoutFinalityEvents(request: PostRequestWithStatus): Promise<PostRequestWithStatus> {
-		const destChain = this.ctx.config.dest
-		const hyperbridge = this.ctx.config.hyperbridge
 		const events: RequestStatusWithMetadata[] = []
-		const commitment = postRequestCommitment(request).commitment
-		const receipt = await destChain.queryRequestReceipt(commitment)
-		const destTimestamp = await destChain.timestamp()
 
 		const commit = (req: PostRequestWithStatus) => {
 			this.logger.trace(`Added ${events.length} timeout events`, events)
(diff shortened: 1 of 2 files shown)


sdk/packages/sdk/src/client/PostRequestClient.ts
```diff
@@ -278,15 +273,26 @@ export class PostRequestClient {
...
+
+		const destChain = this.ctx.config.dest
+		const hyperbridge = this.ctx.config.hyperbridge
+		const commitment = postRequestCommitment(request).commitment
+		const receipt = await destChain.queryRequestReceipt(commitment)
+		const destTimestamp = await destChain.timestamp()
+
 		if (receipt || request.timeoutTimestamp > destTimestamp) return commit(request)
 
-		const is_finished = request.statuses.find((item) => item.status === RequestStatus.DESTINATION)
```
Wizdave97, line 283: Why was this check not working?

ddboy19912: The old logic only prevented adding `PENDING_TIMEOUT`. It keeps going into the rest of timeout inference and could still add `DESTINATION_FINALIZED_TIMEOUT`

ddboy19912: Basically this
```
const receipt = await destChain.queryRequestReceipt(commitment)
const destTimestamp = await destChain.timestamp()

if (receipt || request.timeoutTimestamp > destTimestamp) return

const is_finished = request.statuses.find((item) => item.status === RequestStatus.DESTINATION)

if (!is_finished) {
  events.push(PENDING_TIMEOUT)
}

// code continues...
const delivered = request.statuses.find(HYPERBRIDGE_DELIVERED)

if (delivered) {
  const destFinalized = await queryStateMachineUpdateByTimestamp(...)
  events.push(DESTINATION_FINALIZED_TIMEOUT)
}
```

Wizdave97 (approved):

PR #1008, opened 2026-06-30
[indexer]: index yield vaults to track LP yields

### LINKED ISSUE #1001: [indexer] Index supported yield vaults to track LP yields
We should index the yield vaults for data we can use to track yields earned by LPs running simplex.

royvardhan:
Closes #1001

Indexes the supported ERC-4626 yield vaults so LP yields running simplex can be tracked:

- `Deposit` / `Withdraw` events (redeem emits Withdraw) per vault → `VaultLedgerEvent` audit trail and a running `VaultLpPosition` (net principal per LP).
- A daily block-handler snapshot prices each LP's live share balance into assets via `convertToAssets(balanceOf(lp))` → `VaultPositionSnapshot` (with `yieldEarned`), plus a vault-level `VaultSnapshot` (totalAssets / totalShares / assetsPerShare).

Datasources are generated per configured vault from the existing `yieldVaults` config; snapshots dedupe per UTC day. Generic over any ERC-4626 vault, so it also covers our own StreamingYieldVault.


sdk/packages/indexer/src/handlers/events/yieldVault/deposit.event.handler.ts
```diff
@@ -0,0 +1,32 @@
...
+ * `owner` is the share recipient (the LP), `sender` the caller.
+ */
+export const handleVaultDepositEvent = wrap(async (event: DepositLog): Promise<void> => {
+	if (!event.args) return
+
+	const { args, address, blockNumber, blockHash, transactionHash, logIndex } = event
+	const chain = getHostStateMachine(chainId)
+	const timestamp = await getBlockTimestamp(blockHash, chain)
+
+	await YieldVaultService.recordLedger({
```
seunlanlege, line 19: We probably want to filter this by our actual LPs

royvardhan: Yeah, working on it. I was thinking that the vaults would be only the ones which we deployed or got deployed for simplex usage only.

royvardhan: Can filter with a check, if the LP is a solver account.

royvardhan: Done

seunlanlege (approved):

Wizdave97:
@royvardhan recordLedger drops the event whenever lp isn't already a LiquidityProvider, but those rows are created only by the Hyperbridge phantom handler (a different chain / subql node), so there's no ordering guarantee against these EVM vault events. If a solver deposits before its first phantom bid is indexed, the Deposit is skipped and never backfilled — totalAssetsDeposited is left too low, and since the daily snapshot prices the live balanceOf (which already includes those shares), yieldEarned = assetValue - netPrincipal ends up overstated by the dropped principal, for exactly the LPs this is meant to track.

As a backup when the LiquidityProvider lookup misses, check whether lp is delegated to the our solver accounts before dropping the event — treat a delegated account as ours, so its vault activity is still tracked even before (or without) a LiquidityProvider row.

Wizdave97 (approved):

PR #1022, opened 2026-07-07
[tesseract]: Deliver EVM GET responses from the relayer

dharjeezy:
This PR adds delivery of EVM GET responses on the outbound messaging task, so once Hyperbridge commits a `GetResponse` it is relayed back to the origin EVM chain. The relayer only delivers a response whose origin GetRequest carries a fee that covers the delivery gas, and the EVM host pays that  fee to the relayer inline during dispatch.


tesseract/messaging/messaging/src/events.rs
```diff
@@ -251,9 +252,102 @@ pub async fn translate_events_to_messages(
...
+			coprocessor,
+			&client_map,
+			config.deliver_failed.unwrap_or_default(),
+			consensus_prelude,
+		)
+		.await?;
+
+		unprofitable.extend(profitability.retriable_messages);
+
+		for (msg, keep) in response_messages.into_iter().zip(profitability.queries) {
```
Wizdave97: where's the chunking step like we do for post requests? Messages are meant to be batched

tesseract/messaging/messaging/src/events.rs
```diff
@@ -251,9 +252,102 @@ pub async fn translate_events_to_messages(
 		}
 	}
 
+	// GetResponses coming back from Hyperbridge to an EVM source. Their values
+	// are already carried in the events, so we only need the membership proof
+	// before running them through the same profitability gate as requests.
+	let (response_messages, response_queries) = build_get_response_messages(
```
Wizdave97: Only run this get response block if the source is the coprocessor

tesseract/messaging/messaging/src/events.rs
```diff
@@ -278,6 +372,12 @@ pub fn filter_events(
 			}
 			post.dest == counterparty && is_allowed_module(config, &post.from)
 		},
+		// A GetResponse is delivered back to the chain that made the request, and
+		// only EVM sources can verify the mmr membership proof it rides on.
+		IsmpEvent::GetResponse(res) =>
```
Wizdave97, line 440: The coprocessor check recommended above should be sufficient
```suggestion
		IsmpEvent::GetResponse(res) => true
```

Wizdave97 (approved):

PR #1031, opened 2026-07-09
[simplex] Inflight price curve updates via admin server

### LINKED ISSUE #1023: [simplex] Inflight price curve changes
We want to increase the speed of updating price curves on simplex. Restarting the application to modify price curves is very slow especially when it's running in docker.
Simplex should provide a simple RPC endpoint that can be used to update the price curves it holds in memory.

This should be accompanied with a locally served UI where the can change these prices on the fly.

royvardhan:
Adds an opt-in admin server (`--admin-port <[host:]port>`) serving a JSON API and a locally hosted UI for updating FX bid/ask price curves in memory, without restarting the filler.

- Curve mutation lives on `FillerPricePolicy.replacePoints()`. The server holds the same policy instances the strategies price with, so `FXFiller` is untouched and updates apply on the next order evaluation.
- `GET /api/strategies` lists FX strategies (labelled by their exotic token symbol); `PUT /api/strategies/:index/curves` replaces bid/ask points. Updates are all-or-nothing and go through the same `FillerPricePolicy` validation the TOML curves pass at boot.
- Disabled sides (one-sided LP) and venue-priced strategies are not editable; changes are lost on restart.
- Unauthenticated; binds to `127.0.0.1` by default.

Closes #1023


sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -584,6 +600,43 @@ program
 				}
 			}
 
+			// Start optional admin server for inflight price curve updates
```
Wizdave97: Should not be optional

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -584,6 +600,43 @@ program
...
+			let adminServer: AdminServer | undefined
+			if (options.adminPort) {
```
Wizdave97, line 609: Use a default port if none is supplied

Wizdave97 (approved):

PR #1036, opened 2026-07-13
[indexer]: index intent gateway volumes with FX pricing from phantom snapshots

### LINKED ISSUE #1032: [Indexer]: Index Volume for HyperFX transactions

royvardhan:
Closes #1032

Indexes intent gateway volumes separately: cumulative raw token amounts per chain (`IntentGatewayTokenVolume`) and a USD rollup per chain (`CumulativeIntentGatewayVolumeUSD`), recorded on order placed (inputs) and filled (outputs).

USD conversion prices USDC/USDT via the indexer's USDC `TokenPrice` row (one rate for both) and FX tokens like cNGN via the latest `PhantomOrderPriceSnapshot` exchange rate, falling back to the existing CoinGecko path for other tokens. Raw amounts are always retained so USD values can also be computed off-indexer.

### DIFF
--- a/sdk/packages/indexer/src/configs/schema.graphql
+++ b/sdk/packages/indexer/src/configs/schema.graphql
@@ -2664,3 +2664,80 @@ type VaultSnapshot @entity @compositeIndexes(fields: [["chain", "vault"]]) {
 	"""
 	snapshotTime: Date! @index
 }
+
+"""
+Cumulative raw token volume through IntentGatewayV3, per chain, token and volume type.
+Raw amounts are kept so USD values can also be computed off-indexer with any price source.
+"""
+type IntentGatewayTokenVolume @entity {
+	"""
+	Composite identifier: {chain}-{tokenAddress}-{volumeType}
+	"""
+	id: ID!
+
+	"""
+	Host state machine that emitted the event
+	"""
+	chain: String! @index
+
+	"""
+	EVM address of the token, as a lowercase 0x-prefixed hex string
+	"""
+	tokenAddress: String! @index
+
+	"""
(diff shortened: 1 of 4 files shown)


Wizdave97:
@royvardhan  
1. getTokenMetadata failure fallback permanently poisons data (highest priority).
On any transient RPC failure it returns {symbol: "UNKNOWN", decimals: 18} instead of throwing. Consequences:

The wrong symbol/decimals get baked into IntentGatewayTokenVolume on first sight and are never corrected (the update path doesn't refresh them).
Pricing continues with decimals: 18: a 6-decimal token still resolvable via an address-based snapshot lookup would be under-priced by 1e12 in the cumulative counter — unrecoverable without a reindex.
The existing sibling getTokenValuesUSD (intentGatewayV3.service.ts:328-345) deliberately lets this error propagate, so wrap() rethrows and SubQuery retries the block. Recommend doing the same here, or at minimum skipping the token entirely (both raw and USD) instead of recording under UNKNOWN/18.

Wizdave97:
recordOrderVolume re-fetches symbol()/decimals() for every token on every event, even though (a) getOrCreateOrder/recordFill already fetched the same metadata moments earlier in the same handler, and (b) on the update path the IntentGatewayTokenVolume row already stores them. Reusing the stored values would remove two RPC round-trips per token per event.
Tokens are processed sequentially; the existing getTokenValuesUSD uses Promise.all. Minor, but the FX path adds two DB queries + up to two RPC calls per token, so it adds up on busy chains.

Wizdave97 (approved):

PR #1040, opened 2026-07-14
[runtime]: Scope consensus rewards to the unpaid height span

dharjeezy:
Consensus rewards were sized from the gap between the latest and previous state machine heights, which double counts whenever a height is rolled back and later resubmitted. This tracks the highest rewarded height per state machine in `LastRewardedHeight` and pays only for the span above it, keeping the previous height as the baseline until the first reward is recorded so existing chains carry through the upgrade unchanged.


modules/pallets/consensus-incentives/src/impls.rs
Wizdave97, line 53: We'll need a zero check for reward before transferring

Wizdave97 (approved):

PR #1021, opened 2026-07-07
[evm, sdk, simplex]: add SimplexPaymaster

royvardhan:
Revives #768 on top of current main.

**Contract**: `SimplexPaymaster.sol` — onchain ERC-4337 v0.8 paymaster accepting USDC/USDT for gas via Chainlink oracles, with `PERMIT` and `APPROVE` modes. Deployed behind an ERC1967Proxy with no privileged key: upgrades, pricing parameters, the token registry, and treasury withdrawals are all governed exclusively through Hyperbridge — `onAccept` authenticates requests from Hyperbridge via the host, mirroring IntentGatewayV2's pattern. Includes deploy script and Foundry tests for the pricing math, oracle safety, paymasterData parsing, and the governance actions.

**Pallet**: `pallet-intents-coprocessor` gains the runtime side that emits those governance actions. A `Paymasters` registry maps each chain to its deployed proxy, and six `GovernanceOrigin`-gated extrinsics (`add_paymaster_deployment`, `upgrade_paymaster`, `update_paymaster_params`, `register_paymaster_token`, `deactivate_paymaster_token`, `withdraw_paymaster_assets`) build the authenticated ISMP request and dispatch it to the paymaster, reusing the existing gateway dispatch path. A `SimplexPaymasterRequestKind` discriminator plus `sol!` types mirror the contract's `RequestKind`/`Params` ABI, with round-trip parity tests.

**SDK**: `buildPaymasterAndData()` selects Circle Paymaster (USDC permit) when configured and the filler holds USDC, falling back to SimplexPaymaster (USDC or USDT, permit or capped approval). `UserOpSender` and delegation are provider-aware; delegation forces approve mode so its fixed gas limits stay valid.

**Deployments**: the live BSC (`0xCcE8B5f5C453e1Ff1acd227585d9e680c012462c`) and Base (`0x18654D0319ffDC18c200c90C1A666AcbDf0E3762`) contracts predate the proxy change and will be redeployed behind proxies, alongside new deployments on Ethereum, Arbitrum, and Polygon — oracle and token config for all five chains is wired in `config.mainnet.toml`.

Verified on BSC mainnet: paymaster-sponsored re-delegation (`0x1f15f4b3bb1839d27a3cbffe43a726ee18161458b5274c13f13998ebd2598712`) and order fill (`0xc4dbe8a1e66886cc624f2415d314d565789005a0bc61d8118ac2a702965a0bfd`), gas paid in USDT.

PR #1041, opened 2026-07-15
[evm, simplex]: recycle SimplexPaymaster fees into the EntryPoint deposit

royvardhan:
Follow-up to #1021. The paymaster's accrued stablecoins are now recycled into its EntryPoint deposit fully onchain.

- `swapAndDeposit(token, amountIn)` swaps accrued fees to native through a governance-set V2-style router (`Params.uniswapV2Router`) and deposits the proceeds into the EntryPoint. Treasury-gated: a permissionless version could be invoked from UserOp calldata mid-bundle and swap away other ops' pending prefunds, breaking their postOp refunds.
- The host's `uniswapV2Router()` is not reusable here because the wrapper routers only implement `swapETHForExactTokens`. BSC is configured with the PancakeSwap V2 router; Base stays disabled until a token-selling router is chosen there.
- `pallet-intents-coprocessor` mirrors the two new `Params` fields.
- New `simplex paymaster-keeper` subcommand: a minimal keeper that periodically calls `swapAndDeposit` whenever a registered token's accrued balance exceeds a threshold.


Wizdave97 (approved):

Wizdave97:
Let's implement the required swap tokens function on our uniswap router in this PR

royvardhan:
**On keeping Chainlink for pricing (and Uniswap for execution only)**

We considered replacing the Chainlink oracles with the wrappers' `getAmounts` helpers now that Uniswap is available. Keeping Chainlink, for two reasons:

- **Validation pricing (`_fetchDetails`) can't use them anyway.** It's `view`, but the V3/V4 quoters behind `getAmountsOut` are non-`view` (they simulate a swap and revert to return), so it wouldn't compile, and a heavy quoter call in the validation phase is exactly what ERC-7562 bundler rules penalize.
- **In `swapAndDeposit`, the Chainlink `amountOutMin` is an independent bound.** Deriving the minimum from the same pool we're trading against is circular, a sandwich moves the quote and the execution together, so the "minimum" protects nothing. The oracle is the only reference the swap's counterparty can't influence.

At current sizes the swap-side sandwich risk is negligible (dollar-scale swaps, loss capped to the slippage window), but the oracle bound is a zero-cost guard that also caps the loss if balances ever accrue larger between keeper runs.

Net: Chainlink prices, Uniswap executes.

PR #1044, opened 2026-07-15
[runtime]: Bind validator addresses to their public keys in the Tendermint client

dharjeezy:
Validator addresses are now recomputed from the public key while decoding a proof and rejected when they do not match, so the address a proof carries can no longer diverge from the key that the signed validator set hash actually commits to. The stored next validator  set is only replaced on a signalled rotation and only once the supplied list hashes to the new next_validators_hash, and any set that repeats an address is turned away. The Heimdall address derivation moved into tendermint-primitives so the relayer and the runtime share one implementation rather than keeping their own copies.

### DIFF
--- a/modules/consensus/tendermint/verifier/src/verifier.rs
+++ b/modules/consensus/tendermint/verifier/src/verifier.rs
@@ -1,5 +1,5 @@
 use core::time::Duration;
-use prost::alloc::{format, string::ToString};
+use prost::alloc::{collections::BTreeSet, format, string::ToString};
 
 use cometbft::{block::Height, chain::Id, trust_threshold::TrustThresholdFraction, Hash, Time};
 use cometbft_light_client_verifier::{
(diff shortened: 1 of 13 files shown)


modules/consensus/tendermint/verifier/src/verifier.rs
```diff
@@ -214,6 +214,22 @@ fn convert_timestamp(timestamp: u64) -> Result<Time, VerificationError> {
 		.map_err(|e| VerificationError::Invalid(e.to_string()))
 }
 
+/// Rejects a validator set that reuses the same address twice. Each address is a
+/// hash of its public key, so a duplicate signals a set that was not built honestly.
+fn ensure_unique_addresses(
+	validators: &[cometbft::validator::Info],
+) -> Result<(), VerificationError> {
+	let mut seen = BTreeSet::new();
```
seunlanlege: seems to me that you can collect validators into a BtreeSet and check if they both have different lengths

Wizdave97 (approved):

PR #1047, opened 2026-07-16
[indexer]: price intent order USD values on-chain instead of via CoinGecko

royvardhan:
Removes CoinGecko from the intent order pricing path: USDC/USDT are pinned to $1 and FX tokens are priced from the latest PhantomOrderPriceSnapshot, so order inputUSD, points, user volume stats and the intent USD rollups no longer depend on any external API.

### DIFF
--- a/sdk/packages/indexer/src/addresses/fx-tokens.addresses.ts
+++ b/sdk/packages/indexer/src/addresses/fx-tokens.addresses.ts
@@ -1,0 +1,6 @@
+// Phantom order price snapshots reference token addresses on Base (EVM-8453),
+// so FX tokens are resolved to their Base representation before snapshot lookups.
+export const BASE_CNGN = {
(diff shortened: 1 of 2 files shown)


royvardhan:
@Wizdave97 What do we do with the evm host handlers that try price indexing?

Wizdave97:
> @Wizdave97 What do we do with the evm host handlers that try price indexing?
Let's leave them unchanged for now

Wizdave97:
We need to resolve USDc/usdt to the USDc address on base, same for cNGN. 

Price snapshots only references USDc and cNGN tokens on base

royvardhan:
> We need to resolve USDc/usdt to the USDc address on base, same for cNGN.
> 
> Price snapshots only references USDc and cNGN tokens on base

Done

Wizdave97 (approved):

PR #1048, opened 2026-07-16
[sdk]: buff same chain and cross chain fees

ddboy19912:
## Summary

- Double the automatically estimated solver fee for same-chain Intent Gateway V2 swaps.
- Keep the shared cross-chain redemption POST-request estimate at `400_000` gas, so Simplex evaluates orders against the baseline execution cost.
- Add a `600_000`-gas cross-chain uplift only when the SDK assigns a user's `order.fees`, converted to the source-chain fee token.

## Rationale

Simplex uses the shared fill-order estimate to evaluate solver profitability. Applying the additional cross-chain gas in that shared estimate would increase both the user fee and Simplex’s cost basis, cancelling out the intended extra solver incentive.

Applying the uplift only during SDK order placement preserves the additional incentive without affecting Simplex’s evaluation.

### DIFF
--- a/sdk/packages/sdk/package.json
+++ b/sdk/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@hyperbridge/sdk",
-	"version": "2.4.1",
+	"version": "2.4.2",
 	"description": "The hyperclient SDK provides utilities for querying proofs and statuses for cross-chain requests from HyperBridge.",
 	"type": "module",
 	"types": "./dist/node/index.d.ts",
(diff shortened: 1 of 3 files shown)


seunlanlege (approved):

royvardhan:
The estimate function is used by simplex as well. So bumping here would bump the simplex's order evaluation as well. Meaning the bump will cancel out.

ddboy19912:
> The estimate function is used by simplex as well. So bumping here would bump the simplex's order evaluation as well. Meaning the bump will cancel out.

Please check again

royvardhan (approved):

PR #1050, opened 2026-07-16
[runtime]: Replace state commitment pointer set with a per-chain FIFO queue

Wizdave97:
Retaining a deep history of state commitments was not affordable under the old design. `KnownStateMachineHeights` held a sorted `BoundedBTreeSet` of every retained height per chain, and `insert_bounded_state_commitment` mutated it on every state machine update purely to answer one question: which height is the oldest? That made the per-insert cost scale with the retention cap. At the 192k depth we wanted for BSC the set would encode to ~1.5MB, so each update decoded and re-encoded it — ~3MB of I/O and PoV weight per insert, for an 8-byte answer.

Replace it with a FIFO queue keyed by a monotonic insertion index, plus a `{head, tail}` cursor per chain. Insertion order equals height order — the consensus handler sorts each batch ascending and skips anything at or below `previous_latest_height` — so the head is the oldest height and eviction is a lookup rather than a scan. Each insert now touches a handful of small storage items regardless of the cap, so retention depth is decoupled from per-insert cost. The `tail - head <= cap` bound holds independently of the ordering assumption, so a violation could only misorder eviction, never leave storage unbounded.

Caps become per-chain via `StateMachineCommitmentCap`, since a chain's useful retention is a wall-clock window divided by its block time; one global constant cannot serve both a 450ms chain and a 2s one. Chains without an override keep the existing 1024 default. Evictions per insert are capped so that *lowering* a cap drains gradually instead of doing unbounded work in a single call.

Seed BSC and Polygon at six hours of blocks via `SeedCommitmentCaps`, wired into nexus only. The migration derives each cap from a single retention-window constant and the chain's block time rather than hardcoding a number, and guards each write with `contains_key` so re-running it never clobbers a value governance has since changed.

One accepted wart, pinned by test: vetoing the *latest* height rolls the latest pointer back and re-opens that height for honest resubmission, which gives it a second queue entry. The stale twin then evicts the live commitment one insertion early and burns one queue slot. Both are negligible against the configured caps; making it exact would require a height -> index map on the insert path, reintroducing the cost this change removes.


modules/pallets/ismp/src/migrations.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Seeds [`StateMachineCommitmentCap`] for fast-finality chains. Each cap is
+/// only written if the chain has no cap configured yet, so re-running this
+/// migration never clobbers a value set by governance in the meantime.
+pub struct SeedCommitmentCaps<T>(PhantomData<T>);
+
+impl<T: Config> OnRuntimeUpgrade for SeedCommitmentCaps<T> {
+	fn on_runtime_upgrade() -> Weight {
+		let mut writes = 0u64;
+		for (state_id, consensus_state_id, cap) in COMMITMENT_CAPS {
```
dharjeezy, line 55: we don't neccessarily need to seed the queue too?

dharjeezy (approved):

modules/pallets/ismp/src/migrations.rs
Wizdave97: I don't want us to have to do an extra proposal after this to increase the caps for bsc and polygon

seunlanlege (approved):

PR #1051, opened 2026-07-17
[simplex]: setup wizards (terminal + local web UI) and operator dashboard plumbing

### LINKED ISSUE #1038: Simplex Local UI (Setup Wizard + Operator Dashboard)
## 1. Architecture

Simplex is a Node CLI ([[simplex.ts](https://claude.ai/epitaxy/sdk/packages/simplex/src/bin/simplex.ts)](sdk/packages/simplex/src/bin/simplex.ts), commander-based). The UI is a **locally-served web app embedded in the simplex binary** — no separate deployment:

- **Backend**: a small HTTP + WebSocket server (Hono or Fastify) inside the simplex process, bound to `127.0.0.1` (configurable `--ui-port`, default e.g. `4200`). It exposes a JSON API over the existing services (`ChainClientManager`, `ContractInteractionService`, `MetricsService`, `BidStorageService`, `IntentFiller`).
- **Frontend**: a static SPA (React + Vite) bundled into `dist/ui/` at build time and served by that server. No external assets, works fully offline.
- **Two modes, one server**:
  - **Init mode** — no valid config found → serve the setup wizard.
  - **Operator mode** — filler running → serve the dashboard.
- **Access model**: loopback-only binding is the boundary. The server refuses to bind to non-loopback addresses; anyone with local access to the machine has access to the UI.

### CLI behavior changes

- `simplex` (no args) and `simplex run` without `-c`: look for a config at a default path (`./simplex.toml`, then `$SIMPLEX_HOME/config.toml`). If none exists → **automatically start the init UI**, print the URL, and open the browser. When the wizard completes, it writes the config file (mode `0600`) and the same process transitions straight into `run` using it — no restart needed.
- `simplex run -c config.toml` keeps working headless; add `--ui [port]` / `--no-ui` flags to control the dashboard.
- Keep `-p/--port` Prometheus metrics untouched and independent of the UI.

## 2. Setup wizard (init mode)

A step-by-step flow that produces a complete `FillerTomlConfig`. Every step validates live before allowing "next".

**Step 1 — Signer**
- EVM signer as a tagged choice, matching `SignerConfig`: **private key** (primary path), with **MPC Vault** and **Turnkey** available as advanced tabs (config already supports all three).
- Private key input is masked; derive and display the EVM address for confirmation.

**Step 2 — Substrate key (generated)**
- Generate a mnemonic + sr25519 keypair via `@polkadot/util-crypto` (`mnemonicGenerate`), write it into `simplex.substratePrivateKey`.
- Show the derived Hyperbridge SS58 address prominently with a copy button and a warning: **"Fund this account with BRIDGE tokens — bid submission pays fees from it."**
- Live-check the BRIDGE balance over `hyperbridgeWsUrl` and show funded/not-funded status; allow proceeding anyway.
- Hyperbridge WS URL: pre-filled with the public endpoint (`wss://nexus.ibp.network`), editable.

**Step 3 — Chains & endpoints (Alchemy-first, any provider supported)**
- **Alchemy API key** input at the top (optional but recommended, validated live). Entering it auto-fills `rpcUrls[0]` and `bundlerUrl` for every selected chain Alchemy supports — Alchemy serves bundler RPC on the same endpoint (the codebase already relies on this in [[DelegationService.test.ts:43](https://claude.ai/epitaxy/sdk/packages/simplex/src/tests/services/DelegationService.test.ts:43)](sdk/packages/simplex/src/tests/services/DelegationService.test.ts:43)).
- Chain toggle cards for all SDK-supported chains (Ethereum, BSC, Arbitrum, Base, Polygon, Unichain, …). Each selected chain shows its **RPC URL and bundler URL fields**:
  - Pre-filled from the Alchemy key where available, visually tagged "via Alchemy" — zero typing in the happy path.
  - Freely editable to any provider (e.g. Pimlico bundler); chains Alchemy can't fully serve (BSC has no Alchemy bundler) simply present empty fields to fill manually instead of being blocked.
- Advanced per-chain: **additional RPC URLs** to enable quorum-mode log scanning (UI enforces the distinct-hostname rule from `validateRpcUrls`), and a **watch-only** toggle.
- Validate per chain: `eth_chainId` round-trip matches the selected chain (reusing `resolveChainConfigs`), bundler responds to a probe request.

**Step 4 — HyperFX strategy & price curves**
- The wizard configures exactly one `[[strategies]]` entry of `type = "hyperfx"` (stable strategy is config-file-only for now):
  - `maxOrderUsd`.
  - `token1` — exotic token contract address per selected chain, validated on-chain (bytecode exists, ERC-20 metadata readable; show symbol/decimals for confirmation).
  - **Bid/ask price curve editors**: table of (amount, price) points plus a live-rendered interpolated curve preview (mirroring `interpolated-curve.ts` so the preview matches runtime). Explain one-sided LP: omit the bid curve to only sell exotic, omit the ask to only buy.
  - Optional **pool-based pricing**: Uniswap V4 positions (`chain`, `tokenId`, optional `referencePrice` + `maxDeviationBps` price guard, `side`) with `spreadBps`. The form enforces the rules from `validateConfig`: curves or positions required, `side` only with pool pricing and no static curves, guard fields set together, no conflicting guards per chain.
- **Chain deployment for the strategy**: checkboxes for which configured chains it serves, expressed as `confirmationPolicies` entries (defaults from `DEFAULT_CONFIRMATION_POLICIES` shown and editable as amount→confirmations curves). Surface the caveat: without confirmation policies, HyperFX processes only same-chain orders.

**Step 5 — Treasury & rebalancing (optional, collapsible)**
- `[rebalancing]`: trigger percentage + base balances per asset (USDC/USDT) per chain, as a grid.
- `[binance]`: API key/secret for CEX rebalancing.
- `[vault]`: ERC-4626 vault entries (chain, vault address, sweep `threshold`/`minBalance`, `redeemOnShutdown`, `sweepIntervalMs`); resolve and display the underlying asset on-chain as validation.

**Step 6 — Advanced (prefilled defaults, rarely touched)**
- `maxConcurrentOrders`, queue (`maxRechecks`, `recheckDelayMs`), logging level, gas fee bump percents, **overfill protection** (`maxOverfillBps`, `maxConsecutiveClamps`) — note these matter more for HyperFX since consecutive venue-priced clamps halt the strategy — `[allowlist]` (global `users` + `bySource` per-chain overrides, address-validated), `[phantom_generator]` (enabled, chain, interval, token pairs), metrics port, data dir.

**Step 7 — Review & launch**
- Render the final TOML with secrets masked, a preflight checklist (RPCs and bundlers reachable per chain, substrate account funded, EVM signer has gas/stables/exotic on each chain), then **Save & Start**. Config written with `0600` perms; process transitions to run mode.

## 3. Operator dashboard (run mode)

Reuse the data the `MetricsService` already computes rather than duplicating collectors — refactor its balance/bid collection into a shared provider consumed by both Prometheus gauges and the UI API.

**Overview page**
- Filler status (running / paused / halted), uptime, active strategies, watch-only chains.
- **Balances per network**: USDC, USDT, native token, and exotic (token1) per chain — same sources as the `balanceUsdc/balanceUsdt/balanceNative/balanceExotic` gauges.
- **BRIDGE token balance** on Hyperbridge: free + reserved (existing `hyperbridgeFree`/`hyperbridgeReserved` collection), with a low-balance warning since bids stop when it runs dry.
- Vault positions (ERC-4626 share balances / underlying value) and Uniswap V4 position state, when configured.

**Prices page**
- Edit the HyperFX **bid/ask curves** (and `spreadBps` when pool pricing is active) with the same curve editor from setup. On save: (a) hot-apply by swapping the `FillerPricePolicy` instances on the running `FXFiller` (needs a small setter), and (b) persist to the config file so restarts keep the change. Show last-updated timestamp and the previous curve for comparison.
- If the running config contains a hand-written stable strategy, show it read-only ("edit via config file") rather than hiding it.

**Pause control**
- **Pause/Resume button**: new `pause()`/`resume()` on `IntentFiller` — keep the event monitor alive but stop dequeuing/filling new orders (in-flight fills complete). Effectively runtime-togglable watch-only across all chains. Persist paused state in the data dir so a restart while paused stays paused. Also expose per-chain pause via the existing per-chain `watchOnly` machinery.
- Surface **self-halt** from overfill protection distinctly (the FX strategy halts itself after consecutive venue-priced clamps) with a "review & resume" action — the "requires operator restart" path made operable from the UI.
- Graceful **Stop** button (existing shutdown path incl. optional vault redeem) — clearly separated from Pause.

**Activity page**
- Order feed (detected / filled / skipped with reasons), bid statuses from `BidStorageService` (pending/successful/failed/retracted), volume and profit USD counters, order processing latency. Live via WebSocket.

**Operations page**
- Rebalancing: current vs base balances, trigger threshold visualization, recent rebalance events (incl. Binance legs).
- Vault sweeps: last sweep time, next sweep countdown, manual "sweep now" / "redeem all".
- Allowlist editing (add/remove addresses at runtime + persist).
- Phantom generator status and toggle.
- Log level switch at runtime; live log tail.
- Config drift view: masked rendering of the active config with an "edit & restart" flow for settings that can't hot-reload (chains, endpoints, signer).

## 4. Backend API sketch

```
GET  /api/status              # mode, paused, halted, uptime, version
GET  /api/balances            # per-chain stables/native/exotic + hyperbridge BRIDGE free/reserved
GET  /api/strategies          # active strategies + current curves
PUT  /api/strategies/:i/curves
POST /api/pause | /api/resume | /api/stop
GET  /api/orders, /api/bids   # activity feeds (paged)
WS   /api/events              # order/bid/balance/log stream
# init mode only:
POST /api/setup/validate-alchemy-key    # returns per-chain derived RPC/bundler URLs
POST /api/setup/validate-rpc | validate-bundler | validate-token | check-substrate-balance
POST /api/setup/generate-substrate-key
POST /api/setup/save-and-start
```

## 5. Security notes

- Bind `127.0.0.1` only; refuse non-loopback bind addresses. Local machine access is the trust boundary.
- Strict CORS (no cross-origin requests honored) plus a required custom header on mutating requests, so a malicious website open in the operator's browser can't drive the API cross-origin — CSRF hygiene, not authentication.
- Secrets never round-trip to the browser after setup — API returns masked values (incl. the Alchemy key, which is embedded in the stored URLs); the setup POST that carries keys is localhost-only and never logged.
- Config file written `0600`; recommend (and support) env-var interpolation for `signer.key`, `substratePrivateKey`, Binance secrets as a follow-up.

## 6. Suggested phasing

1. **Backend plumbing**: default-config discovery, UI server skeleton, `pause()/resume()` on `IntentFiller`, shared balance provider extracted from `MetricsService`.
2. **Setup wizard** end-to-end (the "no config → init UI → running filler" path): private-key signer, Alchemy-prefilled chain step with manual overrides, HyperFX with static curves.
3. **Dashboard**: overview + pause + price updates (hot-reload setter on `FXFiller`).
4. **Long tail**: Uniswap V4 pool pricing in the wizard, quorum multi-RPC editing, vault/rebalancing/allowlist/phantom panels, MPC Vault/Turnkey signer tabs, activity feeds.

royvardhan:
Closes #1038.

Guided setup for the simplex filler plus a local operator dashboard.

- `simplex init`: terminal wizard that writes a commented config; re-running it updates an existing one, and legacy `[[strategies]]` configs are migrated to `[[pairs]]`.
- Bare `simplex` with no config: the same flow in the browser on loopback; Save & Start writes the config and transitions the process into running the filler.
- Operator dashboard while running: status, pause/resume, graceful stop, balances, live activity feed, per-market price-curve edits persisted to the config, vault sweep/redeem and runtime vault edits, allowlist, operator send.
- Markets follow the pair engine: any asset pair — same-asset transfer markets, stable-to-stable and exotic FX markets picked by registry symbol, with USD-anchor checks and reference-only price feeds handled in the wizard.
- Both wizards and the server share one validation core: the web UI imports the engine's browser-safe modules directly (`validateConfig`, crossed-book and USD-anchor checks, the asset registry) instead of hand-mirroring them, the server↔browser HTTP contracts live in a single `dto.ts` both sides compile against, and boot's symbol-resolution checks also run at the write gates so a config that would fail `simplex run` is rejected where it is written.

Breaking: `--admin-port` is replaced by `--ui <[host:]port>` / `--no-ui`. The UI is on by default at `127.0.0.1:8686`; mutating requests need the `X-Simplex-UI: 1` header.


seunlanlege:
Damn i was going to take a stab at this, will check it out

royvardhan:
> Damn i was going to take a stab at this, will check it out

Next step in this pr is unifying some interfaces and build steps for better readability.

Wizdave97:
Config rewrite silently deletes [binance], targetGasUnits, every log-level change from the dashboard, every init update run. 

Operator dashboard never re-polls /api/status — the overfill self-halt banner never appears. The PR's own safety feature is invisible in the PR's own UI.

Pimlico key corruption on derivation — silently broken bundler URLs on every derived chain.

Config writes are non-atomic — truncate-in-place on a file holding the only copy of a private key. Crash mid-write loses it. (This half of the write finding stands independent of permissions.)

Curve editor accepts values the filler rejects at boot — wizard says "Start now?", child process dies immediately.

Chains.tsx silent no-op on failed validation, activity feed dropping live SSE events, stale mainnet data surviving a testnet switch.

Wizdave97:
Some UI changes

Let's have a drop down to select our supported exotic tokens instead of manually pasting the addresses, pasting should be optional
Screenshot 2026-07-21 at 15 41 21

Also let's have supported vaults selected from a drop down list, user should be able to select multiple vaults on a chain
Screenshot 2026-07-21 at 15 41 49

Let's have a button to add new addresses to allow list instead of having a comma seperated list
Screenshot 2026-07-21 at 15 42 31

Wizdave97:
The user's account should be displayed on this review page
Both for substrate and evm

image

Wizdave97:
Should be able to update vault and rebalancing on the operator dashboard
image

Wizdave97:
I should be able to add a bid curve here is I want to if I initially configured as one sided LP
image

Wizdave97:
1. Every UI vault edit silently strips redeemOnShutdown. VaultRow in Operations.tsx:7 models only {chain, vault, threshold, minBalance}, and the load path at line 37 drops the field even though the /api/config DTO exposes it (UiServer.ts:98). So an operator who hand-set redeemOnShutdown = true, then tweaks a threshold in the dashboard, gets it persisted as absent → defaults to false at hydrate → that position is left in the vault instead of being redeemed to the wallet on the next graceful Stop. There's no UI control for the flag at all, so any UI vault edit permanently strips it from disk.

2. The vault "restart-needed" path can persist a config that bricks the next boot. In handleVaultUpdate, when the filler booted without a vault, the endpoint runs only VaultFundingPlanner.validateConfig and persists — it never hydrates. But the same-asset-per-chain guard and the chain-configured check live in VaultLiquidityState.hydrate, not validateConfig. So an operator with no vault at boot can add two same-asset vaults on one chain (or a vault on an unconfigured chain), get {applied:false, restartNeeded:true}, and write a TOML that throws on the next boot until hand-edited. The op.vault-exists path is safe — it hydrates first and rejects before persisting.

Missing vaults for all evm chains we support
Screenshot 2026-07-23 at 11 41 03

Filler's  wallet addresses should be  permanently visible on this page
Screenshot 2026-07-23 at 11 54 29

seunlanlege:
I don't think we need all this verbose information, after clicking "fill on this chain"

Screenshot 2026-07-24 at 10 07 14 AM

seunlanlege:
Lets remove the stable strategy 

Screenshot 2026-07-24 at 10 08 33 AM

seunlanlege:
We need to remove the EXT token from here

Screenshot 2026-07-24 at 10 09 12 AM

seunlanlege:
Screenshot 2026-07-24 at 10 10 42 AM


Lets remove rebalancing for now

seunlanlege:
Screenshot 2026-07-24 at 10 11 30 AM


This is too verbose, we should prefill these values for the user

seunlanlege:
Screenshot 2026-07-24 at 10 12 45 AM


UI nit show the substrate address in full & make the addresses more prominent so users can cross-check it when they copy

seunlanlege:
Screenshot 2026-07-24 at 10 13 43 AM

Showing the config.toml is unneccessary

seunlanlege:
Screenshot 2026-07-24 at 10 22 26 AM

logs seems unneccessary here

Screenshot 2026-07-24 at 10 25 28 AM

these don't make much sense

Wizdave97:
Let’s have a send function on the UI. That allows the operator send out any token.

For vault tokens, it should batch the redeem and transfer.

Wizdave97:
USDC, USDT /CNGN market should be the one enabled by default, not USDC or USDT crosschain swaps
image

Wizdave97:
vault shares should not be listed here, they should only be able to send out the underlying assets
image

Wizdave97:
Let's add a wallet history tab with list of transactions and links to block explorer
image

Wizdave97:
Error when trying to set these curve prices, I think i should be able to set anything, we should not correlate ask and bid curves, they are examined independently.
image

Wizdave97:
Also I should be able to delete all points on one curve to restore one sided LP

Wizdave97:
1. DNS rebinding is now a fund-theft vector. The new /api/send endpoint (commit d6845df0d) moves real funds — any token, native, or redeemed vault shares, to an arbitrary recipient. Its input validation is solid (checksum address + isAddress, bigint parseUnits, balance check, per-token decimals). But UiServer.handle() still validates no Host or Origin header — the only defense is the bind-time loopback check plus the X-Simplex-UI header, and DNS rebinding makes an attacker page same-origin, so it sets that header freely. The bind is a fixed, predictable 127.0.0.1:8686. Net: an operator who leaves a malicious tab open while the filler runs can have the full balance of any configured token swept to an attacker address. I verified directly — /api/send and TokenSender.ts exist, and there's no host check in handle(). This was a read-only weakness before; the send feature elevates it to theft. One-line fix: reject non-loopback Host in handle().

2. The CLI update-run write-gate deadlocks on two benign configs (reproduced by the reviewer):

A config with an empty [allowlist.bySource] table — assembleConfig carries it verbatim but emit-toml drops empty bySource, so the round-trip gate throws "Internal error… does not round-trip." The salvage loop only resets pairs, and carryPrefillExtras re-applies allowlist every retry → it can't recover.
A config with [rebalancing] but no baseBalances sub-table — emit-toml does Object.entries(baseBalances) unconditionally → TypeError, same salvage deadlock.
Both are exactly the "empty/degenerate carried section" class the audit set out to fix; two variants slipped through. Fix is small (strip/guard empty bySource and baseBalances in assembleConfig).

3. Confirmation-policy coverage is a boot-only check — a config passes both wizard gates but boot rejects it. ConfirmationPolicy.assertCovers() runs only at boot.ts:287, not in the shared validateConfig. The CLI wizard offers testnet chains (11155111, 421614, 84532, …) that have no DEFAULT_CONFIRMATION_POLICIES. If the operator skips fine-tuning, the wizard writes the config and reports it valid, then simplex run throws No confirmation policy for chain 11155111. That directly contradicts the "one validation core" claim, and the wizards have the chain ids on hand to gate it. Medium, but it's the one real hole in the headline refactor.

Wizdave97:
There's still an issue, simplex is not following my set curves, with the following prices it says the order is not profitable

Screenshot 2026-07-28 at 15 50 31


image

I'm expecting it to submit a bid since the amount requested by the user is less than the ask price I configured

Wizdave97:
Let's add a select all to this
image

Wizdave97 (commented):
LGTM once pending comment is resolved.
Extensive UI improvements can come in another PR.

Wizdave97:
Vault selection or addition on operator dashboard should have the same flow as the vault selection on onboarding flow.
image

Wizdave97:
When vaults are selected on the operator dashboard, show the human readable vault name also instead of just the address for easier identification

Wizdave97:
Add new markets missing on operator dashboard
image
Increase page width so the rows appear correctly, no wrapping
image

Wizdave97 (approved):

seunlanlege (approved):
Beautiful

PR #1058, opened 2026-07-19
[simplex]: pair-based FX markets, unified asset registry & hardened RPC quorum

seunlanlege:
Simplex is now a single pair engine: every market it serves — cross-asset swaps and same-asset cross-chain transfers alike — is a `[[pairs]]` entry, priced by its own curves. The `[[strategies]]` array (and the stable strategy) no longer exist.

## Trading pairs

```toml
# Same-asset cross-chain market (replaces the stable strategy):
# ask below par IS the spread, realized in-kind on every fill.
[[pairs]]
token0 = "USDC"
token1 = "USDC"
maxOrderSize = "100000"
askPriceCurve = [
    { amount = "100",    price = "0.99"  },   # small orders: 1% spread
    { amount = "100000", price = "0.999" },   # large orders: 0.1%
]

[[pairs]]
token0 = "USDC"
token1 = "CNGN"
maxOrderSize = "5000"       # per-order cap, in token0 units
bidPriceCurve = [ { amount = "0", price = "1580" } ]  # CNGN per USDC
askPriceCurve = [ { amount = "0", price = "1550" } ]

[[pairs]]
token0 = "ZARP"             # non-USD quote side — anchored via CNGN below
token1 = "CNGN"
maxOrderSize = "90000"
askPriceCurve = [ { amount = "0", price = "84" } ]

[confirmationPolicies."EVM-1"]  # engine settings are top-level; EVM- keys accepted
points = [ { amount = "5", value = 3 }, { amount = "5000", value = 12 } ]

[vault.uniswapV4]           # V4 positions/side/spreadBps join the [vault] section
positions = [ { chain = "EVM-8453", tokenId = "2087350" } ]
```

- Curves are quoted in **token1 per token0** and are the only price source — no external price feed anywhere. Trade sizing (curve amount axis, per-pair `maxOrderSize`) is pair-local in token0 units.
- **Same-token pairs** (token0 == token1) are the same-asset cross-chain market: ask-only, prices strictly below par, cross-chain only. Any registry asset works, USD-stable or not (cNGN/cNGN), provided the asset is USD-anchored (below).
- Every pair is priced, capped, and gated independently: per-pair one-sided LP via curve omission; the pair list doubles as an allowlist. Same-chain cross-asset swaps are served; same-chain same-token self-swaps are rejected.
- **No backwards compatibility**: `StableFiller`, the direct-fill path, and all legacy config parsing are deleted. A config containing `[[strategies]]` fails at startup with a pointed migration error.

## Two independent profit gates

An order fills only when both hold:

1. **Fees cover execution**: `order.fees ≥ fill gas + relayer fee`, with the relayer message budgeted at `RELAYER_MESSAGE_GAS` (1M) priced on the source chain. The SDK's placement side attaches `(fill gas + the same 1M-gas uplift) × 1.05` while the solver's requirement carries no padding — an SDK-placed order always clears this gate on every route.
2. **Every leg is independently profitable in its own quote asset.** Same-token legs: realized in-kind spread > 0 per leg. Cross-asset legs: the fill marked at the pair's opposite curve nets positive token0, per leg — margins in different quote assets are never summed (a winning leg in a cheap currency cannot mask a losing leg in an expensive one). Profit credits round down when rescaling decimals, so dust cannot flip break-even to profitable.

One-sided pairs and venue-priced (Uniswap V4) legs are **directional**: a single price has no opposite side to mark against, so only gate 1 applies. The venue price guard (band check against a static reference) is the pool-specific defense and now also runs on the confirmation-sizing path. There is deliberately **no overfill clamp**: the curve output is the price the engine quotes, and with per-leg gating every fill nets at least its spread as marked by the operator's own curves.

## USD-anchored confirmation sizing

Confirmation-depth curves have a USD amount axis, and the operator's own curves are the price feed: USDC/USDT/DAI are $1 anchors, every curve-priced cross-asset pair is an FX edge, and each leg's token0 notional converts to dollars through that graph (first anchoring pair in declaration order wins; stables are pinned at $1 and never re-priced). **Anchoring is a startup requirement** — every pair's token0 must be a USD stable or reach one through curve edges, directly or transitively (USDC/CNGN + ZARP/CNGN anchors ZARP through CNGN). A pair marked **`referenceOnly = true`** anchors without opening a market: it contributes its FX edge (and stays live-editable via the admin server) but never matches order legs — how a lone cNGN/cNGN transfer market gets its USD rate. Every configured chain must have a confirmation curve (built-in defaults cover Ethereum, BSC, Polygon, Base, Arbitrum, Unichain) or startup fails.

## Startup validation & live-edit invariants

Configs that could silently lose money or silently do nothing are startup errors, not runtime surprises:

- Crossed or zero-spread two-sided books (bid ≤ ask anywhere on the combined curve grid) are rejected — such a pair would mark every fill as a loss and go silently dead. Same-token asks must be strictly below par for the same reason.
- Chain keys accept the `"EVM-<id>"` style used everywhere else (and bare `"<id>"`); anything else fails loudly in `[confirmationPolicies]` and per-chain `watchOnly` — previously both were silently discarded, one of which meant trading real capital on a chain configured watch-only. `[assets]` chain keys are validated the same way.
- Admin live curve edits enforce the same book invariants before `replacePoints`: an update that would cross the book or push a same-token ask to par returns 400 and applies nothing.
- Aliasing a symbol to an existing token's address, unknown pair symbols, and reverse-duplicate pairs are rejected as before.

## One asset registry

Pairs reference assets purely by symbol; every shipped symbol resolves from the **SDK chain registry** (`chain.ts`) — the single source of truth shared with the rest of the SDK, so there is no parallel table in Simplex to drift out of sync:
- USDC, USDT, DAI, cNGN (testnets included), plus curated mainnet deployments of **ZARP** (Ethereum/Polygon/Base), **EURC** (Ethereum/Base), **XSGD** (Ethereum/Polygon), **TRYB** (Ethereum) — every address taken from the issuer's official docs and verified on-chain (`symbol()`/`decimals()`) before inclusion in `chain.ts`.
- `[assets]` remains only as an escape hatch: a bare `symbol → chain → address` map.

## Hardened RPC quorum

- Registry of keyless public RPC endpoints for Ethereum, Arbitrum, Base, Polygon and BNB Chain (official gateways + PublicNode, dRPC, 1RPC — verified live), appended to operator endpoints for **quorum-checked reads only**, deduplicated by hostname. Fills and gas estimation stay on operator endpoints.
- The quorum is **weighted across two tiers**: a BFT quorum of the operator's own endpoints must agree, plus `min(2, available)` public witnesses. Operator failures are intolerable (the call fails loudly, never proceeding on public endpoints alone); public failures are tolerable.
- **Confirmation counting** runs through the quorum: the tiered quorum must agree on the receipt's inclusion `(blockHash, blockNumber)`; "receipt not found" is a valid *no* vote, so a reorged transaction stops the count instead of erroring past it.
- Reads **resolve the moment the tiered quorum is met** — stragglers settle out of band, so a hung endpoint (30s × 3 retries) can no longer stall the 1s confirmation poll. Negative outcomes still wait for the full response set. Same-chain orders (zero required confirmations) skip the quorum read entirely.

## Verification

- Build passes including the strict DTS build; changed files are Biome-clean.
- 176 offline tests pass: pair validation and the startup rules above, per-leg profit gating (including the masked-loss cross-currency repro), USD-anchor factor derivation and conflict resolution, quorum tiering + early exit with hung providers, receipt-not-found aggregation, admin-server invariant enforcement, registry resolution/overrides, and cross-decimal same-token quoting.
- The live testnet E2E (BSC Chapel → Polygon Amoy) and the phantom-filler E2E pass in CI against this branch.


Wizdave97:
## Blocking

**1. Legacy `maxOrderUsd` becomes a per-pair cap — up to 2× the intended risk limit**
`strategies/fx.ts:60-76` + `:827-837`

`legacyExoticPairs` emits two pair objects (`USDC/EXOTIC`, `USDT/EXOTIC`), each carrying `maxOrderSize = maxOrderUsd`, and `sizeOrder` groups budgets by pair identity (`cappedByPair`). On `main` there was a single `remainingUsd = min(totalInputUsd, maxOrderUsd)` decremented across every leg (main `fx.ts:342/367/431` — I diffed this).

With `maxOrderUsd = 5000` and inputs `[3000 USDC, 3000 USDT]`: old behavior rejected/truncated at 5000; new behavior fills both legs for **6000 USD exposure**. This silently loosens a documented risk control on the path every existing config still uses. Size-tiered curves also now evaluate at each pair's subtotal rather than the combined notional, quoting the better small-order price.

**2. `AssetRegistry.getAddress` returns `0x000…0` instead of `null` for built-ins**
`config/asset-registry.ts:171-185`

The `catch { address = undefined }` is commented "SDK registry has no entry — treat as absent", but `getDaiAsset`/`getUsdcAsset`/`getUsdtAsset` never throw. They return `?? "0x"` for unconfigured chains, and for ~10 *configured* chains `chain.ts` stores a literal `DAI: 0x0000…0000` (Soneium mainnet 1868 included, which also has `USDT: 0x0`). Both sentinels are truthy, so `address ?? null` yields an address, not `null`.

That matters because `fx.ts:969` treats `ADDRESS_ZERO` as the **native token** for balance reads. A natural config — `token0 = "DAI"` (advertised as shipped, and in `USD_STABLE_SYMBOLS`) — passes `validatePairConfigs`, which only checks symbol membership, never resolution. The `token0Source && token1Dest` guard at `fx.ts:999-1013` then passes on Soneium, and a leg with `address(0)` matches the DAI side: the filler values 1e18 wei as "1 DAI" ≈ $1 and pays out 1 native ETH. Newly reachable — on `main` the stable side was matched only against USDC/USDT and DAI wasn't tradeable.

Fix: treat `"0x"`, `undefined`, and `ADDRESS_ZERO` as absent, and reject a pair at startup if a symbol resolves nowhere.

**3. Shipped XSGD address fails the PR's own address validator**
`config/asset-registry.ts:71`

`0x70e8de73cE538DA2bEEd35d14187F6959a8ecA96` has an invalid EIP-55 checksum (needs a capital `E` at index 7). Verified with the repo's own viem: `isAddress()` → `false`. `matchLeg` lowercases before comparing so it survives there, but `bin/simplex.ts:661` passes the un-lowercased value to `MetricsService`.

The contract itself is correct — this is a transcription slip, not a wrong token. The real issue is that `validateAssetDefinitions` runs only on operator `[assets]`, never on `KNOWN_ASSETS`, which is why the docstring's "verified before inclusion" claim went unchecked. Worth a one-line test asserting `validateAssetDefinitions(KNOWN_ASSETS)`.

*The other seven curated addresses verified clean* — correct checksums, and `symbol()`/`decimals()` confirmed on-chain against issuer docs. Notably XSGD-Polygon is correctly the native deployment, not the deprecated bridged one. No `decimals` are stored in the table (they're read on-chain), which is the right call.

## Should fix

**4. Quorum threshold is computed over all N, so a rate-limited public RPC is as fatal as a lying one**
`services/QuorumPublicClient.ts:100-105`

The BFT math is right (`floor(2N/3)+1`, descending sort at index `threshold-1` picks the *lowest* head among the quorum — conservative, never overstates depth). But failed providers are excluded from the agreement set while the threshold stays fixed at `quorumThreshold(N)`. At the documented default N=5 (1 operator + 4 registry), threshold 4 tolerates exactly one unreachable provider.

The scanner hits all 5 providers every 1000ms per chain. An operator running the five registry mainnets hits `1rpc.io` from one IP across all of them concurrently; keyless tiers will 429. Two throttled providers → every quorum read throws → `scanBlocks` throws, the cursor never advances, and **the filler goes blind to all orders** while logging an error each second. There's no opt-out, and this failure mode didn't exist on `main` for single-RPC operators. Registry endpoints also bypass the `fetchChainId` verification that operator URLs get.

**5. Reverse-duplicate pairs are accepted, and matching is declaration-order dependent**
`config/pairs.ts:81-89` + `fx.ts:998-1024`

`seen` keys on the ordered label `${token0}/${token1}`, so `USDC/CNGN` and `CNGN/USDC` both validate. `matchLeg` returns the first array match and matches reversed pairs via its second branch. Declaring both orientations means a `CNGN → USDC` leg prices off whichever came first — if that's the `CNGN/USDC` entry, it's priced in CNGN-per-USDC units, a ~2.5M× mispricing with the cap denominated in the wrong asset. Normalize the label and reject the reverse.

## Worth a look

- **`getOrderUsdValue` sums notionals across different quote assets** (`fx.ts:830-837`): a 1000 USDC leg plus a 1000 ZARP leg (≈$55) reports `inputUsd = 2000` into a USD-calibrated confirmation curve. The cap itself doesn't breach — `token0ForLeg = min(legMaxToken0, remaining)` holds — so this only mis-sizes confirmation depth.
- **Admin index namespaces collide** (`bin/simplex.ts:492` vs `:523`): pairs path uses `adminStrategies.length`, legacy uses `strategyIndex`. Also `adminTokenMaps` is never populated on the pairs path (benign — the pair label survives as the display name).
- **Legacy `token1` maps are now checksum-validated at boot** via `legacyExoticPairs` → `AssetRegistry`, failing with `assets.EXOTIC: invalid address…` — a symbol that appears nowhere in the operator's TOML.
- **`index.ts:14-24` exports mutable `KNOWN_ASSETS` and `USD_STABLE_SYMBOLS`**; combined with `AssetRegistry`'s permanent `addressCache`, a mutation leaves resolution incoherent. Freeze them.
- **`quotePhantomFill` reads the output token's decimals on `order.source`** (`fx.ts:1112`) — carried over from `main`, but riskier now that pairs routinely mix 6- and 18-decimal assets across chains.
- **Venue pricing overrides configured curves** and quotes pool mid on both sides, collapsing the spread to zero (`fx.ts:919-930`). I checked `main` — `getVenuePrice` also returned mid for both bid and ask, so this is **not** a regression, but the new per-pair docs imply curves should win when explicitly set. Worth an explicit precedence decision.

seunlanlege:
Thanks for the thorough review — all points addressed in 8f5a570 (which landed between your read and this reply) and 89161de.

**1. Legacy `maxOrderUsd` → 2× exposure** — resolved by deletion: 8f5a570 removed `legacyExoticPairs` and all legacy config parsing entirely (no shims; a `[[strategies]]` config now fails at startup with a migration error). Per-pair caps and per-pair curve evaluation are the *designed* semantics of `maxOrderSize` — but the legacy path silently reinterpreting a documented USD cap as two per-pair caps was exactly as bad as you described, and it no longer exists.

**2. Sentinel addresses from the SDK registry** — confirmed and fixed in 89161de. You were right that the `catch`-based absence handling was dead code: the getters return `?? "0x"` and chain.ts stores literal zero addresses. `AssetRegistry.getAddress` now filters `undefined`/`"0x"`/`address(0)` (and non-addresses) to `null`, `validateAssetDefinitions` rejects the zero address in `[assets]`, and startup verifies every pair symbol resolves to a real deployment on at least one configured chain — so the `token0 = "DAI"`-on-Soneium path dies at boot instead of matching native ETH.

**3. XSGD checksum** — fixed (`0x70e8dE73…`, verified against viem's `getAddress`), and your suggested test is in: `validateAssetDefinitions(KNOWN_ASSETS)` now runs in the suite, so the curated table goes through the same validator as operator config. My original on-chain verification checked `symbol()`/`decimals()` through a lowercasing path, which is how the casing slip survived — the test closes that gap.

**4. Rate-limited registry endpoints as fatal as liars** — fixed with a two-tier circuit breaker in 89161de. Registry endpoints are now (a) `eth_chainId`-verified once at startup (mismatch → permanent ejection, closing the `fetchChainId`-bypass you noted) and (b) benched for 5 minutes after 5 consecutive errors/timeouts, with the threshold recomputed over the active set. Operator endpoints are never ejected, so the quorum floor is exactly the operator's own configuration — pre-PR behavior. Ejection is failure-only, never divergence-based: a provider that *answers* with different data keeps consuming fault budget rather than being strippable, which preserves the BFT property. Tests cover benching + threshold recompute, the operator floor, and divergent-but-responsive never ejecting.

**5. Reverse-duplicate pairs** — fixed: `validatePairConfigs` and the engine both reject a pair whose reverse orientation is already declared ("a market has one orientation; use its bid curve for the reverse direction").

On the worth-a-look items:
- **Mixed-unit `getOrderUsdValue`** — accepted tradeoff of the no-price-feed design; the docs call it out ("for others it is denominated in the quote asset, so tune the per-chain confirmation curves accordingly"). Mixed-quote multi-leg orders would need a curve-derived cross-rate to do better, which we deliberately removed. Open to revisiting if it bites in practice.
- **Admin index collision / legacy `token1` checksum errors** — gone with the legacy path in 8f5a570.
- **Frozen exports** — `KNOWN_ASSETS` and `PUBLIC_RPC_URLS` are deep-frozen; `USD_STABLE_SYMBOLS` is typed `ReadonlySet`.
- **`quotePhantomFill` destination decimals** — fixed to resolve output decimals on `order.destination`.
- **Venue vs curves precedence** — decided in favor of curves: explicitly configured curves always win; the pool only prices pairs with no curves at all (test added). This also means the price guard now only gates genuinely venue-priced pairs, which matches its intent.

seunlanlege:
Follow-up on point 4: the circuit-breaker approach described above was replaced in b53e0db with something simpler and stateless. Per call: HTTP 429s just exclude that endpoint from the call's quorum (rate limiting is noise, not ill health); any other error pauses the endpoint for 5 minutes; the agreement threshold is computed over the providers that actually responded, floored at the operator set's own BFT bound (`quorumThreshold(operatorCount)`) so missing providers can never weaken the quorum below the configured `rpcUrls`. No failure counters, no benching state, no startup chainId probe. Divergent responders are still never excluded. Tests updated accordingly.

Wizdave97:
1. Quorum reads have no early exit — every poll blocks on the slowest provider (HIGH, liveness)
QuorumPublicClient.ts:159,216,300 — all three methods await Promise.allSettled(allClients), so the result is computed only after every provider settles, even once the threshold is already met. With http(url,{timeout:30_000,retryCount:3}), one hung/throttled public endpoint stalls each read up to ~30–120s. This sits on the fill hot path (filler.ts:508 polls confirmations every confirmationPollMs), and confirmationPollMs is 1000ms for all registry-backed mainnet chains. Fix: resolve as soon as meetsQuorum is satisfiable and let stragglers settle out of band.

2. Venue-priced legs can't pass the new spread gate (MEDIUM, correctness)
fx.ts:1046 sets oppositeRate = venueRate = rate for venue legs, so GATE-2's FX margin (fx.ts:699) computes to ≈0 and fxMarginQuote.lte(0) (fx.ts:746) rejects the fill whenever the output divides evenly (only sub-wei floor residue lets it through otherwise). Venue pricing is pool-mid on both sides by design — it can never produce the spread the new gate demands. The documented Uniswap-V4 venue feature is effectively disabled. Intended behavior is almost certainly oppositeRate: null (directional, GATE-1-only).

3. Confirmation-depth USD sizing bypasses the venue price guard (MEDIUM, security)
referenceRate (fx.ts:970) calls venueUsdPrice(...) with no checkPriceGuard, unlike the trade-pricing path (fx.ts:1041). For a venue-priced, exotic-input cross-chain order, the order's USD notional — and thus required confirmations — is driven by an unguarded pool quote. A manipulated/thin pool understating the price shrinks the notional → fewer confirmations → weaker reorg protection, exactly what the guard exists to prevent. (Partly contained: fully-venue orders are rejected by #2, but mixed venue+curve orders can still be sized down.)

Wizdave97:
1. GATE 2 sums FX margins across different token0 units — demonstrated with a failing test
fx.ts:649,699,704,746

Each leg contributes in its own pair's token0. USDC and ZARP are added 1:1. I reproduced it:

✓ sanity: the loss-making USDC/CNGN leg ALONE is correctly refused
✓ sanity: usdFactors values a +600 ZARP margin at ~$46.80, not $600
× BUG: expected 108 to be +0     ← true USD margin is −$453; engine bids
A +600 ZARP leg (≈$47) masks a −$500 USDC leg; the order is scored +108 and executed. filler.ts:612 gates execution on profitability > 0, so this is a funds gate, not just a ranking signal. The class already has usdFactors() for exactly this conversion and warns about the hazard 200 lines earlier (fx.ts:942: "mixes units — anything decision-making must use … getOrderUsdValue"), then does it anyway.

New in this PR: on main this quantity was USD-denominated and report-only. This PR both generalized token0 to any asset and promoted the sum to a hard gate. Fix: multiply each leg by usdFactors().get(token0) before summing, or gate per-leg as the same-token branch already does. Test saved at scratchpad/gate2-mixed-units-probe.test.ts.

2. SDK-placed cross-chain orders are systematically rejected by GATE 1
IntentGateway.ts:352 vs GasEstimator.ts:386 → fx.ts:720

Placement attaches G + R_raw + G/100; the solver requires G + 1.005·R_raw. The difference is G/100 − R_raw/200, negative whenever R_raw > 2G — the normal case for an expensive source and cheap destination (Ethereum → Base/Arbitrum). For Ethereum→Base that's ~$60 vs fractions of a cent: it fails by orders of magnitude.

I confirmed this is PR-introduced. On main, placement used CROSS_CHAIN_ORDER_FEE_GAS_BUMP = 600_000 while the solver used postRequestGas = 400_000, and the 1005/1000 bump applied to the dest fee token — 50% slack. This PR unified both to RELAYER_MESSAGE_GAS = 1_000_000 and moved the bump to the source side, turning that slack into a deficit. The code comment states the goal as "a user placing via the SDK attaches exactly what a solver requires" — it's short by 0.5% of the relayer fee.

3. The overfill clamp is gone for stable/same-token orders
fx.ts:511-534, stable.ts:278-291 (deleted)

StableFiller bounded payout to (1 + maxOverfillBps) × requested. Its replacement has the clamp explicitly disabled — the comment concedes it "removes the per-leg loss bound that previously protected against a bug / stale cache / manipulated venue price." Demonstrated: user requests 1 USDC against a 1000 USDC input → engine pays 995 USDC (StableFiller: 1.05).

I verified in IntrinsicIntents.sol:84-91 that surplus above totalRequired goes to beneficiary and protocol, never the solver — so it's a pure donation. Nuance: the clamp was already disabled for FX on main (deliberate); the regression is that stable/same-token orders now inherit that.

Should fix
4. Two config keys silently discarded. [confirmationPolicies."EVM-1"] → Number("EVM-1") = NaN, and per-chain watchOnly keyed "EVM-1" → parseInt = NaN (simplex.ts:298). Every other table in the config uses EVM-N keys, so this is the natural thing to write. The merge {...DEFAULT, ...user} adds a dead key rather than replacing, so assertCovers still passes: an operator hardening Ethereum reorg protection silently gets the default (~15× less depth). The second makes the filler commit capital on a chain meant to be observed. Both fail silently, no warning.

6. Same-chain zero-confirmation orders still hit the quorum and are dropped on failure. filler.ts:505 runs the initial getTransactionConfirmations unconditionally even when requiredConfirmations === 0. On QuorumError, Promise.all rejects, the outer catch logs "Error processing order" — no requeue. Same-chain fills, which gain nothing from the quorum, are now gated on third-party public RPC availability.

7. Venue-priced pairs: GATE 2 is decided by rounding dust. fx.ts:1047 sets oppositeRate === rate, so the margin reduces algebraically to zero; the order passes or fails on floor() truncation residue. Where the rate divides evenly it's rejected outright. Combined with the disabled clamp and an optional price guard, a manipulated V4 pool draws no objection.

8. Price guard missing on the confirmation-sizing path. referenceRate (fx.ts:970) calls venueUsdPrice with no checkPriceGuard, unlike resolveLegRates. That feeds getOrderUsdValue → requiredConfirmations. On main the guard ran before any sizing, so coverage narrowed here. Understating the notional shrinks confirmation depth — the reorg exposure the guard exists to prevent.

Worth a look
Partial-fill margin uses full input.amount while escrow releases proportionally (sign is safe only via a non-obvious invariant at fx.ts:582); partial fills forfeit 100% of order.fees (IntentsBase.sol:412) that GATE 1 just certified; funding-venue withdrawal gas (400k/call) reaches callGasLimit but not executionCost; same-token pairs at exactly price = 1 validate but then reject every order; nothing validates bid > ask or that maxOrderSize lies within the curve's domain; adjustDecimals ceilings the input-credited side (fx.ts:680); admin curve edits bypass the par and crossed-book invariants; USD_STABLE_SYMBOLS exported unfrozen; [assets] chain keys unvalidated; metrics token map keyed by chain only. Docs: the docker-compose quickstart in simplex.mdx:512 doesn't start (missing run subcommand).

seunlanlege:
Thanks again — both rounds were excellent, the mixed-units repro especially. Everything is addressed; one item is a deliberate won't-fix. By finding:

**GATE 2 sums FX margins across different token0 units** → fixed in 645da30. Rather than converting to USD before summing, the gate is now **per leg, in each leg's own quote asset** — a sign check needs no common denominator, and a solver has no reason to fill any leg that loses money on its own. Margins from different pairs are never summed for a decision; the USD-converted sum (via the anchor factors) survives only as the ranking score and telemetry. Your masked-loss scenario is in the suite as a regression test (+600 ZARP hiding a −$500 USDC leg → rejected), alongside a multi-pair order where every leg wins → fills.

**SDK-placed cross-chain orders systematically rejected by GATE 1** → fixed in f3f2534. The solver-side 0.5% padding is gone entirely — `RELAYER_MESSAGE_GAS` (1M) is already a generous buffer over actual delivery gas — and placement now attaches `(fill gas + settlement uplift) × 1.05`. What a user attaches is strictly above what a solver requires on every route, including expensive-source/cheap-dest.

**Overfill clamp removed for same-token orders** → deliberate, staying as is. With per-leg gating, every fill nets at least its spread as marked by our own curves — the curve output is the price the engine quotes, and a user requesting less than curve-fair gets curve-fair. The old clamp only captured windfall from mispriced requests; we prefer consistent quoted pricing over windfall harvesting.

**Quorum reads block on the slowest provider** → fixed in d3eda77. All three read paths settle incrementally and resolve the moment the tiered quorum is satisfiable; stragglers settle out of band. Early exit is positive-only (an agreeing group can only grow, so the first satisfiable quorum is final); negative outcomes still wait for the full response set before throwing, and receipt-not-found votes never decide early. Regression tests pin all three paths resolving under 2s with two providers hung forever.

**Venue-priced legs can't pass GATE 2** → fixed in a0f258c, your suggested shape: `oppositeRate: null`. A pool mid is one price, not a book, so venue legs are directional (gate-1-only) like one-sided curve pairs — no more rounding-dust lottery. Test: an exactly pool-fair fill (the old guaranteed-rejection case) now fills.

**Price guard missing on the confirmation-sizing path** → fixed in the same commit. `referenceRate` runs `checkPriceGuard` on the pool quote exactly like trade pricing; a quote breaching the band refuses to size and the order is skipped. Landed together with the venue fix since that's what makes the path live.

**"EVM-1"-style config keys silently discarded** → fixed in 2b430fb. `parseChainKey` accepts `"EVM-<id>"` and bare `"<id>"` in `[confirmationPolicies]` and per-chain `watchOnly`; anything else is a startup error in both `validateConfig` and the `ConfirmationPolicy` constructor. Normalized keys correctly override the built-in defaults instead of sitting beside them. `[assets]` chain keys are validated to `EVM-<id>` for the same reason.

**Same-chain orders gated on the quorum** → fixed in 852ea96. `requiredConfirmations <= 0` short-circuits the wait entirely — no initial read, no QuorumError-driven drop.

From the worth-a-look list, also landed:
- **bid > ask now validated** (29db3bb): crossed and zero-spread books are rejected at startup, pointwise across both curves' breakpoints — and **same-token asks at exactly par** are rejected with them (the validate-but-never-fill case you flagged).
- **Admin edits obey the same invariants** (2b430fb): a live update that would cross the book or push a same-token ask to par returns 400 before `replacePoints`.
- **adjustDecimals ceiling on the credit side** (2b430fb): same-token spread credits now truncate; your break-even-plus-one-atto-unit case is a regression test.
- **Metrics token map** (2b430fb): `chain → address[]`, every pair's token1 keeps its gauge.
- **Docs docker-compose** (2b430fb): `run` subcommand restored.

Still open, acknowledged: partial fills forfeiting `order.fees` that gate 1 certified, and funding-venue withdrawal gas missing from gate 1's execution cost — both are real economics gaps we'll take in a follow-up, along with pinning the partial-fill margin invariant with a comment/test. On `USD_STABLE_SYMBOLS`: `Object.freeze` doesn't actually prevent `Set.add` (it only freezes properties, not the internal slots), so the `ReadonlySet` type is the real guard; leaving as is.

Wizdave97 (approved):

sdk/packages/sdk/src/protocols/intents/GasEstimator.ts
```diff
@@ -43,6 +40,24 @@ import { CryptoUtils } from "./CryptoUtils"
...
  */
+/**
+ * Gas budget assumed for delivering and executing the cross-chain RedeemEscrow
+ * POST message on the SOURCE chain (the message a cross-chain `fillOrder`
+ * dispatches back to release escrow to the filler). The relayer fee carried by
+ * that dispatch — and the amount a filler's `order.fees` must cover — is this
+ * gas priced on the source chain. Sized conservatively so the relayer is
+ * reliably incentivised to deliver.
+ *
+ * TODO: replace this flat budget with a measured estimate via
```
royvardhan, line 51: Is this still left?

sdk/packages/simplex/src/config/asset-registry.ts
```diff
@@ -0,0 +1,217 @@
...
+export const USD_STABLE_SYMBOLS: ReadonlySet<string> = new Set(["USDC", "USDT", "DAI"])
+
+/**
+ * Curated registry of additional stablecoin deployments on the supported
+ * mainnets, maintained here so operators reference assets purely by symbol —
+ * no address configuration required. Every address below was taken from the
+ * issuer's official documentation (Circle, ZARP Stablecoin, StraitsX, BiLira)
+ * and verified on-chain (`symbol()` + `decimals()`) before inclusion.
+ */
+export const KNOWN_ASSETS: Record<string, AssetDefinition> = {
```
royvardhan: This makes a second token address registry alongside the SDK's `chain.ts`. cNGN can't move out of the SDK since phantom quoting depends on it, so overlap between the two is permanent, and the precedence here (user assets, then curated, then SDK built-ins) means a curated entry would silently shadow a later SDK correction. When the FE needs ZARP/EURC/XSGD they'll have to land in the SDK registry anyway, at which point every new asset means updating both places.

A startup invariant: a symbol must not exist in both `KNOWN_ASSETS` and the SDK built-ins, or if it does, addresses must agree per chain or startup fails. Cheap, and turns silent drift into a loud error.

sdk/packages/simplex/src/config/public-rpcs.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Registry of well-known, keyless public JSON-RPC endpoints for the EVM networks
+ * Simplex supports. These are merged into the *quorum* provider set alongside the
+ * operator's own endpoints (see `FillerConfigService.getQuorumRpcUrls`) so that
+ * order detection and cross-chain confirmation counting never rest on a single
+ * provider's — or a single organisation's — view of the chain.
+ *
+ * Selection criteria: each chain lists endpoints operated by organisationally
+ * independent providers (the network's official gateway where one exists, plus
+ * PublicNode/Allnodes, dRPC and 1RPC), all reachable without an API key. The
+ * quorum threshold is `floor(2N/3) + 1`, so four public endpoints on top of a
```
royvardhan: This comment (and the PR description) describes a flat BFT quorum of `floor(2N/3) + 1` over the combined set, "N=5, threshold 4". That's not what `QuorumPublicClient` implements: it's two-tier, a BFT quorum over the operator endpoints plus a flat floor of 2 agreeing public witnesses. With a single operator endpoint the real requirement is 1 operator + 2 public, i.e. 3 of 5, not 4 of 5. Please update this comment and the PR body to describe the tiered model.

sdk/packages/simplex/src/config/pairs.ts
```diff
@@ -0,0 +1,248 @@
...
+	const unanchored = unanchoredToken0Symbols(
+		pairs.map((p) => ({
+			token0: p.token0,
+			token1: p.token1,
+			hasCurve: (p.bidPriceCurve?.length ?? 0) >= 1 || (p.askPriceCurve?.length ?? 0) >= 1,
+		})),
+	)
+	if (unanchored.length > 0) {
+		throw new Error(
+			`pairs: no USD anchor for ${unanchored.join(", ")} — confirmation depth is sized in USD using your own curves as the price feed. Add a curve-priced pair against a USD stable (e.g. USDC/${unanchored[0]}), directly or through an already-anchored asset`,
```
royvardhan: This check means a same-token non-USD market can't run alone. An operator who only wants same-asset CNGN transfers (CNGN/CNGN) is forced to declare a curve-priced USD pair like USDC/CNGN, and that pair is a live tradeable market, not a reference quote. The workaround is a tiny `maxOrderSize` with a one-sided curve, but that's a workaround. Is a price-feed-only pair (something like `referenceOnly = true`) planned? If not, the migration error and docs should at least spell out that anchoring a same-token pair requires opening a real USD market.

sdk/packages/sdk/src/protocols/intents/GasEstimator.ts
seunlanlege: I think it would slow down fee calculation & fills, so leaving it for future optimizations

royvardhan (approved):

PR #1059, opened 2026-07-20
[sdk]: restart GET recovery after pruned consensus proof

ddboy19912:
## Summary

Adds resilient SDK recovery handling for pruned Hyperbridge consensus updates during source-side GET cancellation.


sdk/packages/sdk/src/utils/exceptions.ts
```diff
@@ -23,3 +23,27 @@ export class ExpectedError extends Error {
...
+	constructor(message = "Error fetching Consensus update time", options?: { cause?: unknown }) {
+		super(message, options)
+		this.name = "Hyperbridge/SDK/MissingConsensusUpdateTimeError"
+	}
+
+	static isError(error: unknown): error is MissingConsensusUpdateTimeError {
+		if (error instanceof MissingConsensusUpdateTimeError) return true
+
+		const messages: string[] = []
+		let current = error
```
Wizdave97: Why can't we just use a simple substring check?

ddboy19912: Done

sdk/packages/sdk/src/utils/exceptions.ts
```diff
@@ -35,15 +35,9 @@ export class MissingConsensusUpdateTimeError extends Error {
...
-		let current = error
-		while (current && typeof current === "object" && messages.length < 4) {
-			if ("message" in current && typeof current.message === "string") messages.push(current.message)
-			current = "cause" in current ? current.cause : undefined
-		}
-
-		return messages.some((message) => /error fetching consensus (update time|state)|consensus update time.*(?:missing|not found|prun)/i.test(message))
+		return (
+			error instanceof MissingConsensusUpdateTimeError ||
+			(error instanceof Error && error.message.includes("Error fetching Consensus update time"))
```
Wizdave97: Use tolowercase() for both message and substring

ddboy19912: Done

Wizdave97 (approved):

docs/content/developers/sdk/api/intent-gateway.mdx
```diff
@@ -753,6 +754,7 @@ type CancelEvent =
 |--------|--------|-------------|
 | `DESTINATION_FINALIZED` | `proof` | State proof from the destination chain is ready |
 | `AWAITING_CANCEL_TRANSACTION` | `data`, `to`, `value` | Caller must sign and submit the cancel transaction |
+| `RECOVERY_RESTART_REQUIRED` | — | A consensus update needed for recovery was pruned; begin recovery again |
```
seunlanlege: unfortunately i don't agree with this, why can't we restart the stream internally?

seunlanlege: Instead yield the AWAITING_CANCEL_TRANSACTION

ddboy19912: Done. Please check

sdk/packages/sdk/src/protocols/intents/OrderPlacer.ts
```diff
@@ -106,23 +139,17 @@ export class OrderPlacer {
...
 			address: sessionKeyAddress,
-			commitment: order.id as HexString,
+			commitment: finalizedOrder.id,
 			createdAt,
 		}
 
 		await this.ctx.sessionKeyStorage.setSessionKeyByAddress(sessionKeyAddress, sessionKeyData)
 
-		return { order, receipt }
+		return { order: finalizedOrder, receipt }
```
seunlanlege, line 153: Dropping the in-place mutation is the right call, but it's a breaking behavior change for existing consumers that read `order.id` / `order.nonce` / `order.session` off their own object after driving the generator. Our own tests do exactly that:

- `sdk/packages/simplex/src/tests/strategies/vault.testnet.test.ts` — builds the order without `id`, then calls `pollForOrderFilled(order.id as HexString, …)` after execution. `order.id` is now `undefined`.
- `sdk/packages/simplex/src/tests/strategies/fx.testnet.test.ts` — same pattern in two places, plus `sortBids(order, …)` passing the stale order.
- `docs/content/developers/evm/intent-gateway/placing-orders.mdx` still says "The SDK populates `order.id` with this commitment".

Can you update those two tests and the docs in this PR so the repo is consistent with the new contract (consumers take the finalized order from the `ORDER_PLACED` event)?

sdk/packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@hyperbridge/sdk",
-	"version": "2.4.2",
+	"version": "2.4.3",
```
seunlanlege: Given the `OrderPlacer` no-mutation change is behaviorally breaking for consumers relying on `order.id` being set in place (see comment there), this should be a **minor** bump, not a patch — `2.4.3` misrepresents the compatibility impact. A short migration note in the release/changelog would help too.

sdk/packages/sdk/src/storage/index.ts
```diff
@@ -108,12 +108,26 @@ export function createCancellationStorage(options: CancellationStorageOptions =
...
+	postCommitment: (orderId: string, source?: string, destination?: string) =>
+		`cancel-order:${cancellationScope(orderId, source, destination)}:postCommitment`,
 })
 
+/**
+ * An order id is not sufficient storage scope when the same id is observed on
+ * more than one chain pair. Keep the legacy unscoped shape when callers do not
+ * provide a pair so existing consumers can migrate incrementally.
+ */
+function cancellationScope(orderId: string, source?: string, destination?: string): string {
```
seunlanlege, line 146: `cancellationScope` embeds raw `order.source` / `order.destination`, but the same logical order legitimately circulates in two shapes — plain `"EVM-97"` (from the `OrderPlaced` event / canonical order) and hex `toHex("EVM-97")` (our own fx testnet test builds orders this way). Different shapes produce different keys, so resume silently misses the checkpoint and the old entry is orphaned.

Running both through `normalizeStateMachineId` here (or in `recoveryStorageKeys`) fixes the consistency issue and makes the keys readable.

sdk/packages/sdk/src/protocols/intents/OrderCanceller.ts
```diff
@@ -171,7 +184,7 @@ export class OrderCanceller {
 	 * @throws If the cancel transaction does not contain the expected on-chain event.
 	 */
 	private async *cancelOrderFromSource(order: Order, indexerClient: IsmpClient): AsyncGenerator<CancelEvent> {
-		const orderId = order.id!
+		const storageKeys = this.recoveryStorageKeys(order)
```
seunlanlege: Calling `recoveryStorageKeys(order)` here throws for id-less orders even on the same-chain path, which never touches storage (the contract computes the commitment itself) and previously worked with the old `order.id!`. Moving this below the same-chain block restores that and avoids computing keys (including `postCommitment`, unused in this method) before they're needed.

Related migration note: since the canceller now always reads/writes the pair-scoped keys, recovery state persisted by older SDK versions under `cancel-order:<id>:…` is never resumed and never cleaned — an in-flight cancellation across the upgrade will re-prompt for a new (paid) cancel transaction and leak the old entries. The legacy no-pair branch in `cancellationScope` is currently dead code within the SDK; consider a one-time fallback read of the legacy key instead.

sdk/packages/sdk/src/protocols/intents/OrderCanceller.ts
```diff
@@ -585,19 +624,86 @@ export class OrderCanceller {
...
 		} catch (error) {
+			if (MissingConsensusUpdateTimeError.isError(error)) throw error
 			const message = `Failed to deliver GET request to Hyperbridge; no response receipt found for ${commitment}: ${String(error)}`
 			this.logger.error(message)
 			throw new Error(message)
 		}
 	}
 
+	/** Verify that Hyperbridge still retains the consensus update for a proof. */
+	private async assertProofFresh(hyperbridge: SubstrateChain, proof: IProof): Promise<void> {
```
seunlanlege, line 671: `assertProofFresh` has no retry, and these probes are new failure points: a transient RPC/network error at any probe site now kills a cancellation stream that previously would have proceeded.

The other direction is worse: on the node side, `query_state_machine_update_time` does `.ok().flatten()`, so a runtime-API *error* surfaces with the same "Error fetching Consensus update time" message as a genuinely pruned update. A transient node problem then reads as "pruned", triggering checkpoint-clearing and a new paid cancel transaction via the restart loop.

Suggest wrapping the probe in `retryPromise` so transient failures get retried, with only the typed `MissingConsensusUpdateTimeError` passed through after retries are exhausted.

sdk/packages/sdk/src/tests/transformOrderForContract.test.ts
```diff
@@ -114,3 +117,62 @@ describe("transformOrderForContract", () => {
 		expect(outputs[0].token).toBe(ADDR_32)
 	})
 })
+
+describe("GET cancellation recovery cache", () => {
```
seunlanlege: Two things on coverage:

1. These are `OrderCanceller` tests — they'd be better in their own file than in `transformOrderForContract.test.ts`.
2. The riskiest change in this PR — `deriveCanonicalPlacedOrder`'s commitment parity with the contract — is exported but has no tests. A unit test mapping synthetic `OrderPlaced` event args to an expected commitment (covering the stamped-`user` and fee-reduced-`inputs` cases, and asserting the calldata fields are retained) would lock in exactly the bug this PR fixes.

seunlanlege (commented):
Reviewed both halves of this PR — the pruned-consensus recovery flow and the `OrderPlacer` canonical-order rework.

**Verified correct:**

- `deriveCanonicalPlacedOrder` matches what the contract hashes. `IntentGatewayV2.placeOrder` stamps `user` (bytes32 of `msg.sender`), `source` (`host()`), `nonce`, and replaces `inputs` with actual-received / protocol-fee-reduced amounts before `keccak256(abi.encode(order))` — and the event emits exactly those hashed values. The SDK retains only the two non-emitted calldata fields, `transformOrderForContract` strips the stale spread-in `id`, and the event's Solidity-string `source`/`destination` round-trip through `encodeStateMachineId`. This is a genuine fix: the old code only refreshed `nonce`/`inputs`, so client-side commitments diverged whenever the sender or fee reduction differed from what the client assumed.
- `executeAuto`'s ordering assumption holds — `execute` yields `ORDER_PLACED` before delegating to bid collection.
- The matched error string is unique to `query_state_machine_update_time` in `modules/pallets/ismp/rpc/src/lib.rs`, and `HttpRpcClient` surfaces it as `RPC error: <msg>`, so the substring match works. `waitForChallengePeriod` also calls `stateMachineUpdateTime`, so pruning mid-challenge-wait is correctly caught by the restart loop.
- Restart clears only the GET keys, not `postCommitment` — intentional and tested. New tests pass; the PR files typecheck clean.

**Main asks (details inline):**

1. The `OrderPlacer` no-mutation change breaks consumers that read `order.id` after `execute()` — including our own simplex testnet tests — and ships as a patch bump. Update those tests + docs and bump minor.
2. The restart loop in `cancelOrder` is unbounded and silent, and every restart makes the user sign and pay for a new cancel transaction. Needs a cap and a surfaced signal.
3. The pair-scoped storage keys don't normalize state-machine IDs, so the same logical order can produce different keys.

Smaller notes inline: the same-chain path now requires `order.id` unnecessarily, `assertProofFresh` conflates transient RPC failures with pruning, and `deriveCanonicalPlacedOrder` — the riskiest change here — has no test coverage.

One nit that has no inline anchor: `placeOrder`'s jsdoc now says the order "is not mutated", which is true there, but `IntentGateway.execute` still mutates `order.fees` in place, so callers shouldn't infer full immutability from the lifecycle.

seunlanlege (approved):

sdk/packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # @hyperbridge/sdk
 
+## 2.5.0
+
+### Minor Changes
+
+- `ORDER_PLACED` now returns the finalized canonical order without changing the order object passed to `execute()` or `placeOrder()`. Store and use `update.order` for the commitment, nonce, and session key; do not read those fields from the submitted object after placement.
+- Cancellation recovery now reports `RECOVERY_RESTARTED` when Hyperbridge has pruned a required consensus update. Use `maxRecoveryRestarts` to control how many automatic recovery attempts are allowed.
```
seunlanlege: well this is no longer true, better to remove

ddboy19912: Got it

PR #1075, opened 2026-07-28
[simplex, docs]: self-funding testnet suite; document the current faucet generation

seunlanlege:
The fx.testnet CI wallet ran dry on BSC Chapel (#1065's red `test` job) and the suite died at `placeOrder` with `ERC20: transfer amount exceeds balance`. Root-causing required on-chain archaeology because **every faucet in the docs is a dead generation**: the current ERC6160 USD.h (`0xA801da10…5177`) is mintable only by the undocumented third-generation TokenFaucet at `0x1794aB22…C3a3`, CREATE2-deployed at the same address on Sepolia, BSC Chapel, Base/Arbitrum/OP Sepolia and Gnosis Chiado (each verified on-chain).

**Self-funding suite**: `fx.testnet.test.ts` gains a `beforeAll` that drips 1000 USD.h from the faucet whenever the CI wallet's Chapel balance falls below ~100 runs' worth. The faucet's once-daily cooldown is tolerated while the balance still covers a run; when it can't, the failure names the wallet, faucet, and token — no more opaque ERC20 reverts. This class of CI outage is permanently closed.

**Docs**: the Gargantua V3 tables now carry the live TokenFaucet/FeeToken pair for the five verified chains (Amoy verifiably still runs the older generation and its rows stand; Paseo Asset Hub could not be verified and stands as-is), and `testnet-faucet.mdx` points at the working faucet + token with its typos fixed.


seunlanlege:
Folding into #1065 so its red testnet job picks up the fix directly.

PR #1076, opened 2026-07-28
[runtime, tesseract]: give mandatory proofs their own offchain namespace

dharjeezy:
Mandatory proofs shared an offchain key with messaging proofs, they now sit under a rotation prefix keyed by the set id they rotated to, and the relayer and SDK pick the namespace from the accepted proof's set  id rather than always deriving it from height.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -415,7 +415,7 @@ jobs:
             - name: Run Simnet Tests
               run: |
                   cargo build -p hyperbridge --release
-                  ./target/release/hyperbridge simnode --chain=gargantua-4009 --name=alice --tmp  --state-pruning=archive --blocks-pruning=archive --rpc-port=9990 --port 40337 --log="mmr=trace" --rpc-cors=all --unsafe-rpc-external --rpc-methods=unsafe --pool-type=single-state &
+                  ./target/release/hyperbridge simnode --chain=gargantua-4009 --name=alice --tmp  --state-pruning=archive --blocks-pruning=archive --rpc-port=9990 --port 40337 --log="mmr=trace" --rpc-cors=all --unsafe-rpc-external --rpc-methods=unsafe --pool-type=single-state --enable-offchain-indexing true &
                   ./scripts/wait_for_tcp_port_opening.sh localhost 9990
                   cargo test -p simtests -- --nocapture --ignored --test-threads=1 --skip test_runtime_upgrade_and_fee_migration --skip legacy_storage_items_state_drain_test
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -415,7 +415,7 @@ jobs:
             - name: Run Simnet Tests
               run: |
                   cargo build -p hyperbridge --release
-                  ./target/release/hyperbridge simnode --chain=gargantua-4009 --name=alice --tmp  --state-pruning=archive --blocks-pruning=archive --rpc-port=9990 --port 40337 --log="mmr=trace" --rpc-cors=all --unsafe-rpc-external --rpc-methods=unsafe --pool-type=single-state --enable-offchain-indexing true &
+                  ./target/release/hyperbridge simnode --chain=gargantua-4009 --name=alice --tmp  --state-pruning=archive --blocks-pruning=archive --rpc-port=9990 --port 40337 --log="mmr=trace" --rpc-cors=all --unsafe-rpc-external --rpc-methods=unsafe --pool-type=single-state &
                   ./scripts/wait_for_tcp_port_opening.sh localhost 9990
                   cargo test -p simtests -- --nocapture --ignored --test-threads=1 --skip test_runtime_upgrade_and_fee_migration --skip legacy_storage_items_state_drain_test
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24437,6 +24437,7 @@ dependencies = [
  "mmr-primitives",
  "mockito",
  "nexus-runtime",
+ "pallet-beefy-consensus-proofs",
  "pallet-intents-coprocessor",
  "pallet-intents-rpc",
  "pallet-ismp",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15589,6 +15589,7 @@ dependencies = [
  "log",
  "mmr-primitives",
  "pallet-bandwidth",
+ "pallet-beefy-consensus-proofs",
  "pallet-call-decompressor",
  "pallet-collator-manager",
  "pallet-consensus-incentives",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28392,7 +28392,7 @@ checksum = "8f50febec83f5ee1df3015341d8bd429f2d1cc62bcba7ea2076759d315084683"
 
 [[package]]
 name = "tesseract"
-version = "2.4.9"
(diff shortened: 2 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -415,7 +415,7 @@ jobs:
             - name: Run Simnet Tests
               run: |
                   cargo build -p hyperbridge --release
-                  ./target/release/hyperbridge simnode --chain=gargantua-4009 --name=alice --tmp  --state-pruning=archive --blocks-pruning=archive --rpc-port=9990 --port 40337 --log="mmr=trace" --rpc-cors=all --unsafe-rpc-external --rpc-methods=unsafe --pool-type=single-state &
+                  ./target/release/hyperbridge simnode --chain=gargantua-4009 --name=alice --tmp  --state-pruning=archive --blocks-pruning=archive --rpc-port=9990 --port 40337 --log="mmr=trace" --rpc-cors=all --unsafe-rpc-external --rpc-methods=unsafe --pool-type=single-state --enable-offchain-indexing true &
```
Wizdave97: Offchain indexing is enabled by default on hyperbridge node

Wizdave97:
Bump tesseract's patch version

Wizdave97:
## Blocking

**1. A rotation proof leaves a hole at `latest_height`; two readers still assume "messaging" for it — regression**

`lib.rs:547-575` is an exclusive if/else: a rotating proof writes *only* the rotation namespace and pushes nothing into `MessagingProofs` — but it still advances the ISMP proven height. So after a rotation, `latest_proven_height == H_rot` while `messaging_offchain_key(H_rot)` was never written.

Two readers build a messaging key from exactly that height:
- [[fees.rs:276](https://claude.ai/epitaxy/tesseract/messaging/messaging/src/fees.rs:276)](tesseract/messaging/messaging/src/fees.rs:276) — `let mut proof_key = ProofKey::Messaging(latest_height)`. The wait-loop correctly picks `Rotation` from `event.new_set_id`, but this **default is used verbatim when `max_block <= latest_height`** and the loop is skipped.
- [[substrate.ts:505](https://claude.ai/epitaxy/sdk/packages/sdk/src/chains/substrate.ts:505)](sdk/packages/sdk/src/chains/substrate.ts:505) — `messagingOffchainKey(lastProvenHeight)`, then `return undefined` on miss.

On base both wrote and read `offchain_key(height)`, so the blob was always found — this is introduced here. `fees.rs` retries forever on the same missing key; the SDK can't build the batch at all. Not reliably self-healing: a new messaging proof requires `child_trie_root != LastRewardedDispatchRoot`, and the rotation already bumped that. Fix: on miss, consult `RotationProofs` for the entry whose value equals the height and use `rotation_offchain_key(set_id)`.

**2. `UncleSlotsFull` can revert an applied rotation — permanent consensus halt**

`settle_first_proof` calls `record_uncle_metadata(outcome.latest_height, …)?` for every SP1 first proof, rotation included ([[lib.rs:537-540](https://claude.ai/epitaxy/modules/pallets/beefy-consensus-proofs/src/lib.rs:537)](modules/pallets/beefy-consensus-proofs/src/lib.rs:537)). That does `try_push` into `AcceptedProvers[height]`, bounded at `MaxUncleProvers + 1` = **6** (`MaxBeefyUncleProvers = 5` in both gargantua and nexus), returning `UncleSlotsFull` on overflow.

If a messaging proof at `H` already drew its full complement of 6 provers and the session then rolls over with the parachain stalled, the mandatory proof lands at the same `H`, `verify_and_apply` applies the rotation, and the `?` then reverts the whole dispatch. The author's own comment condemns exactly this:

> Failing here is a dispatch error, so it would roll back the authority-set rotation `handle_incoming_message` just applied, pinning the consensus state on the old set forever — the mandatory-block justification is the only one a prover can obtain for that session, so every retry fails identically.

They guarded the `StaleProof` path against this and missed `UncleSlotsFull` on the same dispatch. **Pre-existing rather than introduced** — but it's the same height-collision class this PR exists to fix, so it belongs here. Either skip uncle metadata when `outcome.rotated`, or make the push non-fatal.

## Should fix

**3. The SDK rotation walk starts two epochs too low, on an inverted comment.** [[substrate.ts:489-495](https://claude.ai/epitaxy/sdk/packages/sdk/src/chains/substrate.ts:489)](sdk/packages/sdk/src/chains/substrate.ts:489) says the proof is signed by set `key + 1` and `currentEpoch` is `nextAuthoritySet.id`. Both are backwards: the map key is the set rotated *to* (so signed by `key − 1`), and `HandlerV2.sol:170` stores `_currentEpoch = nextAuthoritySetId - 1`, making `currentEpoch` the destination's *current* set. The first key needed is `currentEpoch + 1n`. The Rust reader gets this right (`provider.rs:111-124`). Usually harmless (stale proofs short-circuit on EVM), but if `rotationMap` lacks `currentEpoch - 1` the `while` never enters, zero rotation proofs are returned, and the messaging proof reverts with `UnknownAuthoritySet`.

**4. The migration fallback exists only in Rust, and only on one path.** `provider.rs:127` deliberately falls back to the old key (byte-identical to the new messaging key) — correct. But the SDK has none, and it hits the *oldest* (most likely pre-upgrade) entries first. `outbound.rs:172-184` also has no fallback and `continue`s past the notification. Note the fallback can also return the *wrong* blob for a pre-upgrade entry whose height collided — unavoidable, but `provider.rs:120-122` reads as though it's always correct.

**5. The anti-alias test is vacuous — demonstrated.** Setting `ROTATION_OFFCHAIN_PREFIX` byte-identical to `MESSAGING_OFFCHAIN_PREFIX` (total aliasing, the exact thing the test names) still passes. All three assertions hold by construction of `offchain_key(prefix, id)`. Use `assert_ne!(messaging_offchain_key(x), rotation_offchain_key(y))`. Separately, `keys_match_the_sdk_encoding` never touches TypeScript — it's a Rust-side change detector, and there's no TS test for the key builders at all.

**6. The simtests only cover the happy path.** The rotation branch depends on the live Paseo relay happening to cross a session boundary, so most CI runs never execute it; and even then the collision is never constructed (no messaging proof at that height first), so `assert_ne!(aliased, Some(&wire_proof))` passes vacuously against `None`. This would not have caught the original bug, nor a regression. The pallet has no `mock.rs`/`tests.rs` — a mock-runtime test driving two proofs at one height would cover this plus findings 1 and 2 deterministically.

## Verified correct

Set-id semantics are right end to end: a proof signed by set N yields `current_set_id == N+1`, and the offchain key, the `RotationProofs` map key, and every Rust reader use that same value — no off-by-one, no use of the signing set id. Cross-language encoding is exact: prefixes match byte-for-byte including trailing `::`, big-endian on both sides, `bigint` so no 2^53 loss, `PERSISTENT` everywhere matching `offchain_index::set`. Eviction now clears the correct namespace. The `keys().next().copied()` refactor is equivalent. The CI change adding `--enable-offchain-indexing true` is load-bearing — without it `offchain_index::set` is a no-op, which means the pre-existing simtests weren't exercising this path at all.

One nit: `lib.rs:188` still says "`iter().next()` gives FIFO eviction" after the switch to `keys().next()`.

Want me to draft the fixes for findings 1 and 2?

Wizdave97 (approved):

Wizdave97 (approved):

PR #1079, opened 2026-07-29
[runtime]: Turn off dispatchable xcm execute on both runtimes

dharjeezy:
`XcmExecuteFilter` was set to `Everything` on nexus and gargantua, which let any signed account run a local XCM program. That is not safe for restricted origins, because a filter attached to an origin is dropped once the origin is reduced to a location and `Transact` rebuilds a plain signed origin from it. Both runtimes now set `Nothing`

Nothing depends on it, no `Transact` instruction is constructed anywhere in the workspace.


seunlanlege:
Ahh classic xcm

Wizdave97 (approved):

PR #1086, opened 2026-07-30
[runtime, sdk, indexer, simplex]: bundle phantom pairs into one order and index pair-centric liquidity pools

### LINKED ISSUE #1083: [runtime] Phantom order improvements
Instead of generating a distinct phantom order for each pair, let's just add all the pairs to the list of assets in one phantom order.

indexer and simplex should be updated to match

### LINKED ISSUE #1084: [indexer]: Pair-centric liquidity pools for phantom order data
## Background

The indexer currently records phantom order data in a solver-centric shape:

- `PhantomOrderPriceSnapshot` — one row per commitment per bid-window close, carrying the liquidity-weighted median output amount for a single *direction* (`tokenA → tokenB`), keyed by raw EVM token addresses (`sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts`).
- `LiquidityProvider` / `LiquidityProviderBalance` — per-solver, per-(chain, token) balances swept at each snapshot.

This is what the HyperFX "liquidity" page renders today: a list of liquidity **providers**. There is no pair-level identity anywhere in the schema — querying "cNGN/USDC" means knowing the token addresses on a specific chain and filtering snapshots by `tokenA`/`tokenB`, once per direction, then doing the aggregation client-side.

## What we want instead

The product surface should be **liquidity pools**, one per pair (e.g. `cNGN/USDC`), with providers one level down. Each pool is queryable individually and carries:

1. **Total liquidity available to be swapped** — aggregated across all solvers quoting the pair, per side (token0 inventory + token1 inventory).
2. **Buy & sell average prices, weighted by liquidity** — a pair has two phantom order directions (`cNGN → USDC` prices the sell side, `USDC → cNGN` prices the buy side). Each direction already produces a liquidity-weighted median per snapshot; the pool surfaces both under one entity. A pair has **one price across chains** per direction — solvers rebalance and arbitrage away per-chain drift — so per-chain snapshots are samples of the same market price, merged liquidity-weighted into the pool's single buy and sell numbers.

### Drill-downs

- **Pool page → chain distribution**: how the pool's liquidity is spread across chains (per-(pool, chain) aggregate, per side).
- **Pool page → providers**: the solvers backing the pool, with their contribution, linking to…
- **Provider page**: the provider's full inventory (every token, every chain — the latest `LiquidityProviderBalance` rows already give us this), plus their yield-vault positions and yields (`VaultLpPosition` / `VaultPositionSnapshot`).

## Proposed schema changes

Compose pairs from a top-level token registry rather than baking addresses into entities (today `fx-tokens.addresses.ts` is a single hardcoded Base cNGN entry):

- **Token registry**: symbol + decimals + per-chain addresses, so `(chain, address)` resolves to a canonical symbol. Pairs are then symbol-level: `cNGN/USDC` is one pool regardless of which chain a snapshot came from.
- **`LiquidityPool`** entity — `id` = canonical pair slug (e.g. `cNGN-USDC`), with per-side total liquidity, `buyPrice`, `sellPrice`, bid counts, and last-updated block/time. Upserted by `handlePhantomOrderPrices` whenever a snapshot for either direction lands.
- **`PoolChainLiquidity`** — `id` = `{pool}-{chain}`, per-chain, per-side totals for the chain-distribution view.
- **Pair reference on `PhantomOrderPriceSnapshot`** (pool FK + direction flag) so price/liquidity history is queryable per pair directly instead of by address filtering.
- **Provider ↔ pool link** — either a join entity or a pool FK on the balance rows, so a pool can list its providers and a provider can list its pools.

## Cross-chain stablecoin routes

Phantom orders are always same-chain orders — both tokens live on the order's chain. To discover and price cross-chain stablecoin routes (`USDC@BSC → USDT@Base`, `USDT@Polygon → USDC@Ethereum`, …), each chain's phantom order includes the stablecoin pairs alongside the FX pairs: `USDC/USDC`, `USDT/USDT`, and `USDC/USDT` (both directions).

What each same-chain pool measures:

- **Same-asset pools** (`USDC → USDC` on X): solver presence, inventory, and baseline margin + settlement-latency carry. The solver's net inventory change on X is zero, so this rate carries no chain premium by construction.
- **Cross-asset pools** (`USDC → USDT` on X): samples of the pair's price. A pair prices identically across chains — solvers rebalance via CCTP/USDT0 and arbitrage away per-chain drift — so each chain's snapshot is a sample of the same market price. Chains differ in liquidity and source acceptance, never in price.

Route semantics:

- **Liquidity**: filling a cross-chain order only requires delivering the output token on the destination — the solver *receives* the input on the source and needs nothing there. So liquidity for any route `src.tokenIn → dst.tokenOut` = bidding solvers' `tokenOut` inventory on `dst`, which the per-chain pools already measure.
- **Price**: the pool's single pair price for that direction, regardless of route. Source acceptance (below) is binary: a solver that cannot profitably accept payment from some chain at the pair price (e.g. BSC, which has neither CCTP nor USDT0) simply omits that chain, so uncovered-boundary costs show up as thinner route liquidity rather than a price component.

### Accepted source chains in `paymasterAndData`

A same-chain bid proves a solver operates on a chain, not which chains they will accept payment *from*. Bids therefore declare their accepted source chains inside the bid user op's `paymasterAndData`:

- `paymasterAndData` is covered by the userOpHash, so the declaration is authenticated by the solver's existing bid signature. (The `signature` field is excluded from the userOpHash and therefore unsuitable — data there would be unauthenticated.)
- **Format**: a version byte followed by a list of accepted source `stateMachineId`s. Acceptance is binary — the bid's output asset and rate already say what will be provided; there is no per-source price adjustment.
- **Default**: an absent/legacy declaration means "accepts all CCTP/USDT0-covered chains", so existing solvers keep working and only uncovered-chain acceptance requires the new encoding.
- **Scope**: the overload applies to phantom bids only; real fill user ops keep their functional `SimplexPaymaster` semantics.

No pallet changes are needed for the declaration itself: `Bid.user_op` is already opaque bytes to the pallet, and bid verification lives in the indexer. The pallet config only gains the stablecoin pair entries; simplex encodes the declaration from chain config it already has; the indexer decodes and indexes it during aggregation.

**Schema impact**: a derived route view — route `src → dst` is live iff the destination pool has a verified bid backed by inventory whose declaration includes `src`.

## Caveats / open questions

- The liquidity sweep measures a solver's *global* token balances, not per-pair earmarked inventory. The same USDC balance backs every USDC pool the solver quotes, so pool totals are "liquidity reachable by this pair", and summing across pools double-counts. Fine for display, but worth stating in the schema docs.
- Buy & sell requires both directions of a pair to be registered in the pallet's phantom order config. Today `PhantomTokenPair` is directional — a pair configured one-way yields a pool with only one of the two prices.
- Pool liquidity is inventory-inferred, an upper bound on depth: a bid quotes the standard amount, while the solver's real fill appetite is capped by their strategy's `maxOrderSize`, which the bid doesn't disclose. If quote-backed depth is wanted later, the bid payload can carry a committed max size per pair.
- This builds on #1083's single-phantom-order-per-chain model: one order per chain carries all pairs, bids quote a subset of the order's pairs (a rate vector, since a solver only serves some pairs), and the snapshot handler fans out per pair — which is what the pair-centric schema wants anyway.

dharjeezy:
The phantom generator now emits one order per interval carrying every configured token pair in its asset lists, instead of one order, commitment and event per pair, so pair i is inputs[i] against output.assets[i] and CurrentPhantomOrder collapses to a single entry with a migration to clear the old one. The solver quotes each leg independently by narrowing the bundled order per pair, and the indexer prices every leg rather than reading outputs[0] and discarding the rest.  closes #1083

The indexer then turns the per-leg snapshots into pair-centric liquidity pools: a token registry resolves each leg to a canonical pair slug, and LiquidityPool, PoolChainLiquidity and PoolBidder track each direction's price and depth merged across chains. Solvers declare the source chains they accept payment from inside their phantom bids' paymasterAndData, which the bid signature already covers, and the aggregation surfaces the declaration per bidder for route availability. FX USD pricing now reads the pool rate instead of address-filtering snapshots. closes #1084

Per review, the generator now probes both directions of every configured pair: PhantomTokenPair gained standard_amount_b and each pair expands into a forward and a reverse leg, so buy and sell sides come from one config entry. The modified indexer entities were renamed (PhantomOrderV2, PhantomOrderLeg, PhantomOrderPriceSnapshotV2) so deployments create fresh tables. The storage migration clears the pre-upgrade phantom order, config and generation marker — governance must re-set the config with the new standard_amount_b field — and commitments registered before the upgrade never emit PhantomBidWindowExhausted, so their final snapshot is skipped. All pallet weights are now measured, including benchmarks for the previously hand-stubbed governance calls.


Wizdave97:
We should do this inn tjis PR also

https://github.com/polytope-labs/hyperbridge/issues/1084

modules/pallets/intents-coprocessor/src/types.rs
```diff
@@ -415,16 +420,22 @@ pub fn phantom_order_commitment(
 		fees: AlloyU256::ZERO,
 		session: alloy_primitives::Address::ZERO,
 		predispatch: sol_types::DispatchInfo { assets: vec![], call: Bytes::new() },
-		inputs: vec![sol_types::TokenInfo {
-			token: FixedBytes::from(token_a_bytes),
-			amount: AlloyU256::from(standard_amount),
-		}],
+		inputs: pairs
```
Wizdave97: We also want the reverse case input: pair.token_b, output: pair.token_a

sdk/packages/indexer/src/configs/schema.graphql
```diff
@@ -2196,8 +2196,11 @@ type BandwidthTier @entity {
...
 Phantom orders are synthetic benchmark orders that the pallet creates at the start of each
 interval (governed by PhantomBidWindow) to solicit bids from fillers and measure real
-market liquidity and pricing for a given token pair on a destination chain.
+market liquidity and pricing on a destination chain.
 The commitment is keccak256(abi.encode(Order)) and serves as the unique identifier.
+
+Every configured token pair rides in one order, so the pairs live in PhantomOrderPair rather
+than on the order itself.
 """
 type PhantomOrder @entity {
```
Wizdave97: You'll need to change the name of any entities that are modified

Wizdave97:
## Verdict

Both issues are substantially addressed and the core design is sound. The load-bearing safety property — that a solver can't reorder the asset list to attach its quote to the wrong pair — is genuinely enforced: [[phantom-aggregation.ts](https://claude.ai/epitaxy/sdk/packages/sdk/src/protocols/intents/phantom-aggregation.ts)](sdk/packages/sdk/src/protocols/intents/phantom-aggregation.ts) verifies `orderCommitment(fillData.order) === commitment` *before* reading `fillData.legs`, so positional indexing is bound to the pallet's own commitment. The rate normalization is also correct and chain-independent: `resolvePoolLeg` enforces `standardAmount == 10^inputDecimals` and rejects `outDecimals > 18`, so `medianPrice * 10^(18-outDecimals)` really is "output whole units per 1 whole input, 1e18-scaled". That's what makes the cross-chain merge and the FX USD pricing rewrite valid.

Issue #1083 is fully covered. Issue #1084 is covered except the **derived route view** it lists under "Schema impact" — there's no route entity; consumers must join `PoolBidder` themselves. The schema docs describe the derivation, so this may be deliberate, but it's a stated deliverable worth confirming.

## Blocking

**1. Per-leg balance RPC amplification** — [[phantom-aggregation.ts:553](https://claude.ai/epitaxy/sdk/packages/sdk/src/protocols/intents/phantom-aggregation.ts:553)](sdk/packages/sdk/src/protocols/intents/phantom-aggregation.ts:553)

`getTotalSolverBalance` moved *inside* the leg loop, so it's now O(bids × legs) **sequential** `eth_call`s instead of O(bids), plus a `maxWithdraw` per configured vault. At `MAX_PHANTOM_TOKEN_PAIRS = 64` and N verified bids that's 64·N round trips inside a SubQuery block handler, on top of the existing per-bid `sweepSolverLiquidity`. Distinct output tokens across 64 legs are only a handful (USDC/USDT/cNGN), so memoizing by `(chain, token, solver)` for the life of the aggregation collapses this back to near the old cost.

**2. Sequential per-leg quoting vs. the bid window** — [[filler.ts:850](https://claude.ai/epitaxy/sdk/packages/simplex/src/core/filler.ts:850)](sdk/packages/simplex/src/core/filler.ts:850)

`handlePhantomOrder` awaits `quotePhantomLeg` one leg at a time, and each leg runs `canFill` + `quotePhantomFill` per strategy — each doing `getTokenDecimals`, venue price, and sizing RPC. The bid must land within `PhantomBidWindow`: **5 blocks on gargantua**, 25 on nexus ([[gargantua/src/ismp.rs:108](https://claude.ai/epitaxy/parachain/runtimes/gargantua/src/ismp.rs:108)](parachain/runtimes/gargantua/src/ismp.rs:108)). With up to 64 legs this can miss the window entirely, and the failure mode is a silent no-bid. Legs are independent — `Promise.all` (or bounded concurrency) fixes it. Separately, `FXFiller.quotePhantomFill` already calls `canFill` internally, so `quotePhantomLeg`'s gate doubles that work.

**3. Unmeasured weight on a mandatory hook** — [[gargantua weights:181](https://claude.ai/epitaxy/parachain/runtimes/gargantua/src/weights/pallet_intents_coprocessor.rs:181)](parachain/runtimes/gargantua/src/weights/pallet_intents_coprocessor.rs:181)

`generate_phantom_order` is labelled "Placeholder, not measured" in both runtime weight files, yet `on_initialize` now returns it. The hook ABI-encodes and keccaks up to 64 legs and deposits a ~3.6 KB event; the placeholder claims 30 µs + 0.2 µs/pair. `scripts/benchmarking.sh` was updated to include the pallet, so this is a "run the box before merge" item rather than a design problem — but shipping an unmeasured mandatory-hook weight shouldn't land.

## Should fix

**4. `POOL_ROWS_LIMIT = 100` under-bounds the bidder set-diff** — [[liquidityPool.service.ts:624](https://claude.ai/epitaxy/sdk/packages/indexer/src/services/liquidityPool.service.ts:624)](sdk/packages/indexer/src/services/liquidityPool.service.ts:624)

The comment justifies 100 via "registered pairs are pallet-bounded at 64", but `PoolBidder` rows per `(pool, chain)` scale with the number of **solvers** (× direction × output-token representation), which is unbounded. Past 100, `existing` truncates and stale rows are never removed — and `PoolBidder` is exactly what route availability reads, so a solver that stopped bidding keeps advertising capacity. Page until exhausted, or tie the limit to the real bound.

**5. Token registry omits already-configured FX assets** — [[pool-tokens.addresses.ts](https://claude.ai/epitaxy/sdk/packages/indexer/src/addresses/pool-tokens.addresses.ts)](sdk/packages/indexer/src/addresses/pool-tokens.addresses.ts)

`POOL_TOKENS` has USDC/USDT/cNGN but not ZARP or EURC, which are already in `chain.ts` assets for Ethereum, Base and Polygon ([[chain.ts:335](https://claude.ai/epitaxy/sdk/packages/sdk/src/configs/chain.ts:335)](sdk/packages/sdk/src/configs/chain.ts:335), [[:521](https://claude.ai/epitaxy/sdk/packages/sdk/src/configs/chain.ts:521)](sdk/packages/sdk/src/configs/chain.ts:521), [[:587](https://claude.ai/epitaxy/sdk/packages/sdk/src/configs/chain.ts:587)](sdk/packages/sdk/src/configs/chain.ts:587)). A ZARP or EURC pair yields `pool = null` and never creates a `LiquidityPool` row, and the only signal is a `logger.debug` — the pool silently doesn't exist. Since the registry is the thing #1084 asked for, it should cover every FX asset a filler can be configured for.

**6. Schema is inconsistent about whether a reindex is required** — [[schema.graphql:2345](https://claude.ai/epitaxy/sdk/packages/indexer/src/configs/schema.graphql:2345)](sdk/packages/indexer/src/configs/schema.graphql:2345)

`PhantomOrderPriceSnapshot.chain` is documented "Null on rows indexed before pools existed", implying old rows survive. But `pair: PhantomOrderPair!` is non-null and `PhantomOrder` dropped `tokenA`/`tokenB`/`standardAmount`, so they can't. Either make `chain` non-null and drop the caveat, or state plainly that this release requires a full reindex.

**7. `set_phantom_order_config` doesn't validate `standard_amount`** — [[lib.rs:583](https://claude.ai/epitaxy/modules/pallets/intents-coprocessor/src/lib.rs:583)](modules/pallets/intents-coprocessor/src/lib.rs:583)

The type doc is emphatic it must be `10^decimals(token_a)`, and the indexer now *enforces* it — a mismatch silently kills pool attribution for that pair. The pallet can't know decimals, but rejecting `standard_amount == 0` is free and the new empty/duplicate checks are right next door.

## Nits

- [[handlePhantomOrderPrices.handler.ts:181](https://claude.ai/epitaxy/sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts:181)](sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts:181): `pairsByIndex` and `registeredByIndex` index the same rows, and the code checks one then non-null-asserts the other (`registeredByIndex.get(leg.pairIndex)!`). One map carrying both shapes removes the assertion.
- `sortPoolSymbols` uses `localeCompare`, which is locale-sensitive — for a persisted primary key, plain `<`/`>` on the lowercased strings is more obviously deterministic.
- `quotePhantomLeg` writes `setFillerOutputs`/`setPairClassifications` under synthetic ids `${commitment}-${index}`; nothing reads them, so each interval churns up to 64 entries through the TTL cache the real-fill path shares.
- `blockNumber - row.lastUpdatedBlock` goes negative on out-of-order replay and passes the `<= MAX_SAMPLE_AGE_BLOCKS` freshness test; `PoolChainLiquidity` is also overwritten unconditionally. Fine for in-order indexing, but the file's header claims replay convergence.
- `MAX_SAMPLE_AGE_BLOCKS = 1800n` is described as "a small multiple of the phantom generation interval", but that interval is governance-configurable — a long interval makes every sample stale and the merge silently falls back to `directionRows`.
- The migration clears `CurrentPhantomOrder` but not `LastPhantomGeneration`, so pre-migration commitments never get their `PhantomBidWindowExhausted` and lose their final snapshot. Almost certainly acceptable — worth a line in the PR description.
- Same-asset pools (`USDC/USDC`, which #1084 explicitly wants) always resolve to `SELL`, so `buyRate` is permanently null and `token0Symbol === token1Symbol`. Coherent, but the consumer needs to know the baseline-margin measurement lands on `sellRate` only.

## Confirmed good

- `decodeAcceptedSourceChains` correctly distinguishes absent (`null`) from explicit-empty (`[]`), and the strict trailing-byte check rejects non-declaration blobs that happen to start with `0x01`. The `paymasterAndData` placement is genuinely authenticated via userOpHash, as the issue specified.
- The duplicate-direction rejection in `set_phantom_order_config` is the right guard now that legs share one snapshot.
- `phantom_order_body_pairs_legs_positionally` tests something a commitment-equality assertion can't catch (any consistent encoding hashes consistently), and `phantom_order_at_max_pairs_stays_within_budget` pins the 64-pair worst case. The indexer test file covers the pool pipeline well, including bidder removal and the unquoted-pair path.

modules/pallets/intents-coprocessor/src/types.rs
royvardhan: Done — the generator expands every configured pair into its forward and reverse legs. PhantomTokenPair gained standard_amount_b to denominate the reverse leg, pair uniqueness is now on the unordered token set, zero standard amounts are rejected, and the migration clears the stored config so governance re-sets it with the new field.

sdk/packages/indexer/src/configs/schema.graphql
royvardhan: Done — PhantomOrderV2, PhantomOrderLeg and PhantomOrderPriceSnapshotV2, and the snapshot's chain field is non-null now that the tables are fresh.

royvardhan:
All addressed: balance reads are memoized per (chain, token, solver) for the whole aggregation run and shared with the liquidity sweep; phantom legs are quoted concurrently and the duplicate canFill is gone; the PoolBidder reconciliation pages until exhausted; the registry now covers every FX asset in chain.ts (DAI, ZARP, EURC, XSGD, TRYB — decimals read from each token contract on its own chain); set_phantom_order_config rejects zero standard amounts; the schema inconsistency is resolved by the entity renames. Nits taken too: one map in the price handler, code-unit symbol sorting, no cache churn from phantom quoting, staleness and replay assumptions documented, the same-asset-lands-on-sellRate note added, and the migration caveat is in the PR description. Benchmarks are left for the box — scripts/benchmarking.sh already includes the pallet; the placeholder's per-pair term was doubled meanwhile since each pair now encodes two legs. On the route view: per-bidder rows instead of materialized per-(src, dst) rows is deliberate — any source filter stays computable client-side without the row fan-out, and the schema docs describe the derivation.

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -149,15 +154,11 @@ pub mod pallet {
...
-	/// rather than only the last one generated. Replaced as a whole each cycle.
+	/// The phantom order active for the current interval. Every configured token pair rides
+	/// in this one order, so a single commitment covers them all. Replaced each cycle.
 	#[pallet::storage]
-	pub type CurrentPhantomOrder<T: Config> = StorageValue<
-		_,
-		BoundedVec<(H256, PhantomOrderInfo<BlockNumberFor<T>>), ConstU32<MAX_PHANTOM_TOKEN_PAIRS>>,
-		OptionQuery,
-	>;
+	pub type CurrentPhantomOrder<T: Config> =
```
Wizdave97: This should remain a BoundedVec, we are going to emit phantom orders for different chains where hyperfx is deployed so the indexer has an exact value for active liquidity available on all chains.

modules/pallets/intents-coprocessor/src/lib.rs
```diff
@@ -287,10 +296,9 @@ pub mod pallet {
 			ensure!(!user_op.is_empty(), Error::<T>::InvalidUserOp);
 
 			// Phantom orders have stricter rules: one bid per filler, no updates, and only
-			// within the configured acceptance window after the order was registered. Every
-			// active pair is checked, not just the most recently generated one.
-			if let Some(active) = CurrentPhantomOrder::<T>::get() {
-				if let Some((_, info)) = active.iter().find(|(c, _)| *c == commitment) {
+			// within the configured acceptance window after the order was registered.
+			if let Some((active, info)) = CurrentPhantomOrder::<T>::get() {
```
Wizdave97: We should revert this change

sdk/packages/simplex/src/tests/phantom-filler.e2e.simnode.test.ts
```diff
@@ -134,36 +137,24 @@ async function seedStateMachineHeight(api: ApiPromise, chainId: number, height:
 async function setPhantomOrderConfig(api: ApiPromise): Promise<void> {
```
Wizdave97, line 183: Let's set phantom orders on multiple chains with multiple two token pairs each and let the test enforce correct liquidity and route indexing.

sdk/packages/simplex/src/services/ContractInteractionService.ts
```diff
@@ -643,12 +644,18 @@ export class ContractInteractionService {
...
+	 * paymasterAndData so the userOpHash — and therefore the solver's bid signature —
+	 * covers it. Phantom bids never reach a bundler or the EntryPoint, so the field is
+	 * free for this; a real fill's paymasterAndData keeps its functional semantics.
 	 */
 	async preparePhantomBidUserOp(
 		order: Order,
 		entryPointAddress: HexString,
 		solverAccountAddress: HexString,
 		fillerOutputs: TokenInfo[],
+		acceptedSourceChains?: string[],
```
Wizdave97, line 658: acceptedSourceChains is not injected at the call side of this function

Wizdave97:
Finding | Status
-- | --
RPC amplification per leg | ✅ memoizedSolverBalance caches by (chain, token, solver) and is shared with sweepSolverLiquidity; quoted legs now fetch via Promise.all
Sequential leg quoting vs bid window | ✅ Promise.all over legs, redundant canFill dropped, and the cache-pollution nit fixed (id: undefined + setFillerOutputs removed from quotePhantomFill)
POOL_ROWS_LIMIT truncation | ✅ readAllPages in store.helpers.ts with stable orderBy: "id"
Incomplete token registry | ✅ DAI, ZARP, EURC, XSGD, TRYB added across chains
Schema nullability inconsistency | ✅ Entities renamed to PhantomOrderV2 / PhantomOrderLeg / PhantomOrderPriceSnapshotV2; chain is now String!
Unmeasured generate_phantom_order weight | ❌ Still "Placeholder, not measured" in both runtime weight files — coefficient doubled to 400k/pair for the two legs, but still a guess on a mandatory hook


<p>Plus <code>ZeroPhantomStandardAmount</code>, the <code>localeCompare</code> → code-unit fix, and the <code>weightedRate</code> / <code>AttributedLeg</code> dedup.</p>
<p><strong>The best change is one I didn't ask for:</strong> pairs now auto-expand into forward + reverse legs (<a href="https://claude.ai/epitaxy/modules/pallets/intents-coprocessor/src/types.rs">types.rs</a>). That closes #1084's own caveat — "Buy &amp; sell requires both directions of a pair to be registered" — from a single config entry, with the duplicate check correctly moved to the <em>unordered</em> token set.</p>
<h3>New concerns</h3>
<p><strong>Migration now wipes <code>PhantomOrderConfig</code></strong> (<a href="https://claude.ai/epitaxy/modules/pallets/intents-coprocessor/src/migrations.rs">migrations.rs</a>) because <code>PhantomTokenPair</code> gained <code>standard_amount_b</code> and old bytes no longer decode. Correct, but it means <strong>governance must re-set the phantom config after the runtime upgrade or phantom order generation stops entirely</strong>. That belongs in the release checklist, not just a code comment.</p>
<p><strong><code>standard_amount_b</code> doubles the silent-failure surface.</strong> The pallet only checks non-zero; if it isn't exactly <code>10^decimals(token_b)</code>, <code>resolvePoolLeg</code> rejects the reverse leg and the pool silently gets one side — precisely what auto-expansion was added to prevent.</p>
<p><strong><code>readAllPages</code> + offset paging is a footgun.</strong> <code>@subql/types-core</code> warns that <code>getByFields</code> "will first search cache data followed by DB data — consider this when using order and offset options." As written it's safe (the PoolBidder read for pool P precedes P's own writes and filters on <code>poolId</code>), but the helper's doc should say so for the next caller.</p>
<hr>
<h2>Can a user query exact available liquidity for a cross-chain pair?</h2>
<p><strong>No — not exactly, and not in one query.</strong> The raw material is all there; the derivation isn't.</p>
<p>Here's what answering <em>"how much USDC@BSC → cNGN@Base can I route?"</em> actually takes today:</p>
<ol>
<li>Compute the slug client-side: sort symbols → <code>cNGN-USDC</code></li>
<li>Work out the direction: <code>token0 = cNGN</code>, so USDC→cNGN is token1→token0 = <strong>BUY</strong></li>
<li><code>poolBidders(filter: { poolId: "cNGN-USDC", chain: "EVM-8453", direction: "BUY", liquidity: { greaterThan: "0" } })</code></li>
<li><strong>Client-side</strong> filter: keep rows where <code>acceptedSources</code> is null or contains <code>"EVM-56"</code></li>
<li><strong>Client-side</strong> sum <code>liquidity</code></li>
<li>Price separately from <code>LiquidityPool.buyRate</code></li>
</ol>
<p>Six reasons that isn't "exact":</p>
<p><strong>1. No route entity.</strong> #1084 listed "a derived route view — route src → dst is live iff the destination pool has a verified bid backed by inventory whose declaration includes src" under Schema impact. Still absent. Every consumer re-implements the slug rule, the SELL/BUY orientation rule, the null-declaration default, and the sum.</p>
<p><strong>2. The declaration filter can't be expressed in the query.</strong> <code>acceptedSources</code> is a jsonb array with no reliable generated contains-filter — but the deeper problem is that <code>null</code> means <em>"all CCTP/USDT0-covered chains"</em> and <strong>that set exists nowhere in the schema</strong>. The consumer must hardcode it. And since <code>simplex.acceptedSourceChains</code> is opt-in and defaults to omitted, every row will be <code>null</code> at launch — so the source-chain dimension carries no information at all until solvers configure it.</p>
<p><strong>3. <code>liquidity</code> is an upper bound on inventory, not fillable size.</strong> The schema says all three reasons itself: it's the solver's <strong>global</strong> balance of that token on that chain (the same USDC backs every pool they quote, so summing across pools double-counts); the bid quotes one standard unit while real appetite is capped by the strategy's undisclosed <code>maxOrderSize</code>; and it's read at <strong>chain head</strong>, not at the snapshot's block.</p>
<p><strong>4. The obvious field is the wrong one.</strong> <code>LiquidityPool.sellDepth</code>/<code>buyDepth</code> are summed across <em>all</em> destination chains and include solvers that reject the source chain. A consumer reaching for the pool-level number — the natural thing to do — gets a materially wrong answer for a specific route. The docstring warns about double-counting across pools but not about this.</p>
<p><strong>5. No staleness signal.</strong> <code>PoolChainLiquidity</code> deliberately keeps the last known <code>rate</code> while zeroing <code>depth</code> when a window has no quotes, so a stale rate sits next to zero depth. Consumers must compare <code>lastUpdatedBlock</code> themselves.</p>
<p><strong>6. Coverage depends on governance config not in this repo.</strong> Cross-chain <em>stablecoin</em> routes need the same-chain <code>USDC/USDC</code>, <code>USDT/USDT</code>, <code>USDC/USDT</code> pools that #1084 specifies — which only exist if each chain's phantom config includes them. Combined with the migration wiping the config, someone must re-set it post-upgrade with the full pair list or those pools never materialize.</p>
<p><strong>Recommendation:</strong> add the derived route entity #1084 asked for, or at minimum export one helper that encapsulates slug → direction → accepted-source filter → sum, so those four rules have a single home. And put the CCTP/USDT0-covered chain set somewhere queryable so <code>acceptedSources: null</code> is resolvable without every client hardcoding it.</p></body></html>Six new commits since my review. Let me look at what changed.

Pulled 6 new commits (`88414ae7e..cf49f5363`). Pallet suite: **41/41 pass**.

## Follow-up review

Five of my six findings are fixed, cleanly:

| Finding | Status |
|---|---|
| RPC amplification per leg | ✅ `memoizedSolverBalance` caches by `(chain, token, solver)` and is shared with `sweepSolverLiquidity`; quoted legs now fetch via `Promise.all` |
| Sequential leg quoting vs bid window | ✅ `Promise.all` over legs, redundant `canFill` dropped, and the cache-pollution nit fixed (`id: undefined` + `setFillerOutputs` removed from `quotePhantomFill`) |
| `POOL_ROWS_LIMIT` truncation | ✅ `readAllPages` in [[store.helpers.ts](https://claude.ai/epitaxy/sdk/packages/indexer/src/utils/store.helpers.ts)](sdk/packages/indexer/src/utils/store.helpers.ts) with stable `orderBy: "id"` |
| Incomplete token registry | ✅ DAI, ZARP, EURC, XSGD, TRYB added across chains |
| Schema nullability inconsistency | ✅ Entities renamed to `PhantomOrderV2` / `PhantomOrderLeg` / `PhantomOrderPriceSnapshotV2`; `chain` is now `String!` |
| Unmeasured `generate_phantom_order` weight | ❌ **Still "Placeholder, not measured"** in both runtime weight files — coefficient doubled to 400k/pair for the two legs, but still a guess on a mandatory hook |

Plus `ZeroPhantomStandardAmount`, the `localeCompare` → code-unit fix, and the `weightedRate` / `AttributedLeg` dedup.

**The best change is one I didn't ask for:** pairs now auto-expand into forward + reverse legs ([[types.rs](https://claude.ai/epitaxy/modules/pallets/intents-coprocessor/src/types.rs)](modules/pallets/intents-coprocessor/src/types.rs)). That closes #1084's own caveat — "Buy & sell requires both directions of a pair to be registered" — from a single config entry, with the duplicate check correctly moved to the *unordered* token set.

### New concerns

**Migration now wipes `PhantomOrderConfig`** ([[migrations.rs](https://claude.ai/epitaxy/modules/pallets/intents-coprocessor/src/migrations.rs)](modules/pallets/intents-coprocessor/src/migrations.rs)) because `PhantomTokenPair` gained `standard_amount_b` and old bytes no longer decode. Correct, but it means **governance must re-set the phantom config after the runtime upgrade or phantom order generation stops entirely**. That belongs in the release checklist, not just a code comment.

**`standard_amount_b` doubles the silent-failure surface.** The pallet only checks non-zero; if it isn't exactly `10^decimals(token_b)`, `resolvePoolLeg` rejects the reverse leg and the pool silently gets one side — precisely what auto-expansion was added to prevent.

**`readAllPages` + offset paging is a footgun.** `@subql/types-core` warns that `getByFields` "will first search cache data followed by DB data — consider this when using order and offset options." As written it's safe (the PoolBidder read for pool P precedes P's own writes and filters on `poolId`), but the helper's doc should say so for the next caller.

---

## Can a user query exact available liquidity for a cross-chain pair?

**No — not exactly, and not in one query.** The raw material is all there; the derivation isn't.

Here's what answering *"how much USDC@BSC → cNGN@Base can I route?"* actually takes today:

1. Compute the slug client-side: sort symbols → `cNGN-USDC`
2. Work out the direction: `token0 = cNGN`, so USDC→cNGN is token1→token0 = **BUY**
3. `poolBidders(filter: { poolId: "cNGN-USDC", chain: "EVM-8453", direction: "BUY", liquidity: { greaterThan: "0" } })`
4. **Client-side** filter: keep rows where `acceptedSources` is null or contains `"EVM-56"`
5. **Client-side** sum `liquidity`
6. Price separately from `LiquidityPool.buyRate`

Six reasons that isn't "exact":

**1. No route entity.** #1084 listed "a derived route view — route src → dst is live iff the destination pool has a verified bid backed by inventory whose declaration includes src" under Schema impact. Still absent. Every consumer re-implements the slug rule, the SELL/BUY orientation rule, the null-declaration default, and the sum.

**2. The declaration filter can't be expressed in the query.** `acceptedSources` is a jsonb array with no reliable generated contains-filter — but the deeper problem is that `null` means *"all CCTP/USDT0-covered chains"* and **that set exists nowhere in the schema**. The consumer must hardcode it. And since `simplex.acceptedSourceChains` is opt-in and defaults to omitted, every row will be `null` at launch — so the source-chain dimension carries no information at all until solvers configure it.

**3. `liquidity` is an upper bound on inventory, not fillable size.** The schema says all three reasons itself: it's the solver's **global** balance of that token on that chain (the same USDC backs every pool they quote, so summing across pools double-counts); the bid quotes one standard unit while real appetite is capped by the strategy's undisclosed `maxOrderSize`; and it's read at **chain head**, not at the snapshot's block.

**4. The obvious field is the wrong one.** `LiquidityPool.sellDepth`/`buyDepth` are summed across *all* destination chains and include solvers that reject the source chain. A consumer reaching for the pool-level number — the natural thing to do — gets a materially wrong answer for a specific route. The docstring warns about double-counting across pools but not about this.

**5. No staleness signal.** `PoolChainLiquidity` deliberately keeps the last known `rate` while zeroing `depth` when a window has no quotes, so a stale rate sits next to zero depth. Consumers must compare `lastUpdatedBlock` themselves.

**Recommendation:** add the derived route entity #1084 asked for, or at minimum export one helper that encapsulates slug → direction → accepted-source filter → sum, so those four rules have a single home.

Wizdave97:
Let's track cross chain liquidity pools in a searchable manner using the information we get from acceptedSources

modules/pallets/intents-coprocessor/src/lib.rs
royvardhan: Done. CurrentPhantomOrder is a BoundedVec again, now holding one bundled order per configured chain instead of one per pair. PhantomOrderConfiguration takes a list of per-chain entries (chain plus its token_pairs) with a shared interval. on_initialize generates every chain's order in one pass and skips chains without a confirmed height, and place_bid and on_finalize check every active commitment.

modules/pallets/intents-coprocessor/src/lib.rs
royvardhan: Reverted. place_bid iterates the active batch again and enforces the phantom bid rules for whichever chain's commitment the bid targets.

sdk/packages/simplex/src/tests/phantom-filler.e2e.simnode.test.ts
royvardhan: Done. The e2e now registers two chains with two token pairs each (USDC/cNGN and USDT/cNGN). The second chain is another anvil fork of Base reporting Optimism's chain id, which works because the protocol contracts are CREATE2 deployed at the same addresses everywhere; the test's AssetRegistry overrides the token addresses per chain. For each chain it asserts the four positional legs, every filler's bid, per leg prices and liquidity weights from funded balances, and the accepted source declarations in all three states (explicit list, explicit empty list, none). The SubQuery pool pipeline needs the store so it can't run inside the simplex e2e; the cross-chain half of liquidity and route indexing (depth weighted pool merge, per chain PoolBidder reconciliation) is enforced in the indexer handler suite against the same aggregation output.

sdk/packages/simplex/src/services/ContractInteractionService.ts
royvardhan: It is injected. handlePhantomOrder in filler.ts passes this.config.acceptedSourceChains at the call site, threaded from the simplex.acceptedSourceChains TOML key through boot.ts into FillerConfig. The e2e also asserts the declaration round trips from filler config to aggregation output.

Wizdave97:
Three gaps remain on the query surface:

PoolRoute.depth doesn't carry the upper-bound caveat. It is now the field consumers read for "available liquidity", and its doc says only "Sum of the accepting bidders' output-token inventory on chain, 18-decimal normalized." The three reasons it's an upper bound — un-earmarked global inventory, undisclosed maxOrderSize, chain-head reads — are documented on PoolBidder and LiquidityPool but not here. Highest-value doc fix in the PR.
PoolRoute.depth double-counts across sourceChain. A solver declaring ["EVM-1","EVM-56","EVM-137"] contributes full liquidity to three rows. Correct per route; wrong to sum. LiquidityPool.sellDepth has an explicit note about cross-pool double-counting — PoolRoute needs the same for cross-source.
PoolRoute.direction is not @index, while PoolBidder.direction and PoolChainLiquidity.direction both are. A route query almost always pins a direction.
New concern: indexer load on generation blocks
This is the one I'd want addressed before merge. All chains share interval_blocks and PhantomBidWindow, and every chain's order is created on the same block — so all 16 PhantomBidWindowExhausted events fire on the same Hyperbridge block. That block then runs 16 sequential aggregatePhantomBids, and:

memoizedSolverBalance is constructed inside aggregatePhantomBids, so it doesn't survive across the 16 calls. sweepSolverLiquidity walks every chain × every yield-vault token per solver — byte-identical work, repeated 16×.
LiquidityProviderBalance ids are {chain}-{token}-{commitment}-{block}-{solver}, so the same balance is persisted 16× per interval under 16 commitments. 16× row growth for data identical by construction.
Hoisting the memo to module scope keyed by (blockNumber, chain, token, solver), or sweeping LP liquidity once per block instead of once per commitment, fixes both.

Wizdave97:
## Liquidity pool schema — what the frontend can query

Reference for consumers of the pool entities this PR adds. Every query below is against the indexer's GraphQL API and uses only fields that exist on this branch.

### Entity map

```
LiquidityPool  (id: "cNGN-USDC")            one row per token PAIR, merged across chains
├── chainLiquidity → PoolChainLiquidity     per (chain, direction) — the samples the pool merges
├── bidders        → PoolBidder             per (chain, direction, outputToken, solver)
├── routes         → PoolRoute              per (chain, direction, sourceChain)
└── snapshots      → PhantomOrderPriceSnapshotV2   append-only price history

LiquidityProvider (id: solver address)
├── poolPositions  → PoolBidder             reverse link into pools
└── balances       → LiquidityProviderBalanceV2    full inventory time series
```

All ids are deterministic, so every lookup below is a primary-key filter:

| Entity | id |
|---|---|
| `LiquidityPool` | `{token0}-{token1}` |
| `PoolChainLiquidity` | `{pool}-{chain}-{direction}` |
| `PoolBidder` | `{pool}-{chain}-{direction}-{outputToken}-{solver}` |
| `PoolRoute` | `{pool}-{chain}-{direction}-{sourceChain}` |

### Resolving a pair to a pool id and direction

Every query needs this first. Sort the two symbols case-insensitively; the first is `token0`.

```ts
const [token0, token1] = [a, b].sort((x, y) => (x.toLowerCase() <= y.toLowerCase() ? -1 : 1))
const poolId = `${token0}-${token1}`              // "cNGN-USDC"
const direction = input === token0 ? "SELL" : "BUY"
```

- `SELL` = token0 → token1
- `BUY`  = token1 → token0

So for the `cNGN-USDC` pool: `cNGN → USDC` is **SELL**, `USDC → cNGN` is **BUY**.

Same-asset pools (`USDC-USDC`) always resolve to `SELL`; `buyRate`/`buyDepth` stay empty by construction.

### Three unit systems — read this before doing arithmetic

| Fields | Units |
|---|---|
| `LiquidityPool.sellRate`/`buyRate`, `PoolChainLiquidity.rate` | Output units per **1 whole** input token, ×1e18. Decimals-free and chain-independent. |
| `LiquidityPool.sellDepth`/`buyDepth`, `PoolChainLiquidity.depth`/`unrestrictedDepth`, `PoolBidder.liquidity`, `PoolRoute.depth` | Token amounts **normalized to 18 decimals**. |
| `PhantomOrderPriceSnapshotV2.medianPrice`/`lowestPrice`/`highestPrice`/`standardAmount`, `LiquidityProviderBalanceV2.balance` | **Raw** smallest units of the actual token. |

The same solver's USDC-on-Base inventory reads as `1000000000000000000` in `PoolBidder.liquidity` and `1000000` in `LiquidityProviderBalanceV2.balance`. Both correct; not comparable without conversion.

Also note `lowestPrice` and `highestPrice` deliberately carry the **same value** as `medianPrice` — they are not the min/max of the bid set, so an outlier bid can't be read as a tradeable bound. Don't build a candlestick off them.

---

## 1. Total available liquidity across all chains, without double counting

**Do not sum `LiquidityPool.sellDepth` across pools.** Depth is the solvers' *global* inventory, not earmarked per pair — the same USDC balance backs `cNGN-USDC`, `USDC-USDT`, `USDC-USDC` and every other pool quoting USDC. Summing pool depths counts it once per pool. For the same reason `sellDepth + buyDepth` on a single pool is also wrong: they are two different tokens.

The only un-double-counted total is the provider inventory itself, one row per `(provider, chain, token)`. Two steps:

```graphql
query LatestSweepBlock {
  liquidityProviderBalanceV2s(orderBy: BLOCK_NUMBER_DESC, first: 1) {
    nodes { blockNumber }
  }
}
```

```graphql
query TotalLiquidity($block: BigFloat!) {
  liquidityProviderBalanceV2s(
    filter: { blockNumber: { equalTo: $block } }
    first: 1000
  ) {
    nodes {
      providerId
      chain
      tokenAddress
      balance      # RAW units — normalize by the token's decimals before summing
    }
  }
}
```

Every phantom order whose bid window closes on a block shares that block's rows, so pinning `blockNumber` to the latest gives exactly one row per `(provider, chain, token)` — no duplicates. Group by `(chain, tokenAddress)` for a per-chain or per-token breakdown, or normalize to USD for a single headline number.

Caveat: a solver that did not bid in the latest window has no rows at that block. That is usually what you want — it isn't backing any pool right now.

## 2. Total available liquidity for a token pair, across all chains

```graphql
query PairLiquidity($poolId: String!) {
  liquidityPools(filter: { id: { equalTo: $poolId } }) {
    nodes {
      token0Symbol
      token1Symbol
      sellDepth        # token1 receivable, summed across chains (18-dec normalized)
      buyDepth         # token0 receivable, summed across chains
      sellBidCount
      buyBidCount
      lastUpdatedBlock
      lastUpdatedAt
    }
  }
}
```

These are **two separate numbers, one per side** — `sellDepth` is token1 inventory, `buyDepth` is token0 inventory. Never add them.

Pick the side that matches the swap the user is doing: `token0 → token1` is limited by `sellDepth`; `token1 → token0` by `buyDepth`.

## 3. Total available liquidity for a token pair on one chain

```graphql
query PairLiquidityOnChain($poolId: String!, $chain: String!) {
  poolChainLiquidities(
    filter: { and: [{ poolId: { equalTo: $poolId } }, { chain: { equalTo: $chain } }] }
  ) {
    nodes {
      direction          # SELL or BUY
      rate               # this chain's sample, 1e18-scaled
      depth              # all bidders on this chain, 18-dec normalized
      unrestrictedDepth  # the slice from bidders with no source declaration
      bidCount
      lastUpdatedBlock
      lastUpdatedAt
    }
  }
}
```

Returns one row per direction. To fetch a single direction directly, filter on the deterministic id instead: `id: { equalTo: "cNGN-USDC-EVM-8453-SELL" }`.

`depth: 0` with a fresh `lastUpdatedBlock` means the latest bid window had no valid quotes for that direction on that chain — `rate` still carries the last known price, so check `lastUpdatedBlock` before treating a rate as live.

## 4. Total available liquidity for a token pair, for a cross-chain fill

Filling `USDC@BSC → cNGN@Base` only requires the solver to hold **cNGN on Base** — it receives the USDC on BSC and needs nothing there. So route liquidity is destination-side depth, restricted to solvers that accept the source chain.

That splits into two buckets, because solvers declare accepted sources explicitly and the indexer will not guess for those who don't:

```graphql
query CrossChainRoute($poolId: String!, $dest: String!, $direction: String!, $source: String!) {
  # Bidders that EXPLICITLY declared they accept payment from $source
  explicit: poolRoutes(
    filter: {
      and: [
        { poolId: { equalTo: $poolId } }
        { chain: { equalTo: $dest } }
        { direction: { equalTo: $direction } }
        { sourceChain: { equalTo: $source } }
      ]
    }
  ) {
    nodes { depth bidCount lastUpdatedBlock }
  }

  # Bidders with NO declaration — they have not restricted which sources they accept
  undeclared: poolChainLiquidities(
    filter: {
      and: [
        { poolId: { equalTo: $poolId } }
        { chain: { equalTo: $dest } }
        { direction: { equalTo: $direction } }
      ]
    }
  ) {
    nodes { unrestrictedDepth unrestrictedBidCount lastUpdatedBlock }
  }
}
```

For `USDC@BSC → cNGN@Base`: `poolId: "cNGN-USDC"`, `dest: "EVM-8453"`, `direction: "BUY"`, `source: "EVM-56"`.

```
routeDepth = explicit.depth
           + (countUndeclaredFrom(source) ? undeclared.unrestrictedDepth : 0)
```

An absent declaration means only that the solver has **not restricted** its accepted sources — it is not a statement that any particular source is accepted. The indexer therefore does not expand those bidders into route rows, and whether to count their capacity for a given source is the **frontend's policy call**.

Use `explicit.depth` alone for a conservative number you can stand behind: every solver in it has signed a declaration naming that source. Add `unrestrictedDepth` for an optimistic upper bound that assumes undeclared solvers would accept it.

Until solvers start setting `acceptedSourceChains` in their filler config (opt-in, unset by default), expect most depth to land in the `unrestrictedDepth` bucket and `poolRoutes` to be sparse — so an explicit-only number will read as near-zero at launch.

To list **everything** fillable on a destination from one source in a single query, drop the pool filter — this hits the `(sourceChain, chain)` composite index:

```graphql
query RoutesFromSource($source: String!, $dest: String!) {
  poolRoutes(
    filter: { and: [{ sourceChain: { equalTo: $source } }, { chain: { equalTo: $dest } }] }
  ) {
    nodes { poolId direction depth bidCount lastUpdatedBlock }
  }
}
```

**Do not sum `PoolRoute.depth` across `sourceChain`.** A solver declaring three sources contributes its full liquidity to all three rows — correct per route, wrong in aggregate.

## 5. Latest sell and buy price for a token pair

```graphql
query PairPrices($poolId: String!) {
  liquidityPools(filter: { id: { equalTo: $poolId } }) {
    nodes {
      token0Symbol
      token1Symbol
      sellRate           # token1 per 1 whole token0, ×1e18
      buyRate            # token0 per 1 whole token1, ×1e18
      sellBidCount
      buyBidCount
      lastUpdatedBlock
      lastUpdatedAt
    }
  }
}
```

```ts
const sell = Number(sellRate) / 1e18   // 1 cNGN → 0.00062 USDC
const buy  = Number(buyRate)  / 1e18   // 1 USDC → 1580 cNGN
```

`buyRate` is **not** `1 / sellRate` — the two directions are measured independently from their own legs, so the gap between them is the real spread. Either is `null` until that direction is first priced.

For the per-chain price rather than the cross-chain merge, use `PoolChainLiquidity.rate` from query 3. For history, `phantomOrderPriceSnapshotV2s`:

```graphql
query PriceHistory($poolId: String!, $direction: String!, $chain: String!) {
  phantomOrderPriceSnapshotV2s(
    filter: {
      and: [
        { poolId: { equalTo: $poolId } }
        { direction: { equalTo: $direction } }
        { chain: { equalTo: $chain } }
      ]
    }
    orderBy: BLOCK_NUMBER_DESC
    first: 100
  ) {
    nodes { blockNumber snapshotTime medianPrice standardAmount tokenA tokenB bidCount }
  }
}
```

Snapshot prices are **raw** units: the rate is `medianPrice / standardAmount`, adjusted for the two tokens' decimals.

---

## Caveats that apply to every depth number

1. **Depth is an upper bound, not fillable size.** The inventory is the solver's global balance, not earmarked for the pair; solvers cap individual fills with an undisclosed `maxOrderSize`; and balances are read at the chain head when the snapshot is taken, not at the snapshot's block.
2. **`PoolBidder` and `PoolRoute` are latest-window only.** Rows are deleted when a solver stops bidding. Historical bidding lives in `FillerBid`, balances in `LiquidityProviderBalanceV2`, prices in `PhantomOrderPriceSnapshotV2` — there is no history of route availability.
3. **Coverage is bounded twice.** A pool exists only if the pair is in the pallet's phantom config for that chain *and* both tokens are in the indexer's `POOL_TOKENS` registry (USDC, USDT, DAI, cNGN, ZARP, EURC, XSGD, TRYB). A leg failing either check still produces a `PhantomOrderPriceSnapshotV2` row with `pool: null` — the snapshot survives, the pool attribution doesn't.
4. **Always check `lastUpdatedBlock`.** Everything is as of the last bid window close, and a quiet chain keeps its last known rate.

Wizdave97 (approved):

PR #1093, opened 2026-08-01
[evm, sdk]: emit full order call payloads and graffiti in OrderPlaced

### LINKED ISSUE #1092: [evm]: emit full order call payloads and graffiti in OrderPlaced
## Problem

The `OrderPlaced` event (`evm/src/apps/intentsv2/IntentsBase.sol`) emits the order's token arrays as `TokenInfo[]` (token + amount only), which strips the two unbounded `bytes` payloads from the order:

- `predispatch.call`
- `output.call`

Both fields are part of the order commitment, so a filler cannot compute the order ID from the event alone. The simplex filler (`sdk/packages/simplex/src/core/event-monitor.ts`) works around this by recovering the original `placeOrder` calldata with `debug_traceTransaction` (callTracer) for **every order it processes**.

This means solver RPC endpoints must support the `debug` namespace. Free/public RPC endpoints don't offer it, which raises the cost floor for running a solver.

The `graffiti` attribution tag has the same issue for a different consumer: `placeOrder(Order order, bytes32 graffiti)` only receives it as calldata, so attribution indexers must also dig into transaction calldata instead of reading logs.

## Proposal

Emit the complete order in `OrderPlaced`:

1. Add `predispatch.call` and `output.call` bytes to the event (or emit the full `Order` struct).
2. Add the `bytes32 graffiti` parameter to the event.

## Benefits

- Solvers can run entirely on free RPC endpoints — order reconstruction becomes pure `eth_getLogs` + event decoding, no `debug_traceTransaction`.
- Removes the Tron limitation (no debug API there, so today only direct `placeOrder` calls are supported — nested/multicall placements can't be indexed).
- Removes the `occurrenceIndex` bookkeeping for transactions containing multiple `placeOrder` calls — each event becomes self-contained.
- Attribution indexers read graffiti straight from the log.

## Cost

LOG data is 8 gas per byte. Both `call` fields are empty (`0x`) for typical orders, so the common case adds roughly two length words plus one static 32-byte field for graffiti. Orders with large call payloads pay proportionally, but they are the rare case.

## Migration notes

- Event schema change → ABI break for all `OrderPlaced` consumers (simplex filler, indexers).
- The trace-based path should be kept for indexing historical blocks emitted under the old schema.

seunlanlege:
Closes #1092

## Summary

`OrderPlaced` now emits the complete order — `predispatch.call`, `output.call`, and the `bytes32 graffiti` attribution tag are appended to the event — so the order commitment is computable from the log alone.

```solidity
event OrderPlaced(
    ...,
    TokenInfo[] outputs,
    bytes predispatchCall,
    bytes outputCall,
    bytes32 graffiti
);
```

## Contracts

- `evm/src/apps/intentsv2/IntentsBase.sol`, emit site in `IntentGatewayV2.sol`, the `IIntentGatewayV2` interface in `@hyperbridge/core`, and the Tron variant all gain the three fields.
- New foundry test `testOrderPlacedEventCarriesCallPayloadsAndGraffiti` places an order with non-empty predispatch/output calldata and graffiti, decodes the raw log, and asserts an order reconstructed **from the event alone** hashes to the commitment holding the escrow. (Note: event data is the bare parameter tuple, so struct-decoding it requires prepending a `0x20` offset word — same tuple-wrapping distinction documented in CLAUDE.md for `abi_encode_params`.)
- Both `IntentGatewayV2` suites pass: 118/118.

## Simplex

Legacy orders are **not** supported — the filler only watches live blocks:

- `reconstructOrdersFromLogs` is now a pure, synchronous function; every order is rebuilt from its log. Logs missing the call payloads (pre-upgrade gateways) are rejected via `onError`.
- Removed `debug_traceTransaction` calldata recovery, the `occurrenceIndex` per-tx bookkeeping, and the `EvmChain` map that existed only to reach it. Order detection needs nothing beyond `eth_getLogs` — no `debug` namespace.
- Tests rewritten: full-order reconstruction with commitment check against `orderCommitment`, multi-order-per-tx self-containment, and missing-payload rejection.

## SDK

- `OrderPlaced` ABI entries updated in `@hyperbridge/sdk` and simplex; `DecodedOrderPlacedLog` gains the three fields (optional, since the V1 gateway ABI also decodes into this type).
- `deriveCanonicalPlacedOrder` reads the call payloads from the event when present.

## Indexer

- The handler reads beneficiary/calls/graffiti straight from new-schema logs — no trace needed.
- Historical blocks under the old schema still index: the old topic0 lives in its own datasource with a minimal `IntentGatewayV3Legacy.abi.json` asset (subql codegen validates topic filters against the datasource ABI, so the legacy signature can't share the main one). The trace-based recovery path is kept for those logs only.
- `subql codegen` passes and the generated per-chain yamls render both datasources.

## Migration notes

- Event signature change ⇒ new topic0; all `OrderPlaced` consumers must update their ABI.
- SDK integrators, simplex fillers, and the gateway contracts must upgrade in lockstep (new topic0): old SDKs/fillers can't see the new logs — an old-SDK `placeOrder` throws at receipt parsing after funds are escrowed — and the new SDK/filler can't see old-schema logs. The order commitment itself is unchanged. Documented in the 2.8.0 / 0.9.0 changelogs; versions bumped in this PR.

## Cost

LOG data is 8 gas/byte. Both call fields are empty for typical orders, so the common case adds two length words plus the static graffiti word; orders with large payloads pay proportionally.


sdk/packages/indexer/scripts/templates/evm-chain.yaml.hbs
```diff
@@ -82,6 +82,26 @@ dataSources:
...
+      intentGatewayV3Legacy:
+        file: ./abis/IntentGatewayV3Legacy.abi.json
+    mapping:
+      file: ./dist/index.js
+      handlers:
+        - kind: ethereum/LogHandler
+          handler: handleOrderPlacedEventV3
+          filter:
+            topics:
+              - 'OrderPlaced(bytes32,string,string,uint256,uint256,uint256,address,bytes32,(bytes32,uint256)[],(bytes32,uint256)[],(bytes32,uint256)[])'
```
royvardhan, line 104: Since we are adding new fields, do we want to extend the schema as well?

seunlanlege:
Can you please help take this over the finish line @royvardhan

royvardhan:
> Can you please help take this over the finish line @royvardhan 

Yes sure. 🫡

royvardhan:
> Can you please help take this over the finish line @royvardhan

On another look, pr seems complete.

royvardhan (approved):

PR #1099, opened 2026-08-04
 Add the BRIDGE HFT token contract 

dharjeezy:
`BridgeToken` is the EVM representation of BRIDGE, a mint and burn `HyperFungibleToken` for Hyperbridge with its  metadata and its nexus peer fixed in the bytecode, so it is functional the moment it is deployed and lands on the same CREATE2 address on every chain. `HftDecimals` on nexus moves from 18 to 12, the true decimal count of the native token, so the pallet scales by 10^6 against an 18 decimal ERC20 instead of leaving the EVM side off by that factor.

### DIFF
--- a/evm/script/DeployBridgeToken.s.sol
+++ b/evm/script/DeployBridgeToken.s.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.17;
+
+import "forge-std/Script.sol";
+import {HyperFungibleToken} from "@hyperbridge/core/apps/HyperFungibleToken.sol";
+import {BridgeToken} from "../src/apps/BridgeToken.sol";
+import {CallDispatcher} from "../src/utils/CallDispatcher.sol";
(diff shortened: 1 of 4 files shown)


Wizdave97 (approved):

evm/script/DeployBridgeToken.s.sol
```diff
@@ -0,0 +1,23 @@
...
+
+import "forge-std/Script.sol";
+import {HyperFungibleToken} from "@hyperbridge/core/apps/HyperFungibleToken.sol";
+import {BridgeToken} from "../src/apps/BridgeToken.sol";
+import {CallDispatcher} from "../src/utils/CallDispatcher.sol";
+import {BaseScript} from "./BaseScript.sol";
+
+contract DeployBridgeToken is BaseScript {
+    function deploy() internal override {
+        CallDispatcher dispatcher = new CallDispatcher{salt: salt}();
```
seunlanlege, line 12: calldispatcher is already deployed

PR #1108, opened 2026-08-07
[sdk]: bump order gas

ddboy19912:


seunlanlege (changes requested):
Better to make it 10% by default

seunlanlege (approved):

PR #1112, opened 2026-08-08
[indexer]: track placement and fill userop hashes, delivered amounts, and fee token

### LINKED ISSUE #1104: [indexer]: track fill userop hashes, actual delivered amounts, and placeOrder fee token
The IntentGatewayV3 indexer (`sdk/packages/indexer`) is missing three pieces of data that consumers (SDK order status streams, dashboards, solvers) need.

## 1. Userop hash for fills / partial fills

Solvers fill orders through the ERC-4337 `SolverAccount` (`evm/src/apps/intentsv2/SolverAccount.sol`), so the `transactionHash` currently stored on `IOrderV3Fill` / `IOrderV3PartialFill` is the bundler's EntryPoint transaction — not the userop the solver submitted and tracks on their side.

- Add a nullable `userOpHash` field to `IOrderV3Fill` and `IOrderV3PartialFill` in `sdk/packages/indexer/src/configs/schema.graphql`.
- In `orderFilledV3.event.handler.ts` / `partialFilledV3.event.handler.ts`, resolve it from the fill transaction's receipt: find the EntryPoint `UserOperationEvent` whose `sender` matches the `filler` address and record its `userOpHash`. Leave null for plain EOA fills.

## 2. Actual amount received by the user in a fill

The fill handlers record output amounts straight from the `OrderFilled` / `PartialFill` event args, but `ExtrinsicIntents.fill()` emits `outputFills[i].amount = totalRequired` — the order's *promised* amount (`evm/src/apps/intentsv2/ExtrinsicIntents.sol:134`). When a solver overfills, the beneficiary actually receives `totalRequired + beneficiaryShare` from the surplus split, and the solver's actual delivered amount (`solverAmount`) never appears in the event at all.

- Add an `amountReceived` field (alongside the existing promised `amount`) to `IOrderV3FillOutputAsset` and `IOrderV3PartialFillOutputAsset`.
- The solver transfers output assets directly to the user, so the actual delivered amount can be read from the `Transfer` logs (filler → beneficiary) in the fill transaction receipt.

## 3. Fee amount and fee token for placeOrder

`IOrderV3.fees` already records the `fees` value from the `OrderPlaced` event, but not the token it is denominated in. Fees are paid in the host's fee token (`IDispatcher(host()).feeToken()`), which differs per chain — so consumers can't display or price the fee without hardcoding per-chain knowledge.

- Add a `feeToken: String!` field to `IOrderV3` in the schema.
- Populate it in `orderPlacedV3.event.handler.ts` by reading `feeToken()` from the host contract (the value is static per chain, so cache it per chain rather than calling on every order).

All three are schema additions plus handler changes in `sdk/packages/indexer/src/handlers/events/intentGatewayV3/`; no reindex-incompatible changes expected beyond the new nullable/derivable fields.

seunlanlege:
Closes #1104.

Adds optional ERC-4337 operation hashes for Intent Gateway order placements and fills, actual delivered output amounts, and the fee token used at placement. Existing transaction hashes remain available separately, including the bundler transaction hash for account-abstraction operations.

## Schema

| Entity | Field | Behavior |
|---|---|---|
| `IOrderV3` | `userOpHash: String` | Placement operation hash, nullable and indexed |
| `IOrderV3Fill`, `IOrderV3PartialFill` | `userOpHash: String` | Fill operation hash, nullable and indexed |
| `IOrderV3FillOutputAsset`, `IOrderV3PartialFillOutputAsset` | `amountReceived: BigInt` | Actual delivery, nullable when unavailable or ambiguous |
| `IOrderV3` | `feeToken: String`, `feeTokenDecimals: Int` | Source host fee denomination at the placement block |

All seven additions are nullable. Existing rows retain their values and default to null for the new fields; this PR does not perform a historical backfill.

## Operation attribution

Fill matching uses the first canonical EntryPoint `UserOperationEvent` after the fill log and requires its sender to match the filler. A different sender ends the search rather than falling through to another operation later in the bundle.

Placement matching supports both current and legacy `OrderPlaced` events. It matches the placing account from `OrderPlaced.user`, independently of the output beneficiary, and also requires the EntryPoint's `BeforeExecution` boundary to precede the placement. This prevents placements during validation or before a later bundle from acquiring the wrong operation hash. Multiple placements inside one operation share its hash; repeated operations by the same account are distinguished by log order.

Direct placements, unknown EntryPoints, sender mismatches, and missing/unavailable receipts leave the placement hash null. Placement receipt lookup is optional and does not block order indexing. Replays can populate existing rows; a failed lookup preserves a previously stored hash.

## Delivered amounts

The gateway's fill event reports the promised output amount, which can differ from the beneficiary's delivery after an overfill. `amountReceived` is matched from ERC-20 `Transfer` logs from filler to beneficiary, bounded by successive gateway fill events. Repeated tokens consume transfers in log order; missing or extra transfers leave attribution unset. Native outputs remain null, and zero-valued partial-fill slots remain zero without consuming a transfer.

When the source order is not indexed yet, beneficiary recovery checks direct or traced `fillOrder` calldata and verifies the recomputed order commitment. Both historical and current fill selectors are supported.

Receipt lookup failures leave a fill's `userOpHash` and `amountReceived` fields null while the fill, event amounts, and normal status and inventory processing continue.

## Fee denomination

The host fee token and decimals are read through SubQuery's provider, which scopes calls to the indexed placement block. A per-chain cache keyed by block hash deduplicates successful reads within that block. Missing host configuration or failed contract reads leave the fee-token fields null without blocking placement. Placement replays preserve previously stored fee-token fields when a lookup fails.

## Validation

- 61 focused helper and handler tests pass, including current/legacy placements, batched operations, repeated senders, execution boundaries, spoofed events, direct placements, placement and fill receipt failures, fee-token lookup failures, replay preservation, delivered amounts, host metadata caching, and nested trace recovery.
- `ENV=mainnet npm run build:release` passes, including schema and ABI code generation.
- The real SubQuery/Postgres migration test passes from the main schema used in the PR refresh to the expanded schema, preserves old rows in all five affected tables, accepts all seven nullable fields, and restarts with unchanged SDL.

The field descriptions are recorded in `sdk/packages/indexer/docs/ai/changelog/2026-09-13-intent-gateway-order-fields.md`.

Repository guidance now requires PR documentation across all packages to describe the final feature or fix and omit routine work logs and cleanup narratives. `CLAUDE.md` defines the rule; `AGENTS.md` and the package documentation guides refer to it.


Wizdave97:
@royvardhan for this PR we'll need to modify subquery to allow us add new fields to schema entities, please look into it.

Wizdave97:
@royvardhan i think we can resume work on this, the inplace schema change is working now

CLAUDE.md
```diff
@@ -37,28 +37,29 @@ The Solidity test helper contract is `evm/tests/foundry/AbiEncodeTest.sol`.
...
-the code runs.
-
-When changing code in a package that has `docs/ai/`, writing these files is part of the
-change, not optional follow-up. When starting substantial work in a package that has none
-yet (lz-endpoint), create the three directories and the README as the first step and seed
-them from that task's actual work — never with empty templates.
-
-These are not release notes. Package `CHANGELOG.md` files are changesets release logs,
-managed separately.
+This rule applies to every PR and package, including indexer, simplex, SDK, and core.
```
Wizdave97, line 40: **Blocking: unrelated repo-wide policy change.** This rewrites the AI workflow docs rule for every package, adds `AGENTS.md`, and replaces all four `docs/ai/README.md` files. It drops the changelog/decisions/flows layout and the lz-endpoint setup rule. That is a team policy decision unrelated to the indexer fields, so please move it to its own PR. Merged here, it also makes other open PRs (e.g. #1185) non-compliant as soon as it lands.

sdk/packages/indexer/src/utils/fill.helpers.ts
```diff
@@ -0,0 +1,235 @@
...
+	},
+): Promise<FillEnrichment> {
+	const { commitment, filler, outputs, chain } = params
+
+	if (!fillLog.transaction) {
+		logger.warn(`No transaction on fill log ${fillLog.transactionHash}, skipping fill enrichment`)
+		return {}
+	}
+
+	const receipt = await fillLog.transaction.receipt()
```
Wizdave97: **Blocking: a receipt failure now fails fill indexing.** `receipt()` isn't caught, and both fill handlers await `resolveFillEnrichment` before `recordFill`. `wrap` rethrows non-decode errors, so one RPC error on the receipt fails the `OrderFilled`/`PartialFill` handler. The placement path (`resolvePlacementUserOpHash`) already catches and returns undefined. Please do the same here, returning `{}` on failure so the fill is still recorded with null `userOpHash`/`amountReceived`. None of the current tests make the receipt throw on the fill path.

sdk/packages/indexer/src/handlers/events/intentGatewayV3/orderPlacedV3.event.handler.ts
```diff
@@ -109,10 +115,15 @@ export const handleOrderPlacedEventV3 = wrap(async (event: OrderPlacedLog): Prom
 
 	logger.info(`[Intent Gateway V3] Order Commitment: ${commitment}`)
 
+	// Fees are paid in the host's fee token, which differs per chain.
+	const feeToken = await getHostFeeToken(chain, blockHash)
```
Wizdave97: **Blocking: the fee-token read can block order placement.** `getHostFeeToken` throws on a missing `CHAINS_BY_ISMP_HOST` entry or an RPC error, so `OrderPlaced` fails even though `feeToken`/`feeTokenDecimals` are nullable. It should fall back to null like the userOp lookup does. The handler tests always mock this call to succeed.

Related: `{ blockTag: blockHash }` in `host.helpers.ts` does nothing. SubQuery's `SafeEthProvider.call` ignores `blockTag`, logs "Provided parameter 'blockTag' will not be used", and already reads at the block being indexed. So it can be dropped, which also removes two warnings per uncached placement.

Wizdave97 (commented):
Three blocking issues below. The attribution logic itself checks out against the contracts: fill transfers are `transferFrom(msg.sender, beneficiary, fill + beneficiaryShare)` and are emitted before the fill event, and `placeOrder` sets `order.user = msg.sender`.

sdk/packages/indexer/src/utils/fill.helpers.ts
Wizdave97: @royvardhan

sdk/packages/indexer/src/handlers/events/intentGatewayV3/orderPlacedV3.event.handler.ts
Wizdave97: @royvardhan

Wizdave97 (approved):

PR #1118, opened 2026-08-10
feat(sdk): public-RPC gas estimation — token storage slots + fixed no-bundler budgets

royvardhan:
## Summary

Makes gas estimation work on RPCs without the debug API:

1. Adds ERC20 balance/allowance storage slots and decimals for the exotic stablecoins to the SDK chain config, so `GasEstimator` state overrides resolve from config instead of falling back to a live `debug_traceCall`.
2. Replaces the no-bundler `estimateContractGas` fallback (which needs `eth_estimateGas` with state overrides, also unsupported on public RPCs) with fixed gas budgets.

### Token storage slots (`sdk/packages/sdk/src/configs/chain.ts`)

- Extended the `tokenStorageSlots` type with `cNGN`, `ZARP`, `EURC`, `XSGD`, `TRYB`, `USDR` keys (previously only `USDT | USDC | WETH | DAI` were expressible)
- Extended the `tokenDecimals` type likewise and added on-chain-verified decimals
- Added slot entries for Ethereum (1), Base (8453) and Polygon (137):

| Token | balanceSlot / allowanceSlot | Decimals | Chains |
|---|---|---|---|
| cNGN | 201 / 202 (custom upgradeable layout) | 6 | ETH, Base, Polygon |
| ZARP | 51 / 52 | 18 | ETH, Base, Polygon |
| EURC | 9 / 10 (Circle FiatToken) | 6 | ETH, Base |
| XSGD | 7 / 8 | 6 | ETH, Polygon |
| TRYB | 9 / 10 | 6 | ETH |
| USDR | 51 / 52 | 6 | ETH, Base, Polygon |

### No-bundler gas estimation (`sdk/packages/sdk/src/protocols/intents/GasEstimator.ts`)

- Removed the `estimateContractGas` + `stateOverride` fallback in `estimateFillOrder`. Without a bundler, `callGasLimit` is now `NO_BUNDLER_FILL_GAS_BASE` (700k, sized for a vault unlock and redeem plus cross-chain dispatch) plus `NO_BUNDLER_FILL_GAS_PER_OUTPUT` (150k) per output leg.
- The no-bundler quote now also carries paymaster gas (250k verification, 100k postOp) instead of zero, mirroring the limits the paymaster packs into `paymasterAndData`, so paymaster-sponsored fills are no longer under-quoted.
- State override building is skipped entirely when no bundler is configured, so the no-bundler path never touches storage-slot resolution.

## Verification

- Slots discovered via `debug_traceCall` (prestateTracer) and reverse-engineered by matching keccak mapping locations across candidate base slots; every slot confirmed with a second independent address. All tokens use the Solidity mapping layout.
- Base cNGN (201/202) matches the values already recorded in the indexer's `config-mainnet.json`, and the yield-vault holder's balance matched `eth_getStorageAt` exactly.
- End-to-end through `getRecordedStorageSlot()`: computed locations matched live chain state (`eth_getStorageAt` vs `balanceOf`) for cNGN, XSGD and EURC with real holders.
- Gas budgets sanity-checked against a live Base fill (`0xb16fe2eabe02406d9cc41191b11a28b6328ca433d9d4342f37428f135c47e66c`): 432,375 gas used for the whole handleOps transaction against a 1,200,000 call limit, so a single-output budget of 850k leaves roughly 2.5x margin over actual call-phase usage.
- `tsc --noEmit` on the sdk package introduces no new errors.

Note: EURC/TRYB (FiatToken v2.2) pack a blacklist flag into the top bit of the balance slot; harmless here since the estimator's override value (`maxUint256/2`) keeps that bit clear, same as the existing USDC entries.

### DIFF
--- a/sdk/packages/sdk/CHANGELOG.md
+++ b/sdk/packages/sdk/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @hyperbridge/sdk
 
+## 2.8.1
+
+### Patch Changes
+
+- Recorded balance/allowance storage slots and decimals for cNGN, ZARP, EURC, XSGD, TRYB and USDR on Ethereum, Base and Polygon, so `GasEstimator` state overrides for these tokens resolve from config instead of requiring an RPC with `debug_traceCall`.
+
 ## 2.8.0
 
 ### Minor Changes
--- a/sdk/packages/sdk/package.json
+++ b/sdk/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@hyperbridge/sdk",
-	"version": "2.8.0",
+	"version": "2.8.1",
 	"description": "The hyperclient SDK provides utilities for querying proofs and statuses for cross-chain requests from HyperBridge.",
 	"type": "module",
 	"types": "./dist/node/index.d.ts",
--- a/sdk/packages/sdk/src/configs/chain.ts
+++ b/sdk/packages/sdk/src/configs/chain.ts
@@ -158,12 +158,23 @@ export interface ChainConfigData {
 		USDT: number
 		cNGN?: number
 		EXT?: number
+		ZARP?: number
+		EURC?: number
+		XSGD?: number
+		TRYB?: number
+		USDR?: number
 	}
 	tokenStorageSlots?: {
 		USDT?: { balanceSlot: number; allowanceSlot: number }
 		USDC?: { balanceSlot: number; allowanceSlot: number }
 		WETH?: { balanceSlot: number; allowanceSlot: number }
 		DAI?: { balanceSlot: number; allowanceSlot: number }
+		cNGN?: { balanceSlot: number; allowanceSlot: number }
+		ZARP?: { balanceSlot: number; allowanceSlot: number }
+		EURC?: { balanceSlot: number; allowanceSlot: number }
+		XSGD?: { balanceSlot: number; allowanceSlot: number }
+		TRYB?: { balanceSlot: number; allowanceSlot: number }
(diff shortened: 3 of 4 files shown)


royvardhan:
Goal of this PR: let operators run simplex on public RPCs. Both changes remove the chain-RPC features public endpoints lack (`debug_traceCall` slot discovery, `eth_estimateGas` with state overrides).

The one remaining non-public dependency is the ERC-4337 bundler — `bundlerUrl` is still mandatory in the simplex config, and a bare RPC doesn't provide one. We could drop the requirement and run on the fixed budgets added here (delegation, token sends and vault hydration already fall back to direct transactions without a bundler). But Pimlico has a free tier, so even public-RPC operators can keep realistic estimates and UserOp sending.

Drop the mandatory bundler check, or keep it and point operators at a free bundler tier?

royvardhan:
Remaining public-RPC gap: the event scanner (`sdk/packages/simplex/src/core/event-monitor.ts`).

- 1s tick per chain (`getBlockNumber` + `getLogs`), fanned out to every quorum endpoint; confirmation polling (300ms single, 1s quorum) adds to it. Free tiers (10-25 rps) throttle, throttled providers drop out of quorum, and detection stalls silently.
- `getLogs` window is fixed at 1000 blocks; many public providers cap lower. The error handler only recognizes the two "not indexed yet" messages, so a smaller cap fails every tick, the cursor never advances, and the scanner wedges permanently.

Proposed fix: adaptive window (halve on range-cap errors, grow back after clean scans) + exponential backoff on rate limits. But doesn't this make the filler slow?

seunlanlege (approved):

PR #1122, opened 2026-08-12
[sdk]: improve bid fallback and bundler retry handling

ddboy19912:


sdk/packages/sdk/src/protocols/intents/Bid.ts
```diff
@@ -207,10 +225,31 @@ export class BidImpl implements Bid {
...
-			CryptoUtils.prepareBundlerCall(signedUserOp),
-			entryPointAddress,
-		])
+		// The hash is deterministic and does not include the signature. Computing it
+		// before submission lets retries recover when the bundler has accepted the
+		// operation but answers a repeated request with "already known".
+		const expectedUserOpHash = CryptoUtils.computeUserOpHash(signedUserOp, entryPointAddress, this.chainId())
+		let userOpHash = expectedUserOpHash as HexString
+		try {
+			userOpHash = await this.crypto.sendBundler<HexString>(BundlerMethod.ETH_SEND_USER_OPERATION, [
```
Wizdave97: There's no need for this, just cycle through all the bids until the order is expired

Wizdave97: If a bid submission returns an error, move to the next bid

Wizdave97: The sdk was doing that before, it was likely removed during a refactor

Wizdave97 (approved):

PR #1123, opened 2026-08-13
[simplex, docs]: publish @hyperbridge/simplex as a library

seunlanlege:
Makes `@hyperbridge/simplex` publishable and usable as a library, alongside the existing binary.

## Why

The package already exported `IntentFiller`, `FXFiller`, `EventMonitor`, `ChainClientManager` and `ContractInteractionService` — but none of them could be constructed from outside. `FillerConfigService` is a constructor parameter of all five and was never exported, and `TradingPair` required a `FillerPricePolicy` that wasn't either. The one API worth having, `bootFiller`, was internal.

It also wasn't installable: `ckb-mmr-wasm` was a `link:` dependency pointing at a directory that doesn't exist.

## The API

```ts
import { Simplex } from "@hyperbridge/simplex"
import { SqliteDataStore } from "@hyperbridge/simplex/sqlite"

const simplex = await Simplex.start({
  config,                                    // plain object, no TOML
  data: new SqliteDataStore("./simplex-data"),
  logger: process.stdout,                    // omit and it logs nothing
})

simplex.on("order:filled", ({ orderId, profitUsd }) => …)

await simplex.pairs.setCurve(0, "ask", [{ amount: "0", price: "1550" }])
await simplex.chains.setRpcUrls(8453, ["https://base-new.example"])
await simplex.stop()
```

Controllers cover `pairs`, `chains`, `vaults`, `assets`, `wallet` and `rebalancing`; events are typed. Most of this already existed inside `UiServer`'s `OperatorContext` as a bag of optional closures reachable only over HTTP — it's now a class the UI server consumes like any other caller. `bootFiller`, `FillerConfigService`, `IntentFiller`, `FXFiller` and the funding planners become internal.

## Shared scanners

Scanning a chain is identical work for every solver — the `getLogs` filter carries no solver-specific term, and `OrderFilled.filler` is `indexed: false` so it could never be narrowed to a topic anyway. So N solvers on one chain issued N copies of the same request stream forever, and at any real fleet size the scanning, not the filling, is what exhausts an RPC quota.

`OrderScanner` and `HyperbridgeScanner` run the scan once and feed every solver they're handed to:

```ts
const orders = await OrderScanner.create({
  chains: config.chains,       // same shape as the config's chains array
  scanIntervalSecs: 0.5,       // optional; defaults to 3, same as the TOML knob
})
const solverA = await Simplex.start({ config: configA, orderScanner: orders })
const solverB = await Simplex.start({ config: configB, orderScanner: orders })
await orders.close()   // yours to close; stop() leaves it running
```

Sharing is explicit — omit them and a solver builds private ones from its own config and closes them on `stop()`. Nothing is shared implicitly.

Delivery is at-least-once and per-chain FIFO. A scanner never blocks on a slow subscriber, because one would stall every other solver reading it; a subscriber that falls behind loses its oldest events and counts them in `subscription.dropped`.

`setRpcUrls` swaps a chain's endpoints while keeping its block cursor. The previous implementation was remove-then-add, which drops the `ChainScanner` and restarts from the head — silently skipping every block since the last scan. With one cursor shared by every solver, that gap is missed by all of them.

## Per-instance logging

**The library writes nothing anywhere until told where.** Importing it doesn't touch stdout, install a formatter or spawn a transport worker.

Sinks and levels are per instance rather than process-wide, so two solvers in one process can log to different destinations at different levels without one seeing the other's records. Each owns a `LoggerContext`; every service resolves its logger from the one it was built with.

## Pluggable persistence

`SimplexDataStore` = `{ bids, activity, state }`, all async, so a consumer can back it with Postgres or a hosted API instead of only SQLite. `MemoryDataStore` is the default; the SQLite implementation moves to the `./sqlite` subpath so nothing native is loaded unless you import it.

`bids` is the correctness-critical half — a successful bid locks a deposit only reclaimable by retraction, and the sweep finds what to reclaim by querying it. Two ordering fixes fell out of making those calls async:

- the bid write on the fill path is awaited before the `pendingRetractions` check that reads it
- `handleOrderFilledOnChain` flags the deferral *before* reading, so a bid landing in the gap can't strand itself for the full TTL

## Live chain edits

Chains can be added, removed and repointed at new RPCs without a restart. This was blocked by `ViemClientFactory` being a process-wide singleton keyed on chain id with no invalidation — which also meant two fillers in one process would silently share clients. Per-instance now, which unblocks both.

`add` probes endpoints and reads the chain id back before mutating, and rolls back everything it touched on failure. `remove` drains in-flight fills, then stops scanning. On a scanner you supplied, `add` requires the chain to already be in it and `setRpcUrls` refuses outright — repointing endpoints under the other solvers reading it would be action at a distance.

## Oversized orders are filled to the cap, not skipped

`maxOrderSize` is a per-order exposure cap, so an order larger than it is now **partially filled** down to the cap rather than skipped. The gateway releases escrow strictly pro-rata to the outputs provided (`inputs[i].amount * fillAmount / totalRequired`), so scaling the output by the cap fraction scales the exposure taken on by the same factor.

A partial collects **no `order.fees`** — the gateway releases those only to whoever completes the order — while still paying gas. What pays for it is the margin already in the operator's curve, which the engine cannot measure. So partials are exempt from the profitability floor: configuring a curve is the decision to fill at it. Three cases still skip, because the fill would revert or be mispriced rather than merely be unprofitable: cross-chain orders (the gateway rejects any under-fill), orders carrying output calldata, and orders another solver has already partially filled.

A price shortfall still skips outright. Escrow releases at the *order's* rate, so a smaller slice of a bad rate loses the same per unit and costs the same gas.

Two adjacent bugs fell out of this. The P&L valued every leg against the whole escrow while a partial releases only pro-rata — the balance-shortfall path has been emitting same-chain partial fills all along and booking them as if the full escrow arrived. And an under-fill was only refused for cross-chain orders, so a same-chain calldata order the wallet couldn't cover was bid on and reverted with `PartialFillNotAllowed`.

## Config

- `[simplex.queue]` removed. It was prompted by the wizard, written into every generated config, parsed, and copied to `FillerConfig.pendingQueueConfig` — which nothing has ever read. The TOML type keeps the key optional so existing configs still load and are ignored exactly as before.
- `maxConcurrentOrders` is now optional, defaulting to 5, and validated (integer ≥ 1) — `0` silently meant 5 via `||`, and fractional values crashed p-queue with a bare TypeError.
- `blockScanIntervalSeconds` (from #1119) is wired through to where the scan loop now lives, documented on all three pages, and exposed on `OrderScanner.create` as `scanIntervalSecs` with the same 0.1s floor.
- `entryPointAddress` / `solverAccountContractAddress` are marked accepted-and-ignored (addresses come from the SDK chain registry); the wizard stops emitting them.
- `SimplexOptions.dataDir` is dropped. Nothing ever wrote to it — storage location belongs entirely to the `data` store — and its only reader was the CLI dashboard's status payload, which the CLI now feeds directly.

## Packaging

- drop `ckb-mmr-wasm`; pin `@polkadot/api`, `util`, `util-crypto` off `"latest"`
- split tsup — library entries externalize dependencies so a consumer resolves one copy of `viem` and `decimal.js` (`instanceof Decimal` is false across two copies, and `TradingPair.maxOrderSize` is a `Decimal` callers construct); the CLI stays a single bundled file
- inline `@uniswap/*` and `ethers` anyway: they publish ESM with extensionless relative imports that esbuild resolves but Node's loader rejects, so `import "@hyperbridge/simplex"` threw `ERR_MODULE_NOT_FOUND` in a plain Node app. Neither reaches the public API.
- add `types` conditions, `repository`/`homepage`/`bugs`/`keywords`, `sideEffects`
- the bundled CLI keeps its source map — an operator crash names `src/config/filler-toml.ts:269`, not `simplex.js:328059`. The flag rides in the shebang (`env -S node --enable-source-maps`; the runtime API cannot map the module it is called from) and the docker ENTRYPOINT. Paid for by dropping the CLI's CJS twin, which nothing executed: **19.6MB → 11.2MB packed**
- add the npm publish step. The `simplex-v*` workflow built docker images and cut a release but never published, so the package has never existed on the registry — every `npm install @hyperbridge/simplex` in these docs was a 404. Needs `submodules: recursive` and `protoc` (the build runs protoc over `proto/mpcvaultapis`), and builds `@hyperbridge/sdk` first since its types are inlined into simplex's `.d.ts`. Refuses a tag that disagrees with `package.json`, which would otherwise publish one version under a tag announcing another.

## Also

The Prometheus metrics server is removed outright — `-p/--port`, `MetricsService`, `prom-client`, and the Prometheus + Grafana compose services. The typed event stream is the telemetry surface; the exporter consumed it with no privileged access, which proved events suffice, so counters are the operator's to build outside the process. The container healthcheck now probes the UI server's `/health` on 8686.

Watch-only orders emitted a bespoke `orderDetected` event nothing listened to, so they never appeared in the activity feed or the skip metric. They now emit `orderSkipped` with `reason: "watch-only"`.

## Docs

The Intent Gateway [Simplex page](https://docs.hyperbridge.network/developers/evm/intent-gateway/simplex) stays where it is, and the TypeScript SDK section gets a new Simplex page covering embedded usage plus a full API reference — the class and its six controllers, the event map, the `SimplexDataStore` interfaces, and the exported validators and types.

Auditing every factual claim on those pages against the source turned up mismatches worth calling out separately, because most were the docs describing behaviour the code no longer has:

- The overfill clamp was disabled in #964, which leaves the `maxConsecutiveClamps` halt unreachable and `isHalted()` pinned to `false`. All three pages documented it as a live per-leg loss bound. The docs no longer advertise it; **the dead machinery is still in the code and wants a decision** — re-enable a loss bound, or delete `halted` / `consecutiveClamps` / `isHalted()` / `resetHalt()` / `status().halted` and the halt-controls plumbing.
- `--admin-port` doesn't exist; the live-edit surface is the `--ui` dashboard
- only same-token legs are gated on margin — cross-asset and one-sided legs fill at the operator's own curve and their margin is telemetry
- a crossed book is allowed with a wizard warning, not a startup failure
- `redeemAll` only unwinds positions flagged `redeemOnShutdown`; stopping does not sweep outstanding bids
- a `uniswapV4` price guard needs both `referencePrice` and `maxDeviationBps`, and `maxDeviationBps` is a bare number — the documented example would not boot

## Testing

423 passing, 64 new across chain lifecycle, scanner ownership and cursor semantics, per-instance client and logger isolation, the exposure-cap math, and a store-parity suite that runs the same behavioural tests against both the memory and SQLite bid stores — the memory one is the library default, so a divergence there would strand deposits.

3 failures are pre-existing and environmental: `tests/rebalancers/{cctp,usdt0,binance}.test.ts` read RPC URLs and an API key from `.env.local`; they fail identically with this branch stashed.

Rebased onto latest `main` (through #1136); history is linear. Two integrations were real work rather than conflict resolution. From the earlier merge window: main's `compactError` serializer lives inside this branch's per-instance `LoggerContext`, and `blockScanIntervalSeconds` is wired into `ChainScanner`, where the scan loop now lives. From the rebase: #1136's phantom-bid batching composes with the shared scanners — `HyperbridgeScanner` delivers per-block batches (`onPhantomOrders`), because the batch boundary is what lets a solver bid on every chain's order in one extrinsic instead of queueing per-chain extrinsics on the account nonce; and `BalanceProvider` reuses the filler's Hyperbridge connection (main's shape) merged with this branch's per-instance logging and shutdown guard.

Verified beyond the suite: built ESM and CJS entries load from outside the repo under plain Node with `moduleResolution: nodenext`, `dist/bin/simplex.js --help` runs, every TOML block in the docs parses with the same `toml` package the CLI uses, and the docs site builds and serves all three pages.

The branch was audited adversarially in four passes, each finding verified against source before being acted on. The third found two criticals introduced by earlier commits in this PR (a partial-fill exemption that inferred its decision from a cache written before the profit gates ran, and a browser wizard blank-screened by the `queue` removal missing the `ui/` bundle). The fourth audited the fixes themselves and confirmed 22 more — five of them regressions in the previous fix commit, since corrected: scanner teardown races and unbounded probe waits, a wizard that aborted its own round-trip gate on configs carrying current knobs, a transient `--watch-only` getting persisted, and a CI publish that would have pinned an `@hyperbridge/sdk` version whose npm content had drifted from the workspace (the workflow now refuses when `packages/sdk` changed since its `sdk-v*` tag).

## Not done

`config/filler-toml.ts` keeps its filename. `SimplexConfig` is the exported name so the TOML association is gone from the public API, and moving the file is churn across many imports for no consumer-visible gain.


Wizdave97:
1. Simplex.start() hangs forever on an unreachable Hyperbridge node

simplex.ts:759 awaits HyperbridgeStream.create(wsUrl), which awaits ApiPromise.create with a default WsProvider — that retries indefinitely and never rejects. Verified empirically against a dead port: still pending after 20s, no rejection.

IntentFiller got a 30s Promise.race for exactly this hazard (filler.ts:108), but the stream didn't. This is a new boot-path blocker: previously IntentsCoprocessor.connect was fired and not awaited during startup. And hyperbridge-stream.ts:49 documents @throws if the endpoint cannot be reached, which it does not — the catch at line 117 only fires on a synchronous construction error. Same race guard belongs here.

2. Simplex.start() leaks the streams it owns when boot fails

simplex.ts:754-769: OrderStream.create (starts a scan loop per chain) and HyperbridgeStream.create (opens a socket) run before bootFiller, with no try/catch. Any boot rejection — including validateConfig, which now runs after the streams are up — leaves scanners polling and a socket open with nothing holding a reference. That's the same defect the "mid-boot failure orphaned a filler" commit set out to fix, one layer up: bootFiller unwinds what bootFiller started, but nothing unwinds what Simplex.start started.

3. chains.add() leaves the config service mutated when it rejects on a caller-supplied stream

simplex.ts:397-409:

configService.addChain(resolved)                        // mutates
if (…) intentFiller.setWatchOnly(chainId, true)         // mutates
if (!this.ownedStream) throw new Error(…)               // no rollback
The rollback block at line 412 only covers intentFiller.addChain. So on a shared stream the chain is left registered in configService (and in getWatchOnly()) but absent from resolvedChains and config.chains — list() won't show it, a retry fails with "already configured", and the next syncWatchOnlyToConfig() writes a nonexistent chain into the persisted config. The API reference states the opposite: "a failed add rolls itself back — a chain never ends up registered but unscanned." Move the ownership check above the first mutation.

royvardhan:
Thanks, commissioning some mental compute to review this.

Wizdave97:
Findings:

- [[scanner/fan-out.ts:69](https://claude.ai/epitaxy/sdk/packages/simplex/src/scanner/fan-out.ts:69)](sdk/packages/simplex/src/scanner/fan-out.ts:69) — a synchronous publish burst larger than the 1000-event capacity drops the oldest orders before the consumer ever gets a microtask, and the cursor advances so they're never re-scanned.
- [[services/CacheService.ts:141](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/CacheService.ts:141)](sdk/packages/simplex/src/services/CacheService.ts:141) — `partialFills` is only pruned alongside a stale `fundingPrepends` entry, which most orders never have; unbounded growth.
- [[data/sqlite/bids.ts:104](https://claude.ai/epitaxy/sdk/packages/simplex/src/data/sqlite/bids.ts:104)](sdk/packages/simplex/src/data/sqlite/bids.ts:104) — `byCommitment` orders by second-resolution `created_at`; a same-second re-bid can return the retracted row and permanently suppress retraction. Diverges from the memory store the new parity suite compares against.
- [[strategies/fx.ts:542](https://claude.ai/epitaxy/sdk/packages/simplex/src/strategies/fx.ts:542)](sdk/packages/simplex/src/strategies/fx.ts:542) — `hasExistingPartialFill` probes one contract read per output on every same-chain order, even when the result is discarded; should be lazy.
- [[strategies/fx.ts:784](https://claude.ai/epitaxy/sdk/packages/simplex/src/strategies/fx.ts:784)](sdk/packages/simplex/src/strategies/fx.ts:784) — cross-chain under-fill guard is now unreachable; the generic guard above subsumes it.
- [[core/boot.ts:595](https://claude.ai/epitaxy/sdk/packages/simplex/src/core/boot.ts:595)](sdk/packages/simplex/src/core/boot.ts:595) — `stop()` closes a caller-supplied data store, contradicting the ownership rule applied to scanners; breaks two solvers sharing one store.
- [[scanner/hyperbridge-scanner.ts:177](https://claude.ai/epitaxy/sdk/packages/simplex/src/scanner/hyperbridge-scanner.ts:177)](sdk/packages/simplex/src/scanner/hyperbridge-scanner.ts:177) — `close()` doesn't await an in-flight connect, so `stop()` can resolve with a WsProvider still retrying for up to 30s.
- [[bin/simplex.ts:131](https://claude.ai/epitaxy/sdk/packages/simplex/src/bin/simplex.ts:131)](sdk/packages/simplex/src/bin/simplex.ts:131) — pair add/remove is duplicated between the CLI's `marketCapabilities` and `PairController`, and the two have already diverged (validation vs. `strategyTypes`, `pairIndex` vs. `index`).
- [[data/recorder.ts:145](https://claude.ai/epitaxy/sdk/packages/simplex/src/data/recorder.ts:145)](sdk/packages/simplex/src/data/recorder.ts:145) — the `.catch` also covers `emit`, so a throwing listener is logged as a failed store write.
- [[core/event-monitor.ts:47](https://claude.ai/epitaxy/sdk/packages/simplex/src/core/event-monitor.ts:47)](sdk/packages/simplex/src/core/event-monitor.ts:47) — unused `clientManager` field and dead imports left behind by the scan-loop move; `getLogger` now unused in eight files, `configureLogger` in `UiServer`.

seunlanlege:
All ten confirmed and fixed in c00b99d5f. Point-by-point:

- **fan-out.ts:69** — real, and the sharpest catch of the set. The drop policy was designed for a slow consumer, but a synchronous catch-up burst overflows the queue before the drain microtask ever runs, and the cursor advance makes the loss permanent for every subscriber. The scanner's publish loops now yield to the macrotask queue every 250 events, which flushes every consumer's drain completely between chunks.
- **CacheService.ts:141** — real. `partialFills` is swept on its own timestamps now instead of piggybacking on `fundingPrepends`, which most orders never set.
- **sqlite/bids.ts:104** — real, and nastier than it reads: the retracted row winning a same-second tie suppresses retraction of the live deposit permanently. `ORDER BY created_at DESC, id DESC` restores insertion order and matches the memory store.
- **fx.ts:542** — real. The probe is lazy now: cheap eligibility (same-chain, no calldata) is computed upfront, and the per-output contract read runs at most once per evaluation, only when an under-fill is actually on the table.
- **fx.ts:784** — real, deleted. The generic under-fill guard returns first in every reachable state.
- **boot.ts:595** — real, and a genuine ownership contradiction. `ownsData` now says whether shutdown may close the store: `Simplex.start` sets it only for the `MemoryDataStore` it defaults, and the CLI closes the store it opens itself. Two solvers sharing one store are safe.
- **hyperbridge-scanner.ts:177** — real. `close()` awaits an in-flight connect before tearing down, so `stop()` can no longer resolve while a WsProvider is still dialling out the connect timeout.
- **bin/simplex.ts:131** — real, and fixed structurally rather than by re-syncing: `marketCapabilities` is now a thin adapter over `simplex.pairs` — the same `PairController` path the library exposes — keeping only the dashboard's strategy-index addressing. The capability contract went async to allow it, so the two cannot diverge again.
- **recorder.ts:145** — real. The write's rejection handler and the emit's try/catch are separate now; a throwing listener logs as a listener bug, not a failed store write.
- **event-monitor.ts:47** — real, and your count was conservative: the sweep found unused `getLogger` imports in fifteen files, not eight. All gone, along with `clientManager` (field, param, assignment, import) and UiServer's comment-only `configureLogger`.

CI is green on the branch with all of this in.

Wizdave97:
**Five new findings, all introduced by the fix commit:**

- [[scanner/chain-scanner.ts:250](https://claude.ai/epitaxy/sdk/packages/simplex/src/scanner/chain-scanner.ts:250)](sdk/packages/simplex/src/scanner/chain-scanner.ts:250) — the two new `await` yields have no `this.stopped` recheck, breaking the invariant stated four lines above them and `stop()`'s "publishes nothing" guarantee; holding the mutex across the yields also makes the 5s drain timeout more likely to fire.
- [[core/event-monitor.ts:2](https://claude.ai/epitaxy/sdk/packages/simplex/src/core/event-monitor.ts:2)](sdk/packages/simplex/src/core/event-monitor.ts:2) — the `clientManager` field was removed but its import stayed, along with four other dead SDK imports. The commit message says the import was removed; it wasn't.
- [[bin/simplex.ts:128](https://claude.ai/epitaxy/sdk/packages/simplex/src/bin/simplex.ts:128)](sdk/packages/simplex/src/bin/simplex.ts:128) — the PairController refactor orphaned `assetRegistry`/`balanceTokens` and the `adminStrategyFor`/`tradingPairFrom`/`normalizeSymbol` imports.
- [[services/server/UiServer.ts:934](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/server/UiServer.ts:934)](sdk/packages/simplex/src/services/server/UiServer.ts:934) — `pairs.splice()` now edits a detached array because the controller reassigns `config.pairs`; correct today only by accident, and a double-removal the moment the controller splices in place.
- [[bin/simplex.ts:163](https://claude.ai/epitaxy/sdk/packages/simplex/src/bin/simplex.ts:163)](sdk/packages/simplex/src/bin/simplex.ts:163) — the UI's Stop button calls `runtime.shutdown` + `process.exit` directly, so it bypasses the store close that moved into the signal handler; ctrl-C and UI Stop now diverge.

Wizdave97 (approved):

seunlanlege:
All five confirmed and fixed in 696252576 — and worth saying plainly: all five were regressions from the previous fix commit. Point-by-point:

- **chain-scanner.ts:250** — right on both halves. The yields had no `stopped` recheck (the same await-interleaving bug that commit was fixing elsewhere), and holding the mutex across them lengthened the drain race. Every yield rechecks now, an aborted publish also skips the cursor write so the undelivered remainder gets rescanned, and stop() releases the mutex within one 250-event chunk. Pinned by a new test — 600-event burst, stop() mid-burst, assert silence after stop and an unmoved cursor — verified to fail with the recheck removed.
- **event-monitor.ts:2** — right, and the cause is instructive: the import reads `from "@/services"`, the guarded removal matched `@/services/ChainClientManager`, so it silently no-oped and the commit message shipped the claim anyway. Rather than hand-edit a third time, `noUnusedImports` is now a Biome error with a `pnpm lint` step in test-sdk.yml — the sweep it drove found 17 dead imports across 12 files, so your count was again conservative. `noFocusedTests` rides along and immediately caught a live `it.only` in usdt0.test.ts that had been silently skipping the file's other test.
- **bin/simplex.ts:128** — right; the orphaned destructure and imports are gone (the imports via the rule, the destructure by hand — the rule doesn't cover variables).
- **UiServer.ts:934** — right, and "correct only by accident" was the accurate framing. Config mutation now has one owner: with a capability present the controller's reassignment is the only write, and the server's splice/assignment runs solely in the no-capability fallback. The capability mocks in ui-server.test.ts now perform the contract, which doubles as documentation of it.
- **bin/simplex.ts:163** — right. One teardown now: the dashboard's Stop runs the identical path as SIGINT/SIGTERM via a `stopAll` handed to `operatorContextFrom`, so the two exits cannot diverge again.

Scope note: enabling Biome's full recommended set would flag ~500 pre-existing style violations (`noNonNullAssertion` ×259, `noExplicitAny` ×120, …). Deliberately not imposed in this commit — the config enables exactly the two rules above; the ratchet is a separate decision.

CI is green on 696252576, including the first run of the new lint step.

royvardhan:
`core/filler.ts:120` — if the bidding connection to Hyperbridge can't be established within the 30s race, the filler fails silently but looks like it's running.

`this.hyperbridge` becomes a permanently rejected promise (nothing resets it), and it's awaited on the fill path (`:867`), on every retraction including the sweep (`:991`), and by phantom start (`:1025`). Nothing at boot awaits it, so `Simplex.start()` still resolves and `status()` reports healthy. The losing `IntentsCoprocessor.connect` is also never disconnected, so its `WsProvider` keeps retrying with no owner and `stop()` (`:497`) can't reach it.

Flow — solver started with a caller-supplied `hyperbridgeScanner`, `hyperbridgeWsUrl` unreachable at boot:

1. `Simplex.start()` skips its own scanner connect, the race starts, boot completes, `status()` looks healthy.
2. t+30s: race rejects, logged once as "Hyperbridge connection failed".
3. Every order after that is scanned and fully evaluated, then `:867` throws "Timed out connecting to Hyperbridge" and is swallowed as "Order execution task failed". No bid, ever — the node coming back changes nothing.
4. Unretracted bids in the store fail the same way on each sweep; deposits stay locked.
5. `await simplex.stop()` resolves, but the orphaned provider keeps the host process alive.

`HyperbridgeScanner.start` handles the same race correctly (owns the provider, `provider.disconnect()` on race loss). Suggest awaiting this connect in `initialize()` so a timeout rejects `Simplex.start()` like the scanner's `create()` does, and owning the provider so a lost race can be disconnected.

royvardhan:
`strategies/fx.ts:706` — a leg the wallet can't cover at all is skipped, but the order isn't; it still goes out with a `0` for that leg, and on-chain a zero leg is a partial fill.

The `finalOutputAmount === 0n` path pushes `{ amount: 0n }` and `continue`s above the new eligibility checks at `:761-800`, so `partialEligible()` is never asked and `partialFill` stays `false`. `:804` only refuses when every leg is zero, so the cached outputs go out verbatim.

Flow — same-chain order on Base, outputs `[1000 USDC, 1000 USDT]`, output calldata attached; solver holds USDC, no USDT, no vault:

1. Leg 0 (USDC): fully covered, no flags.
2. Leg 1 (USDT): `finalOutputAmount = 0n` → `:706` pushes `0n`, `continue`. The calldata check in `partialEligible()` never runs.
3. Post-loop gate passes (leg 0 is non-zero); GATE 1 runs the full-fill rule on `order.fees`; `setPartialFill(order.id, false)`; the bid is submitted with `[1000 USDC, 0 USDT]`.
4. On chain (`IntrinsicIntents.sol`): leg 1 `solverAmount == 0 && remaining > 0` → `isFullyFilled = false` (`:74-77`); calldata + not fully filled → `revert PartialFillNotAllowed` (`:118`). The whole call unwinds, nothing fills, gas paid.

Same root, other shapes: cross-chain reverts `InvalidInput` (`ExtrinsicIntents.sol:106`); a plain same-chain order fills as a partial while GATE 1 and `totalProfit` counted `order.fees` that a partial never receives (`_withdraw(..., finalize=false)`).

Not a regression — main had the same `continue`-before-check ordering — but the PR states it closed this ("a same-chain calldata order the wallet couldn't cover was bid on and reverted with `PartialFillNotAllowed`", and the `:756-760` comment "any shortfall makes this an under-fill, whatever caused it"). It's closed for a leg that's short, not for one that's empty. Suggest running `partialEligible()` before the `continue` at `:706` (setting `partialFill = true` when it passes), or moving the zero-push below the `finalOutputAmount < output.amount` block so `0n` takes the same path.

royvardhan:
Other things look good

PR #1125, opened 2026-08-13
[indexer]: Intent gateway specific daily filled volumes

### LINKED ISSUE #1085: [indexer] Intent gateway specific daily filled volumes
The current daily volumes we track are tied to a filler's address, we should also index daily volumes independent of filler.

Current result

```
"data": {
    "dailyVolumeUSDs": {
      "nodes": [
        {
          "last24HoursVolumeUSD": "2342670016941045000000",
          "id": "IntentGatewayV3.FILLER.0x13E41CdE1D55880cbe031c69f206C2E9BC3c94C2.EVM-8453.2026-07-30",
          "createdAt": "2026-07-30T00:00:00"
        },
        {
          "last24HoursVolumeUSD": "992770366500000000",
          "id": "IntentGatewayV3.FILLER.0xb52c73CB71428a2a727044BCC82ed48E545AE468.EVM-8453.2026-07-30",
          "createdAt": "2026-07-30T00:00:00"
        },
        {
          "last24HoursVolumeUSD": "2628045320000000000000",
          "id": "IntentGatewayV3.USER.EVM-8453.2026-07-30",
          "createdAt": "2026-07-30T00:00:00"
        }
      ]
    },
```

We should have an entry that is just the intent gateway address recording the daily volume filled

royvardhan:
Daily volumes were only tracked per filler. This records a filler-independent `IntentGatewayV3.FILLED` entry through the same `VolumeService` call site and pricing as the per-filler records, so the gateway entry equals the sum of filler entries per chain per day. Queryable from the existing `dailyVolumeUSDs` query as `IntentGatewayV3.FILLED.EVM-8453.2026-07-30`.

On a resumed database the new series is initialized from the volumes already indexed for individual fillers: the first fill per chain after deploy backfills the gateway cumulative record and all historical daily buckets from the existing `IntentGatewayV3.FILLER.*` records (`VolumeService.seedAggregateVolume`), before that fill's own updates so it is never double-counted. A fresh from-genesis reindex seeds nothing and behaves as before.

Also adds the first unit tests for `VolumeService`.

Introduces AI workflow docs under `sdk/packages/indexer/docs/ai/`:

- `ChangeLog.md` — a running log of every AI-assisted change in the package: date, what changed and why, files touched. Separate from the changesets `CHANGELOG.md` release logs.
- `Decisions.md` — non-obvious choices with the alternatives considered and why they lost. Seeded with the two decisions from this change: reusing `DailyVolumeUSD` instead of a new indexed entity, and recording at the same call site as the filler volume so gateway totals stay equal to the sum of filler totals.
- `Flow.md` — how verified code paths execute end to end. Seeded with the OrderFilled volume indexing flow: the handler, `updateOrderStatus` including the pending-status early return, the `VolumeService` upserts, and how the parallel `recordOrderVolume` path can disagree with it.

A rule in the root CLAUDE.md makes updating these files part of every change in a package that has them, and has the first substantial task in each remaining sdk package create its own seeded set.

Closes #1085


Wizdave97:
<html><head></head><body><p><strong>Short answer: the daily seeding is valid; the cumulative seeding is valid at t₀ but is summed on a basis the code does not maintain afterwards.</strong></p>
<p>I checked the logic against the live Nexus data. Reconstructing what the seed would write today:</p>

chain | seed from FILLER cumulatives | seed from FILLER dailies
-- | -- | --
EVM-1 | $146.67 | $146.67
EVM-56 | $3.38 | $3.38
EVM-8453 | $235,711.99 | $235,711.99


<p>66 daily buckets, 418 <code>DailyVolumeUSD</code> rows and 38 <code>CumulativeVolumeUSD</code> rows total — so the paging scan is 5 pages, and the "tables are small" claim in <code>Decisions.md</code> holds.</p>
<h2>Daily buckets: valid</h2>
<p><code>FILLED</code> daily = Σ <code>FILLER.*</code> daily holds by construction — both go through <code>updateDailyVolume</code>, which has no same-timestamp guard, from the same <code>outputUSD.total</code> on the same code path. The prefix and chain filters are correct: <code>record.id.slice(0, -11).endsWith(chainSuffix)</code> strips exactly <code>.YYYY-MM-DD</code> (11 chars), and <code>.EVM-1</code> can't collide with <code>.EVM-11155111</code> under <code>endsWith</code>. Re-seeding is idempotent for dailies because <code>create().save()</code> overwrites rather than adds, and <code>IntentGatewayV3.FILLED</code> doesn't match the <code>IntentGatewayV3.FILLER.</code> prefix, so the aggregate never sums itself.</p>
<h2>Cumulative: the seed and the post-seed path count differently</h2>
<p><a href="https://claude.ai/epitaxy/sdk/packages/indexer/src/services/volume.service.ts">volume.service.ts:36</a> skips the addition when <code>lastUpdatedAt === timestamp</code>. That guard fires <strong>per record</strong>, so it means different things for a component and for an aggregate:</p>
<ul>
<li><code>FILLER.0xaaa</code> collides only when <em>that filler</em> fills twice in one block.</li>
<li><code>FILLED</code> collides when <em>any two fills on the chain</em> land in one block — including two different fillers.</li>
</ul>
<p>So from the first multi-filler block after seeding, <code>FILLED</code> cumulative &lt; Σ <code>FILLER</code> cumulative, and the gap only widens. The seed sets them equal at t₀ by summing component cumulatives (<a href="https://claude.ai/epitaxy/sdk/packages/indexer/src/services/volume.service.ts">:106</a>), which quietly assumes an invariant the running code then breaks. The PR body's claim that "the gateway entry equals the sum of filler entries per chain per day" is true for the daily series and false for the cumulative one.</p>
<p><code>Decisions.md</code> notes the same-block quirk but frames it as pre-existing and neutral. It isn't neutral here: the aggregate's collision surface is strictly larger than any component's, and this PR is what introduces an aggregate on that path.</p>
<p>Empirically it hasn't bitten yet — for <strong>every</strong> existing series, including <code>IntentGatewayV3.USER</code> (already a chain-wide aggregate, $226,189.45 on Base), Σ dailies equals the cumulative exactly. So this is latent, not currently observable. That's also why the seed numbers match either way in the table above.</p>
<p><strong>Fix, which also deletes code:</strong> drop the cumulative scan at <a href="https://claude.ai/epitaxy/sdk/packages/indexer/src/services/volume.service.ts">:97–110</a> entirely and derive the cumulative seed from <code>dailyTotals</code> — <code>Σ bucket.volume</code>, with <code>lastUpdatedAt = max(bucket.lastUpdatedAt)</code>. The daily rows are guard-free, so they're the more faithful record of "every fill", the max timestamp still clears the same-timestamp guard for the triggering fill, and it's one full-table scan instead of two. It doesn't fix the forward-going drift — for that you'd need <code>updateCumulativeVolume</code> to stop guarding, which the PR rightly declines to change — but at least the seed then matches the daily series it's paired with.</p>
<h2>Smaller points</h2>
<ul>
<li><strong><code>readAllPages</code> already exists.</strong> <a href="https://claude.ai/epitaxy/sdk/packages/indexer/src/utils/store.helpers.ts">store.helpers.ts</a> has <code>readAllPages()</code> and <code>STORE_PAGE_SIZE = 100</code> with a comment describing exactly this hazard. The PR hand-rolls the loop twice and declares a local <code>PAGE_SIZE = 100</code> (<a href="https://claude.ai/epitaxy/sdk/packages/indexer/src/services/volume.service.ts">:93</a>).</li>
<li><strong>Empty <code>getByFields</code> filter is fine</strong>, but note d1e9d681 deliberately moved <code>PendingStatusService</code> off <code>getByFields([], ...)</code> — not because it's invalid (it isn't; x-sequelize returns <code>""</code> for the empty case) but because a zero-row result gives no diagnostic signal. Same tradeoff applies here.</li>
<li><strong>New failure mode in the status path.</strong> The seed runs inside <code>updateOrderStatus</code>, which the handler wraps in try/catch. A store error during the scan now swallows the whole fill's status update, points, and user activity — previously nothing in that stretch could fail on a read of unrelated tables. Consider its own try/catch: a failed seed should not cost an order's status.</li>
<li><strong>The at-risk case is untested.</strong> The tests cover seed-then-update, no-op-on-marker, and fresh-reindex. None covers two fillers at the same timestamp after seeding — the one scenario where the aggregate and the sum of components diverge. That's the test worth adding, whichever way you decide to resolve it.</li>
<li><strong><code>cumulativeLastUpdatedAt</code> can't equal the triggering fill's timestamp</strong> because history is written by the pre-deploy code and the resume point is strictly after the last indexed block. Correct, but it's load-bearing and only stated as "so the guard does not swallow that fill" — worth saying <em>why</em> it's safe.</li>
</ul></body></htm

Wizdave97 (approved):

PR #1126, opened 2026-08-13
[runtime]: Reject purchases whose app identifier exceeds AppKey

Wizdave97:
`app` is stored as an `AppKey` (`BoundedVec<u8, ConstU32<32>>`) via `AppKey::truncate_from`, so bytes past the bound were silently discarded rather than rejected. A tier's price is flat per month and does not scale with body length, so those bytes cost the buyer nothing extra while making the delivered message arbitrarily longer — and the credited subscription was keyed on a truncation of what the buyer actually asked for.

Reject at decode time in `PurchaseMessage::try_from`, and mirror the check in `BandwidthManager.purchase` behind `MAX_APP_LENGTH` so the purchase fails before the payer is charged for a message hyperbridge would drop.


seunlanlege (approved):

seunlanlege:
This is a nothing-burger imo

Issue #1132, opened 2026-08-14
[simplex]: MPCVault TYPE_SIGN_TYPED_DATA signatures don't verify against the submitted payload


seunlanlege:
## Summary

Signatures returned by MPCVault's `CreateSigningRequest` with `evm_message.type = TYPE_SIGN_TYPED_DATA` do not verify against the EIP-712 digest of the submitted payload — or against any digest derivable from it that we could find. Every bid UserOp and delegation-via-bundler UserOp signed through this path since the MPC-backed filler upgraded to typed-data signing (post #1030 builds) fails on-chain validation with `AA24` / "Invalid account signature". Signatures from the same vault through the **raw-message path** (`raw_message` + `ECDSA_HASH_FUNCTION_USE_MESSAGE_DIRECTLY`) validate and execute on-chain.

A hotfix routes all MPC typed-data signing through local hashing + the raw-message path. This issue tracks the underlying investigation.

## Affected setup

- Solver EOA / MPCVault wallet: `0xb52c73CB71428a2a727044BCC82ed48E545AE468` (EIP-7702-delegated to `SolverAccount` at `0xfCd233b937D7622AAc63ced3C9A1A12F4a6B64E3` on Base)
- Filler substrate account: `13Vq4AEmUdFe61GB2Vrx58j3d9S4bjscUB3hDFDvfSj5YsBf`
- Client: `MpcVaultService` (`sdk/packages/simplex/src/services/wallet/mpcvault.ts`), gRPC stubs generated from `mpcvault/mpcvaultapis` @ `6ea3272` (verified current: the two upstream commits since are additive only)

## Evidence (2026-08-14)

Data sources: `fillerBids` / `iOrderV3s` / `iOrderV3Fills` on the nexus indexer, `intents_getBidsForOrder` on the nexus RPC, Base RPC.

1. **The vault and key are fine.** The EIP-7702 delegation direct tx and its authorization tuple — both signed via the raw-message path — landed on Base (the EOA carries the `0xef0100…` delegation designator; nonce 30).
[...]

PR #1134, opened 2026-08-14
[sdk]: emit standard eth_signTypedData_v4 payload for userop signing

royvardhan:
`packedUserOpTypedData` omitted `EIP712Domain` from `types` and passed `chainId` as a bigint. Local signers hash this fine, but backends that hash the typed data server-side from JSON do not: viem's `serializeTypedData` strips the domain to `{}` when `types.EIP712Domain` is absent (Turnkey path), and MPC Vault builds an empty domain separator, so the signature recovers to the wrong address and userops fail ERC-4337 validation with AA24 "invalid account signature".

Emit the canonical v4 payload instead: `EIP712Domain` listed in `types`, `chainId` as a number. The digest is unchanged for local signers (`hashTypedData` parity with `computeUserOpHash` is covered by the existing tests).

### DIFF
--- a/sdk/packages/simplex/package.json
+++ b/sdk/packages/simplex/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@hyperbridge/simplex",
-	"version": "0.9.4",
+	"version": "0.9.5",
 	"license": "Apache-2.0",
 	"description": "IntentGateway simplex package for hyperbridge",
 	"main": "dist/index.js",
--- a/sdk/packages/sdk/package.json
+++ b/sdk/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@hyperbridge/sdk",
-	"version": "2.8.2",
+	"version": "2.8.3",
 	"description": "The hyperclient SDK provides utilities for querying proofs and statuses for cross-chain requests from HyperBridge.",
 	"type": "module",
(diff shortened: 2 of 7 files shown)


sdk/packages/simplex/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@hyperbridge/simplex",
-	"version": "0.9.4",
+	"version": "0.9.3",
```
Wizdave97:
```suggestion
	"version": "0.9.5",
```

Wizdave97:
Also revert this commit `c96c995bf141a548f983b4165f8c8f890f1f7403` when you merge main into your branch

seunlanlege:
Empirically validated against a live MPCVault vault: the canonical payload (`EIP712Domain` in `types`) makes `TYPE_SIGN_TYPED_DATA` sign the correct EIP-712 digest — recovery-asserted via `pnpm test:mpcvault` on this branch and an isolated A/B probe (results in https://github.com/polytope-labs/hyperbridge/issues/1132#issuecomment — see issue). The missing `types.EIP712Domain` alone was the trigger; string-vs-numeric `chainId` had no effect, though emitting a number is still the more canonical shape.

seunlanlege (approved):

seunlanlege:
Tested & working

PR #1146, opened 2026-08-18
[runtime]: Add aura as a fallback for calculating timestamp in parachain consensus client

### LINKED ISSUE #1135: [runtime]: Add aura as a fallback for calculating timestamp in parachain consensus client
Our current parachain consensus client expects the parachain to have pallet-ismp. Polkadot hub does not.

Let's add a fallback using the previous technique of calculating timestamp using aura slots when pallet ismp digests is not found. 
This fallback should exclusively  only be allowed for polkadot hub.

dharjeezy:
Closes #1135

Polkadot Hub doesn't run pallet-ismp, so its headers never carry the ismp timestamp digest and the parachain consensus client rejected every one of them. The client now derives the block time from the aura slot when that digest is missing, reading the slot duration from storage so it can be retuned by governance, and only Asset Hub is allowed that route. A derived timestamp that runs ahead of the local clock is rejected instead of stored, since that is what a stale slot duration looks like.

## Notes

Digest parsing now goes through `fetch_overlay_root_and_timestamp` rather than a second copy of the same loop, and that helper only reaches for the slot when the ismp timestamp digest is absent, instead of leaving it to digest ordering.

Polkadot Hub currently authors on 24000ms slots with several blocks per slot, so the derived timestamp is quantised to the slot boundary. It is always behind the real block time, never ahead, which is the safe direction for timeout proofs.


modules/ismp/state-machines/substrate/src/lib.rs
```diff
@@ -411,5 +417,14 @@ pub fn fetch_overlay_root_and_timestamp(
 		};
 	}
 
+	// chains without pallet-ismp never deposit the timestamp digest, so their block time can
+	// only come from the slot they were authored in
+	if digest_result.timestamp == 0 {
+		digest_result.timestamp = slot
+			.map(|slot| Duration::from_millis((*slot).saturating_mul(slot_duration)).as_secs())
+			.unwrap_or_default();
```
seunlanlege, line 422: should probably be an error to neither have slot or timestamp digest

modules/ismp/clients/parachain/client/src/consensus.rs
```diff
@@ -148,41 +148,30 @@ where
...
 			}
 
+			// A slot always begins before the block it authored, and that block is already
+			// included on the relay chain, so a timestamp ahead of our own clock means the
+			// configured slot duration no longer matches the parachain.
+			if timestamp_from_slot && timestamp > host.timestamp().as_secs() {
+				Err(Error::Custom(format!(
+					"Slot duration for parachain {id} is stale, derived timestamp {timestamp} is in the future"
+				)))?
+			}
```
seunlanlege: this imo, is unneccessary

modules/ismp/clients/parachain/client/src/consensus.rs
```diff
@@ -148,40 +148,16 @@ where
...
-					_ => {},
-				};
-			}
+			// Asset Hub doesn't run pallet-ismp, so it never deposits the timestamp digest and its
+			// block time is derived from the aura slot instead. No other parachain is allowed
+			// that route, hence the slot duration is only ever read for Asset Hub.
+			let slot_duration = (id == ASSET_HUB_PARA_ID)
+				.then(|| SlotDurations::<T>::get(id))
+				.flatten()
+				.unwrap_or_default();
```
seunlanlege: if the id == asset hub then zero slot duration is an error.

seunlanlege (approved):

PR #1147, opened 2026-08-18
[evm, simplex]: PERMIT2 mode for SimplexPaymaster

### LINKED ISSUE #1071: [evm, simplex]: Permit2 mode for SimplexPaymaster
## Motivation

`SimplexPaymaster` supports two authorization modes: **PERMIT** (EIP-2612, executed during validation — fully gasless) and **APPROVE** (pre-existing on-chain allowance). On chains whose stablecoins lack EIP-2612 — BSC's pegged USDC (`0x8ac76a51…`) and USDT (`0x55d39832…`) — only APPROVE mode works, which has two costs:

1. **Native-funded bootstrap**: the initial `approve(paymaster, cap)` is a plain transaction the solver EOA must pay for — on the very chains where the paymaster is supposed to remove the native-gas requirement.
2. **Recurring re-approvals**: the allowance is deliberately capped (currently 5 tokens); every refill is another native-funded transaction, making native gas a *permanent* operational dependency on approve-mode chains.

Permit2 cannot remove cost (1) — its own `approve(Permit2, max)` on the token is equally a funded transaction, since the token contract knows nothing about Permit2. But it eliminates cost (2) entirely and improves the risk shape:

- **One** funded approval per token for the account's lifetime; every subsequent authorization is an off-chain Permit2 signature carried in the userOp.
- Per-op, amount-scoped, deadline-boxed authorization (`permitTransferFrom`) instead of a reusable standing allowance to the paymaster.
- One uniform flow for every non-permit ERC-20, collapsing the permit/approve fork in `buildSimplexPaymasterData`.

## Scope

- **Contract**: add a PERMIT2 mode (e.g. `0x02`) to `SimplexPaymaster` — validate the Permit2 signature and pull the fee via `permitTransferFrom` (canonical Permit2: `0x000000000022D473030F116dDEE9F6B43aC78BA3`, deployed on all supported chains). Paymaster changes need the corresponding audit attention.
- **Client** (`buildSimplexPaymasterData`): prefer PERMIT, then PERMIT2 (when the token's Permit2 allowance is in place), then APPROVE; sign the per-op Permit2 payload.
- **Bootstrap**: `ensureCappedApproval` becomes a one-time `approve(Permit2, max)` per token on non-permit chains.
- Redeploy + registry update for the supported mainnets.

royvardhan:
Closes: #1071

Adds mode `0x02 PERMIT2` to `SimplexPaymaster`: the prefund is pulled through Permit2 `permitTransferFrom` (owner = `userOp.sender`, spender = paymaster, single-use nonce, deadline surfaced as `validUntil`), so tokens without EIP-2612 need one lifetime `approve(Permit2, max)` instead of a capped allowance to the paymaster refilled with native-funded approvals. No storage changes; the live proxies remain upgradeable.

Client (`simplex` 0.10.0): selection is PERMIT, then PERMIT2 when the token is approved to Permit2, then APPROVE while a legacy allowance lasts, then a one-time `approve(Permit2, max)` bootstrap. Mode 2 is only emitted against deployments that expose `PERMIT2()`, so it is safe before the redeploy lands. `forceApproveMode` became `skipPermit`.

### ERC-7562: the anticipated blocker

Bundler validation rules only let a paymaster write storage associated with the sender or with itself. Permit2's `nonceBitmap[owner][word]` and the token's `allowance[owner][Permit2]` are neither (nested mappings hash the owner on the inner level), so a spec-enforcing bundler could reject mode 2 at `eth_sendUserOperation` no matter how correct the contract is, and staking cannot fix it. Permit2 also executes `TIMESTAMP` for the deadline check.

Why it was expected to work anyway: this paymaster already reads `block.timestamp` (equally banned) in `_getOraclePrice` in every mode and is live through Alchemy on BSC, so the bundlers in use are not running the ERC-7562 tracer. Fork tests could not settle this (they execute the EVM, not bundler policy), so it was probed live: a fresh `SimplexPaymaster` on Base Sepolia (`evm/script/SimplexPaymasterPermit2Probe.s.sol`) with Alchemy's bundler accepted mode 2 for both a fresh EOA (Permit2 ecrecover path) and a delegated account (ERC-1271 path through `SolverAccount`), with `Permit2Executed` landing on-chain. Pimlico was not probed. If a bundler ever rejects mode 2, PERMIT/APPROVE remain intact and the client can fall back to refilling the APPROVE allowance inside sponsored ops.

The probe also surfaced that the first op right after a one-confirmation bootstrap approve is rejected `AA33` (the bundler's simulation node lags the RPC); the bootstrap now waits for two confirmations.

### Security review fixes

A threat-model audit of the paymaster ran alongside this work, executed against mainnet forks rather than read-only. It found no path for a user, bundler, relayer or order creator to extract value, but two issues worth fixing before the redeploy are included here.

**Stake was permanently unrecoverable.** `addStake` was ungated while `unlockStake`/`withdrawStake` route through `_authorizeWithdraw()`, which reverts unconditionally, and no governance request kind covered stake. Confirmed against the live deployments: 0.1 native is already stuck on each of Ethereum, BSC and Arbitrum, and any unprivileged address could call `addStake{value: 1 wei}(type(uint32).max)` to push the unstake delay from one day to 136 years, defeating even a future upgrade. Adds `RequestKind.UnlockStake` / `WithdrawStake` (both empty-payload, always paying out to the treasury, with matching pallet extrinsics) and gates `addStake` to the treasury.

**The postOp gas cap was burning half the margin.** The EntryPoint penalises the unused part of `paymasterPostOpGasLimit` after fixing the cost handed to `postOp`, so that penalty is never billed to the user, and it waives the penalty entirely while `gasLimit <= gasUsed + 40_000`. The SDK pinned the limit at the contract's 100k cap while postOp needs ~8-12k. `MAX_POST_OP_GAS_LIMIT` is now 40,000 — the largest unconditionally penalty-free value — with a 30,000 floor closing the refund-underflow window an unbounded-below limit left open. Fork-measured margin per sponsored op rises from 9,053,104,589,778 to 21,354,113,277,112 wei, and is now identical at the cap and the floor. The SDK's shared `POST_OP_GAS_LIMIT` splits per paymaster, since Circle's is a different contract whose postOp was never measured here.

Deferred deliberately: per-feed oracle staleness, a non-reverting refund, governance message epochs, and a treasury pause (which would reintroduce a privileged local role, so it is a design call rather than a fix).

Tests: foundry unit cases plus fork suites (Ethereum in CI via `MAINNET_FORK_URL`, BSC locally) covering the full `handleOps` round trip against the real Permit2, the gas-accounting measurements, and the stake lock verified against the live deployed contracts. The Base Sepolia probe script and its env-gated live test are included but never run in CI.

Redeploy on the 5 chains and registry update to follow.


royvardhan:
## Security review of the PERMIT2 path

Backed by fork tests against the real Permit2, USDT/USDC, the real EntryPoint v0.8 and the live deployed contracts (`Permit2CompromiseForkTest`, `SimplexPaymasterPermit2ForkTest`, `SimplexPaymasterGasGriefTest`, `SimplexPaymasterStakeLockForkTest`).

### 1. Bugs in the added code
Reviewed `_prefund` (mode 2), `_fetchDetails`, `_parsePermit2Data` and the client selection logic. No exploitable defect found. The invariants that make it safe:
- The Permit2 call runs only after `_fetchDetails` confirms the token is registered and active, so validation never calls into an attacker-chosen token.
- `owner` is always `userOp.sender` and `to` is always `address(this)` — neither is reachable from `paymasterData`.
- `prefundAmount > permitAmount` reverts before any external call (nonce untouched), and Permit2 independently enforces `requestedAmount <= permitted.amount`, so even a manipulated oracle can't pull more than the solver signed.
- Failures revert (never silently succeed); the refund path is unchanged and pays the sender.

`_prefund` now charges the `IERC20 token` that `_fetchDetails` validated rather than re-deriving it from the mode byte, so it no longer depends on the base contract's call ordering for that guarantee.

### 2. Can a compromised paymaster drain a filler who approved `max` to Permit2?
**No.** Proven in `Permit2CompromiseForkTest`:
- **The max approval is inert by itself.** It approves the *token* to Permit2, which grants no pull right on its own — Permit2's signature-less `transferFrom` checks a *separate internal* allowance the client never sets, so the paymaster's Permit2 allowance is `0` and a signature-less drain reverts.
- **The attacker can't forge a pull.** Every `permitTransferFrom` needs a fresh filler signature over `(token, amount, spender, nonce, deadline)`; a permit signed with the attacker's key reverts `InvalidSigner`.
- **Exposure is bounded to what the filler actually signed.** A malicious paymaster can redirect an *outstanding, unexpired, unused* permit (e.g. a losing bid) to itself — but only up to its signed amount (inflating `requestedAmount` reverts `InvalidAmount`) and only once (nonce spent). At rest, with no pending bids, exposure is **zero**.

This is strictly better than APPROVE mode, where a fixed allowance sits exposed to the paymaster at all times and auto-refills. "Compromised paymaster" here also means a malicious Hyperbridge-governance upgrade, not a single stolen key.

One caveat worth stating: for a **policy-constrained signer** (Turnkey/MPC), if the policy restricts transactions but not typed-data signing, the ceiling on what a single signature can move rises from a $5 EIP-2612 permit to the account's full balance, since nothing on-chain caps `permitAmount`. For a raw private-key signer this changes nothing.

### 3. Wider audit — two fixes folded into this PR
The review was extended to the whole contract and its usage, which found no value-extraction path but did surface issues worth fixing before the redeploy. Two are now included here:

- **Stake was permanently unrecoverable.** `addStake` was ungated while `unlockStake`/`withdrawStake` route through `_authorizeWithdraw()`, which always reverts, and no governance request kind covered stake. 0.1 native is already stuck on each of Ethereum, BSC and Arbitrum, and anyone could call `addStake{value: 1 wei}(type(uint32).max)` to stretch the unstake delay to 136 years — defeating even a later upgrade. Now has `UnlockStake`/`WithdrawStake` request kinds plus pallet extrinsics, and `addStake` is treasury-gated.
- **The postOp gas cap was burning half the per-op margin.** The EntryPoint penalises unused `paymasterPostOpGasLimit` after fixing the cost handed to `postOp`, so it was never billed to the user; the penalty is waived entirely at or below 40k. `MAX_POST_OP_GAS_LIMIT` is now 40,000 with a 30,000 floor (which also closes a refund-underflow window). Fork-measured margin per op goes from 9,053,104,589,778 to 21,354,113,277,112 wei.

Also checked and **not** changed: inflating `callGasLimit` does not drain the paymaster — the penalty is billed straight through to the user, with margin constant to five significant figures across a 166× inflation. And soft-failing the prefund instead of reverting was considered and rejected: EntryPoint v0.8 turns both a paymaster revert (AA33) and a sig-failure (AA34) into `revert FailedOp`, so it would trade the `Permit2Failed` diagnostic for no bundle-level benefit.

Deferred deliberately: per-feed oracle staleness (one `maxOracleAge` currently covers both a 1-hour and a 24-hour heartbeat feed; USDC/USD on mainnet was measured at 77,786s against the 90,000s bound), a non-reverting refund, governance message epochs, and a treasury pause — the last reintroduces a privileged local role, so it is a design call rather than a fix.

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -44,6 +73,10 @@ interface AggregatorV3Interface {
 ///                      uint256(deadline), uint8(v), bytes32(r), bytes32(s))
 ///   Mode 0x01 (approve):
 ///     abi.encodePacked(uint8(1), address(token))
+///   Mode 0x02 (permit2):
+///     abi.encodePacked(uint8(2), address(token), uint256(permitAmount),
+///                      uint256(nonce), uint256(deadline), bytes(signature))
+///     signature is 65 bytes (r, s, v); the signed spender is this paymaster.
```
seunlanlege: lets actually prefer the full r, s, v instead of bytes

seunlanlege:
one more thing is, if simplex has to do a native EIP-7702 delegation, then execute the Permit2 unlimited allowance to the Paymaster in the same transaction

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -105,11 +152,25 @@ contract SimplexPaymaster is Initializable, HyperApp, PaymasterERC20 {
...
-    ///      caller-chosen beneficiary; the cap keeps that penalty under the
-    ///      `_postOpCost` cushion the user already pays.
-    uint256 public constant MAX_POST_OP_GAS_LIMIT = 100_000;
+    /// @dev Caps the caller-supplied postOp gas limit. The EntryPoint penalises the
+    ///      unused part of this limit *after* fixing the cost handed to postOp, so that
+    ///      penalty is never billed to the user and comes out of the `_postOpCost`
+    ///      cushion instead. The penalty is waived entirely while
+    ///      `limit <= gasUsed + PENALTY_GAS_THRESHOLD (40k)`, so any cap at or below
+    ///      40k is unconditionally penalty-free whatever postOp ends up costing.
+    uint256 public constant MAX_POST_OP_GAS_LIMIT = 40_000;
```
seunlanlege: **The narrowed postOp band is safe for a fresh-address redeploy, but not for the upgrade-in-place that stake recovery needs.**

`Decisions.md` argues the drop from 100k to 40k is safe because "new proxy addresses ship in the same `chain.ts` release as the new SDK: an old solver keeps using the old deployment and never meets the new bound." That holds only if the five proxies are *replaced*. But recovering the 0.1 native already stuck on Ethereum/BSC/Arbitrum — the stated motivation for `UnlockStake`/`WithdrawStake` — requires delivering `UpgradeContract` to the **existing** proxies, and there the new band applies immediately to every client still on `@hyperbridge/simplex` ≤ 0.11.

Concretely: governance upgrades the live Ethereum proxy `0xD4340d7466e040626383cb9cda9307ba8E081149` to reach the stuck stake. Any filler still on 0.11.2 keeps sending `paymasterPostOpGasLimit = 100_000` (the old shared `POST_OP_GAS_LIMIT`), so validation now reverts `InvalidPostOpGasLimit(100000, 30000, 40000)` → AA33 on every sponsored op, and the filler silently drops to native gas or stops filling. Same for any third-party integrator of a paymaster this contract's own natspec calls "permissionless".

Worth stating the rollout order explicitly (ship + adopt the client first, then upgrade), or splitting the stake-recovery upgrade from the band change.

sdk/packages/simplex/src/services/DelegationService.ts
```diff
@@ -99,13 +103,34 @@ export class DelegationService {
...
 		// authorization list handles that in its own `signTransaction` — see
 		// `mpcVaultSigner` — not here.
+		//
+		// When an approval is folded in, the tx calls `token.approve(spender, max)` while
+		// carrying the authorization: the delegation still lands (the authorization list is
+		// processed independently of `to`), and the approve executes with the EOA as sender.
+		// Otherwise it is a no-op self-call whose only payload is the authorization.
+		if (approval) {
+			return (await walletClient.sendTransaction({
+				to: approval.token,
```
seunlanlege, line 121: **Folding the approve into the delegation tx makes the last-resort path fail closed, permanently.**

The native set-code tx is the fallback for when the sponsored path is unavailable. Its payload is now an ERC-20 `approve`, so any approve revert reverts the authorization with it — and since `resolvePendingPermit2Approval` recomputes the same approval on the next attempt, the failure repeats forever rather than being transient.

Reachable today on Ethereum: a solver holding USDT with any pre-existing non-zero Permit2 allowance below $5 (from any other Permit2 integration). The resolver returns `{USDT, Permit2}`, `approve(Permit2, max)` hits USDT's `require(!((_value != 0) && (allowed[..] != 0)))`, the whole tx reverts, and the solver can never delegate on that chain. Also reachable via a blacklisted/paused fee token, or an approve costing more than the fixed `BATCHED_APPROVE_GAS = 60_000`.

The `try/catch` in `resolvePendingPermit2Approval` guards resolution only, not execution. Either send the approve as a separate tx after the delegation lands, or fall back to the plain self-call when the batched tx reverts.

sdk/packages/simplex/src/services/paymaster/provider/simplex.ts
```diff
@@ -79,17 +93,87 @@ export async function buildSimplexPaymasterData(
...
+	if (permit2 && permit2Allowance >= recommended) {
+		return permit2Mode()
+	}
+
+	if (paymasterAllowance >= THRESHOLD_USD * 10n ** BigInt(tokenDecimals)) {
+		return approveMode()
+	}
+
+	if (permit2) {
+		await sendFundedApprove(client, walletClient, solverAccount, tokenAddress, permit2, maxUint256)
```
seunlanlege, line 144: **`approve(Permit2, max)` reverts on Ethereum USDT when a stale non-zero allowance exists.**

This branch is reached whenever `permit2Allowance < recommended`, which includes `0 < allowance < 5e6`. USDT's approve refuses a non-zero → non-zero transition, so the bootstrap can never complete for such an account and every fill/bid on that chain aborts.

`setUp` in `SimplexPaymasterPermit2ForkTest` only ever calls `_approvePermit2` from a zero allowance, so the fork suite can't catch it.

Zeroing first (`approve(spender, 0)` when the current allowance is non-zero) covers it; the same fix applies to the batched delegation approve.

seunlanlege (commented):
Review of the PERMIT2 mode, at high effort. Ten inline findings below, roughly most-severe first.

Verified as sound and not commented on:

- The Permit2 EIP-712 payload matches `_PERMIT_TRANSFER_FROM_TYPEHASH` and Permit2's versionless `EIP712Domain`; the `v ‖ r ‖ s` paymasterData layout, the contract's re-ordering to `abi.encodePacked(r, s, v)`, and the 182-byte length agree across the contract, the SDK packer and the fork-test helper.
- `SolverAccount._rawSignatureValidation` is plain `ECDSA.recover` with no ERC-7739 wrapping, so a 65-byte signature satisfies Permit2's ERC-1271 branch for a 7702-delegated sender — the fixed-length layout costs nothing.
- 40,000 is right: EntryPoint v0.8 waives the unused-gas penalty while `gasLimit <= gasUsed + PENALTY_GAS_THRESHOLD` and applies it per-limit, so any cap at or below 40k is unconditionally penalty-free.
- Ordering in `_prefund` is correct — the amount check and the `PERMIT2.code.length` check both precede the external call, `_fetchDetails` validates the registry before `_prefund` runs, and charging the validated `token` rather than re-reading the mode byte is the right call.
- `UnlockStake`/`WithdrawStake` = 5/6 match the Solidity enum ordering, both Rust variants are appended (no SCALE break), and call indices 17/18 are free.
- `packValidationData(true, 0, validUntil)` uses OZ's `(sigSuccess, validAfter, validUntil)` ordering, and the `deadline > uint48.max → 0` mapping reads back as `type(uint48).max`.

Several things I'd otherwise have raised — random nonces over a bitmap counter, reverting rather than soft-failing the prefund, the two-confirmation bootstrap — are already argued correctly in `Decisions.md`.

The one worth resolving before merge is the first: the postOp band change and the stake-recovery upgrade want opposite rollout strategies.

sdk/packages/simplex/src/services/paymaster/provider/simplex.ts
```diff
@@ -173,34 +257,110 @@ async function buildPermitMode(
...
+	const cached = permit2Support.get(key)
+	if (cached && (cached.supported || Date.now() - cached.checkedAt < PERMIT2_SUPPORT_NEGATIVE_TTL_MS)) {
+		return cached.supported
+	}
+	let supported = false
+	try {
+		await client.readContract({ address: paymasterAddress, abi: SIMPLEX_PAYMASTER_ABI, functionName: "PERMIT2" })
+		supported = true
+	} catch {
+		supported = false
```
seunlanlege: **A transient RPC failure here is cached as "this deployment has no PERMIT2 mode" — and that has on-chain consequences.**

The bare `catch` can't tell a contract revert (old implementation) from a timeout or a 429, and the negative is cached for 5 minutes.

A solver already migrated to PERMIT2 mode has a zero paymaster allowance, so when `permit2` comes back `undefined` the builder falls all the way through to `sendFundedApprove(token, paymasterAddress, recommended)` — a native-gas transaction that re-establishes exactly the standing paymaster allowance this PR set out to remove — and then stays on APPROVE mode for the rest of the TTL.

Narrow the catch to viem's `ContractFunctionExecutionError` / `ContractFunctionRevertedError` and let transport errors propagate (or at least go uncached).

sdk/packages/simplex/src/services/paymaster/provider/simplex.ts
```diff
@@ -173,34 +257,110 @@ async function buildPermitMode(
...
+const permit2Support = new Map<string, { supported: boolean; checkedAt: number }>()
+
+/**
+ * Only paymaster implementations that expose PERMIT2() accept mode 0x02; older
+ * deployments reject it with InvalidMode. Probing keeps the client safe while the
+ * redeploy lands chain by chain. Positive results are cached for the process
+ * lifetime, negative ones briefly so an upgrade is picked up without a restart.
+ */
+async function paymasterSupportsPermit2(client: PublicClient, paymasterAddress: HexString): Promise<boolean> {
+	const key = paymasterAddress.toLowerCase()
```
seunlanlege: **The support cache carries no chain identity.**

`permit2Support` is process-wide and keyed on the paymaster address alone, with positive results cached for the process lifetime, while one filler process serves all five chains.

Latent today — the five deployments in `chain.ts` have distinct addresses. But this PR *is* a multi-chain redeploy, and if it uses CREATE2 for a common address (the usual pattern), the first chain upgraded marks that address supported forever; every op built for a not-yet-upgraded chain then emits mode `0x02` and fails validation with `InvalidMode(2)` — which, per `Decisions.md`, means a lost fill for a bid.

`` `${chainId}:${address}` `` as the key removes the coupling to the deployment strategy.

evm/tests/foundry/SimplexPaymasterGasGriefTest.t.sol
```diff
@@ -0,0 +1,228 @@
...
+        ENTRY_POINT.handleOps(tooHigh, payable(beneficiary));
+
+        PackedUserOperation[] memory tooLow = _ops(60_000, 20_000);
+        vm.expectRevert();
+        ENTRY_POINT.handleOps(tooLow, payable(beneficiary));
+    }
+
+    /// Probe the minimum viable postOp gas limit per token: postOp must still fit, and the
+    /// EntryPoint penalty must be waived (limit <= gasUsed + 40k).
+    function testPostOpFitsWithinFiftyThousandForBothTokens() public onFork {
```
seunlanlege: **This test has no assertions.**

It runs four sponsored ops and only `emit log_named_uint`. The 50,000 in the name is never exercised either — the loop uses 40,000 and 30,000.

A change that makes postOp more expensive for one token, or that reintroduces the EntryPoint penalty inside the band, leaves the margin shrunken but positive and this stays green, reporting the regression only into CI logs.

`assertGe(charged, spent)` per iteration, plus an equality check against the cap margin if the intent is to prove the band is flat.

evm/tests/foundry/SimplexPaymasterGasGriefTest.t.sol
```diff
@@ -0,0 +1,228 @@
...
+
+        emit log_named_uint("margin @ cap   40k (wei)", marginAtCap);
+        emit log_named_uint("margin @ floor 30k (wei)", marginAtFloor);
+        // Identical across the whole accepted band: the EntryPoint waives the penalty
+        // while gasLimit <= gasUsed + 40k, so the cap is penalty-free whatever postOp costs.
+        assertEq(marginAtCap, marginAtFloor, "penalty charged inside the allowed band");
+    }
+
+    /// The band the contract enforces is exactly [MIN, MAX]; outside it validation refuses.
+    function testPostOpLimitsOutsideTheBandAreRefused() public onFork {
```
seunlanlege, line 143: **Bare `vm.expectRevert()` accepts any revert, including ones that never reach the check under test.**

If the op stops reaching the postOp-limit branch at all — stale nonce from a preceding test, exhausted paymaster deposit, an oracle that went stale on the pinned fork block — `handleOps` still reverts (AA25/AA31/`StaleOraclePrice`) and this passes while proving nothing about `MIN`/`MAX_POST_OP_GAS_LIMIT`.

Expect the `FailedOpWithRevert` payload carrying `InvalidPostOpGasLimit`. The three `vm.expectRevert()` calls in `Permit2CompromiseForkTest` have the same gap and should name `InvalidSigner` / `InvalidAmount` / `InvalidNonce` — `SimplexPaymasterPermit2ForkTest` already does this well.

sdk/packages/simplex/src/services/paymaster/types.ts
```diff
@@ -30,12 +30,14 @@ export interface PaymasterOptions {
...
-	 * invalidate them.
+	 * Skips EIP-2612 permit detection for the Simplex paymaster (PERMIT2 and APPROVE
+	 * modes stay available). Delegation UserOps rely on fixed, measured gas limits;
+	 * executing a permit during paymaster validation adds tens of thousands of
+	 * verification gas and would invalidate them.
 	 */
-	forceApproveMode?: boolean
+	skipPermit?: boolean
+	/** Lifetime of a Permit2 signature; defaults to {@link PERMIT2_DEADLINE_SECONDS}. */
+	permit2DeadlineSeconds?: bigint
```
seunlanlege: **`permit2DeadlineSeconds` is unreachable from any production caller.**

It's threaded through `buildPaymasterAndData` → `buildSimplexPaymasterData`, but `SponsoredUserOpRequest` in `UserOpSender` has no such field, and `ContractInteractionService.prepareBidUserOp` passes no options at all. Only tests can set it.

`Decisions.md` documents the deadline as "1 hour, overridable per call", which no runtime path supports. Either thread it through `SponsoredUserOpRequest` (and the bid path, where a shorter window for losing bids is the case that actually motivates it) or drop the field and the doc claim.

sdk/packages/simplex/src/services/paymaster/provider/simplex.ts
```diff
@@ -173,34 +257,110 @@ async function buildPermitMode(
...
+		paymasterVerificationGasLimit: VERIFICATION_GAS_LIMIT_PERMIT2,
+		paymasterPostOpGasLimit: POST_OP_GAS_LIMIT_SIMPLEX,
 	}
+}
 
-	// The approve is a plain EOA tx paid in native — the very thing the paymaster
-	// exists to avoid needing. Pre-check the balance so an unfunded solver fails
-	// with one actionable line instead of viem's estimateGas error chain.
+/**
+ * Sends `approve(spender, amount)` from the solver EOA and waits for one
```
seunlanlege: Stale after the rename: the docstring says "waits for one confirmation" while the body waits for two. The inline comment three lines below and `Decisions.md` both say two; `docs/ai/Flow.md` repeats the stale "wait one confirmation".

Worth fixing precisely because the confirmation count *is* the fix for the AA33-after-bootstrap failure — a reader trusting either doc would take it back to one while trimming the latency it adds.

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -390,6 +470,83 @@ contract SimplexPaymaster is Initializable, HyperApp, PaymasterERC20 {
...
+        prefundAmount = _erc20Cost(maxCost, userOp.maxFeePerGas(), tokenPrice);
+        if (prefundAmount > permitAmount) revert InsufficientPermitAmount(permitAmount, prefundAmount);
+        // Solidity's own extcodesize revert on a code-less target is not caught by try/catch.
+        if (address(PERMIT2).code.length == 0) revert Permit2NotDeployed();
+
+        // Charge the token _fetchDetails already validated as registered and active, rather
+        // than re-trusting the mode byte's token field, so the external call never depends
+        // on the caller-ordering of the base contract.
+        address tokenAddr = address(token);
+        address owner = userOp.sender;
```
seunlanlege: `prefunder_` is received and never read here; the mode-0/1 branch above forwards it to `super._prefund`, so the two branches disagree about where the prefunder comes from.

They coincide today (`PaymasterERC20` passes `userOp.sender`), so nothing breaks. But the base declares `prefunder_` as the hook for who funds the op, and the context this branch writes records its own `owner` as the `_postOp` refund recipient — if the base or a subclass ever routes the prefund through a different funder, mode `0x02` diverges silently with no compile-time signal.

`address owner = prefunder_;` keeps the branches aligned.

evm/src/utils/SimplexPaymaster.sol
royvardhan: Resolved with a third option: keep the contract ceiling at 100k and lower only the client. The margin gain comes entirely from the client sending 40k — the contract just has to accept it — so `MAX_POST_OP_GAS_LIMIT` stays 100k (plus a new 30k floor for the refund-underflow window) and the SDK sends 40k. An in-place upgrade of a live proxy is then safe for any client still on 100k, and the full margin is still realized. We give up only the marginal anti-grief tightening (a griefer inflating postOp to 100k forfeits ~8.85k gas from the cushion the op already pays), revisitable once every integrator is on a 40k client. `Decisions.md`/`CHANGELOG.md` updated.

sdk/packages/simplex/src/services/paymaster/types.ts
royvardhan: Dropped the field and the "overridable per call" doc claim. No production caller could set it, and the fixed 1h deadline covers every path today (bid-to-execution runs well under an hour). If per-bid tuning is wanted later — a shorter window for losing bids — we can thread it through `SponsoredUserOpRequest` and the bid path then.

seunlanlege (approved):

PR #1149, opened 2026-08-19
[simplex]: quorum client suspends rate-limited endpoints for 5 minutes

seunlanlege:
Re-querying a 429ing endpoint on every call wastes the request and deepens the throttle. The quorum client now **drops** an endpoint for 5 minutes when its failure is unambiguously a request-rate limit — out of the query set and out of the quorum bar both.

## Semantics

Each call's threshold is `quorumThreshold(endpoints actually queried)`. A throttled endpoint answers nothing either way, so counting it could only fail calls the remaining endpoints agree on — and a scanner that cannot form a quorum misses orders. Concretely: 2 endpoints with one 429ing, the healthy endpoint alone carries reads for the window instead of five minutes of guaranteed failures; 5 endpoints with one benched, the bar is 3-of-4. With every endpoint benched, all are queried again. The public `threshold` field still reports the full-set bar; every `QuorumError` carries the per-call bar.

(First shipped with a fixed full-set threshold plus a query-them-anyway fallback; reversed on maintainer review — the availability loss under ordinary provider throttling outweighed the marginal hardening, and the voters are always exclusively the operator's own endpoints. Both designs are recorded in `docs/ai/Decisions.md`.)

## The classifier split

Benching gets its own classifier, stricter than the diagnostic label. `isRateLimited` stays loose for log tags; `isSuspendableRateLimit` (HTTP 429, throttle-specific codes −32016/−32097, or throttle text, reading no URL-bearing metaMessages) is what benches. The distinction matters: `-32005` is EIP-1474's generic *limit exceeded*, and Infura returns it for **deterministic** `eth_getLogs` result caps — which catch-up scans over busy 1000-block ranges hit repeatedly. Benching on it would suspend a healthy endpoint indefinitely.

## Diagnostics

`suspended()` exposes benched URLs; `QuorumError` messages read `responders: N/M queried (K/S suspended for rate limiting)`, with the skipped count snapshotted at endpoint selection (a long call can outlive a suspension window). Failure detail in `getTransactionConfirmations` now names real endpoint URLs (previously `unknown`).

## Verification

Tested against **real HTTP 429/500/−32005 responses** through viem's actual error shapes (local servers, CI-safe). Notable: the traffic-stops assertion goes through uncached `getLogs` — viem caches `eth_blockNumber` for 4s, and adversarial review **proved the naive assertion passed with suspension disabled entirely**; the hardened suite fails 2 tests under that exact regression (verified by probe). Also covered: the degraded-bar behavior at n=2 and n=5, re-query + re-bench after expiry, −32005 not benching, 500s staying stateless.

Full ungated suite passing over the baseline network-gated set.


Wizdave97:
### 1. The `metaMessages` guard in `isSuspendableRateLimit` does not work — verified

[[QuorumPublicClient.ts:78](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/QuorumPublicClient.ts:78)](sdk/packages/simplex/src/services/QuorumPublicClient.ts:78) claims the free-text match "reads `message`/`details`/`shortMessage` only, not `metaMessages` (which embed the request URL)". viem's `BaseError` constructor **folds `metaMessages` into `message`**:

```
HTTP request failed.

Status: 429
URL: https://rpc.example/ratelimit-key-429-abc
Request body: {"method":"eth_blockNumber"}

Details: Too Many Requests
Version: viem@2.47.6
```

So the URL is inside `message`, and reading `message` reads `metaMessages`. I probed this end-to-end through the real class: four endpoints, one serving a plain **HTTP 500** at a URL path containing `ratelimit` — `client.suspended()` returns that endpoint. A healthy endpoint gets benched for five minutes on a non-throttle failure, which is exactly the failure mode the comment says is prevented.

Narrow trigger in practice (needs "rate limit"/"ratelimit"/"rate-limit" in the URL or key), but the documented guard is nominal, and the same claim is repeated in [[Decisions.md](https://claude.ai/epitaxy/sdk/packages/simplex/docs/ai/Decisions.md)](sdk/packages/simplex/docs/ai/Decisions.md) and [[ChangeLog.md](https://claude.ai/epitaxy/sdk/packages/simplex/docs/ai/ChangeLog.md)](sdk/packages/simplex/docs/ai/ChangeLog.md).

Fix I tested — read `message` only when there are no `metaMessages` to have leaked into it:

```ts
const text = [e.details, e.shortMessage, e.metaMessages ? undefined : e.message].filter(Boolean).join(" ")
```

Real viem errors put the provider's own text in `details`/`shortMessage`, so 429 and throttle-text detection are unaffected; the `metaMessages`-free synthetic errors the unit test builds still match on `message`. Full suite passes with this (42/48). Dropping `message` outright also fixes the leak but breaks the `rpcError(-32005, "too many requests")` assertion at [[rate-limit-detection.test.ts:295](https://claude.ai/epitaxy/sdk/packages/simplex/src/tests/rate-limit-detection.test.ts:295)](sdk/packages/simplex/src/tests/rate-limit-detection.test.ts:295).

### 2. The test guarding that behaviour uses an error shape viem never produces

[[rate-limit-detection.test.ts:303](https://claude.ai/epitaxy/sdk/packages/simplex/src/tests/rate-limit-detection.test.ts:303)](sdk/packages/simplex/src/tests/rate-limit-detection.test.ts:303) builds `Object.assign(new Error("HTTP request failed."), { metaMessages: [...] })` — `message` and `metaMessages` disjoint, which viem never emits. That's precisely the circularity the file's own header docstring calls out ("passes even if viem's actual error shape for an HTTP 429 changed"). It should go through `rpcServer` with a URL path carrying the pattern, like the rest of the suite does.

### 3. The degraded bar also removes the pre-fill confirmation gate, which Decisions.md doesn't cover

The reversal is argued entirely in terms of the scanner missing orders. But the same client is the confirmation gate on the money path: [[filler.ts:746](https://claude.ai/epitaxy/sdk/packages/simplex/src/core/filler.ts:746)](sdk/packages/simplex/src/core/filler.ts:746) polls `getTransactionConfirmations` on the **source** chain before filling. At n=2 — which [[quorum-public-client.test.ts:17](https://claude.ai/epitaxy/sdk/packages/simplex/src/tests/quorum-public-client.test.ts:17)](sdk/packages/simplex/src/tests/quorum-public-client.test.ts:17) calls "the one operators most commonly run" — one benched endpoint makes the bar 1, so a single endpoint's receipt decides that a source order is confirmed to the required depth. `aggregateConfirmations`' whole purpose ("a minority still serving a reorged/fabricated receipt can never reach the quorum") is off for the window.

Also worth correcting in the rationale: on a `getLogs` failure the scanner does **not** advance its cursor ([[chain-scanner.ts:213](https://claude.ai/epitaxy/sdk/packages/simplex/src/scanner/chain-scanner.ts:213)](sdk/packages/simplex/src/scanner/chain-scanner.ts:213), [[:244](https://claude.ai/epitaxy/sdk/packages/simplex/src/scanner/chain-scanner.ts:244)](sdk/packages/simplex/src/scanner/chain-scanner.ts:244)), so the pre-PR behaviour was a stall-then-catch-up — lost *timeliness*, not lost orders. The degraded path is the one that can advance the cursor on a 1-of-1 answer, and per that file's own comment a skipped range is skipped for every consumer, permanently. The trade is real either way, but it's latency-vs-silent-gap, not liveness-vs-nothing.

If you want to keep the degradation for reads and not for the gate, a floor of 2 voters when `size >= 2` in `getTransactionConfirmations` is a small, targeted change.

### 4. Suspension is completely invisible in the happy path

`suspended()` is called by nothing outside tests, and the class has no logger. When the bar silently drops from 2-of-2 to 1-of-1 for five minutes, nothing is logged unless a `QuorumError` happens to be thrown. Given #3, a `warn` on bench (and on a degraded bar) is worth more than the diagnostic string work in this PR.

### 5. Bench state is per-instance, and the two hot paths use different instances

`ChainScanner` constructs its own client at [[chain-scanner.ts:80](https://claude.ai/epitaxy/sdk/packages/simplex/src/scanner/chain-scanner.ts:80)](sdk/packages/simplex/src/scanner/chain-scanner.ts:80); the filler goes through `ChainClientManager.getQuorumClient` ([[ChainClientManager.ts:172](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/ChainClientManager.ts:172)](sdk/packages/simplex/src/services/ChainClientManager.ts:172)), whose doc comment already claims — incorrectly — that "the event monitor and the confirmation waiter share one provider set". So a throttled endpoint benched by the scanner keeps getting hammered by the confirmation poller (and vice versa), which halves the intended relief. `ChainScanner.setRpcUrls` also drops all bench state by rebuilding the client.

### 6. Minor

- **All-benched fallback**: `participants()` reports `skipped: 0` when every endpoint is suspended, so the `QuorumError` says nothing about it — and the fallback re-queries all of them at full rate, abandoning the "don't deepen the throttle" goal exactly when the throttle is total. Not a regression, but the message should at least say so.
- **`Retry-After` ignored**: viem's `HttpRequestError` carries `headers`. A per-second limiter answering `Retry-After: 1` still costs five minutes of degraded quorum.
- **Indentation**: the task body at [[QuorumPublicClient.ts:348-357](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/QuorumPublicClient.ts:348)](sdk/packages/simplex/src/services/QuorumPublicClient.ts:348) is one level short after the `{ idx, task }` rewrap. Cosmetic — there's no format check in CI.

---

Blocking, in my view: #1 (with the doc lines in `Decisions.md`/`ChangeLog.md`) and #2. #3 is a maintainer call that's already been made once, but the confirmation-gate half of it isn't in the recorded rationale and should be.

seunlanlege:
All addressed in b5d9a8ba8 — and thank you for the end-to-end probe on #1, it was exactly right.

**1. Confirmed and fixed.** Reproduced your probe first: viem folds `metaMessages` into `message`, so the guard was nominal exactly as you said. Took your fix verbatim — `message` is consulted only when the error carries no `metaMessages` — and verified both directions: your 500-at-`/team-ratelimit-key` probe is now a permanent test, and reverting the one line fails it. The false claim is corrected in the docblock, `Decisions.md` and `ChangeLog.md`, with the two-pass history recorded rather than papered over.

**2. Fixed.** The impossible-shape synthetic is gone; the real-server probe replaces it (the remaining unit case now uses the folded-message shape viem actually builds, and the e2e test is the authority).

**3. Maintainer ruled: degradation stands everywhere, the fill gate included.** I put your floor-of-2 suggestion and the full exemption to them as explicit options; they chose uniform degradation. Recorded in `Decisions.md` with the accepted consequence spelled out (at n=2 with one endpoint benched, a single endpoint's receipt vouches inclusion for the window). Your cursor-stall correction is also folded into the recorded rationale — you're right that pre-drop was stall-then-catch-up, and that the degraded path is the one that can advance the cursor past a 1-of-1 answer; the rationale now says latency-versus-silent-gap.

**4. Done.** The class takes an optional `LoggerContext` (wired from both `ChainClientManager` and `ChainScanner`); benching warns with the endpoint, duration, and the new effective bar.

**5. Done.** Bench state is a static map keyed by endpoint URL — shared across the scanner's and the poller's instances, and it survives `setRpcUrls` rebuilds. `ChainClientManager`'s incorrect sharing claim is corrected. (`clearAllSuspensions()` exists for tests.)

**6. All three.** The all-benched fallback names itself in `QuorumError` messages; `Retry-After` is honoured with a [1s, 5min] clamp (your per-second-limiter example now costs seconds, with a test); the indentation was prettier-ed.

Wizdave97 (approved):

PR #1168, opened 2026-08-25
[sdk]: recover intent execution from rpc failures

ddboy19912:
Add bounded RPC timeouts and recovery handling to SDK intent execution. Timed-out UserOperation submissions are reconciled before another bid is selected, and retry/confirmation status events improve observability


Wizdave97:
This is not necessary for now, we might come back to it.

PR #1170, opened 2026-08-26
[indexer]: run substrate chains on the Polytope subql node build

royvardhan:
Substrate chains run on `polytopelabs/subql-node-substrate:v5.10.1-0` (from the `polytope-labs/subql` fork) instead of the upstream image, which exits on an RPC disconnect and caches failed requests. Image reference and docs only.

Depends on the fork image being published first (polytope-labs/subql#1).

### DIFF
--- a/sdk/packages/indexer/docker/docker-compose.local.yml
+++ b/sdk/packages/indexer/docker/docker-compose.local.yml
@@ -1,6 +1,6 @@
 services:
     subquery-node-hyperbridge-gargantua-local:
-        image: subquerynetwork/subql-node-substrate:v5.9.1
+        image: polytopelabs/subql-node-substrate:v6.4.7-0
         restart: always
         network_mode: host
         environment:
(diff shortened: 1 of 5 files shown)


seunlanlege:
This isn't actually neccessary because the subql node supports http

Wizdave97:
> This isn't actually neccessary because the subql node supports http

Yeah, but we need to remove the caching which requires publishing a new image

Wizdave97:
@royvardhan  is the substrate node not a seperate javascript package that needs to be linked to the changes made in the node-core before we can publish?

royvardhan:
Good question. `@subql/node` and `@subql/node-core` live in the same monorepo, so node-core doesn't need to be published separately. The build packs the patched node-core into a tarball and points node at it (`file:./node-core.tgz`), so the image build picks up both sets of changes. Verified on the built image: node-core is 17.1.1-0 with the reconnect changes in it. So publishing `polytopelabs/subql-node-substrate` from polytope-labs/subql#1 is enough.

Wizdave97:
@royvardhan  Please resolve conflicts

Wizdave97 (approved):

PR #1062, opened 2026-07-21
polkadot 2606.0.0 upgrade

dharjeezy:
This PR upgrades the polkadot sdk version to 2604.0.0

PR #1174, opened 2026-08-27
[crates]: republish ismp-abi and beefy-verifier-primitives, and keep polkadot-sdk out of ismp

### LINKED ISSUE #1172: [crates]: republish ismp-abi and beefy-verifier-primitives, published 0.1.1 still pins polkadot-sdk =2512.3.2
## Summary

`ismp-abi 0.1.1` and `beefy-verifier-primitives 0.1.1` are the only published versions of those crates, and both declare `polkadot-sdk = "=2512.3.2"`. The workspace moved to `=2606.0.0` in #1062, but neither crate was republished, so the artifacts on crates.io still carry the pre-upgrade pin.

Because `ismp 2606.0.0` depends on `ismp-abi ^0.1.1`, every consumer of the 2606 line resolves **two `polkadot-sdk` umbrella versions into one dependency graph**.

The manifests in this repo are already correct, this looks like a missing republish rather than a code change.

## Observed

```toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]
pallet-ismp = "2606.0.0"
ismp-grandpa = "2606.0.0"
```

```console
$ cargo generate-lockfile
$ grep -A1 'name = "polkadot-sdk"' Cargo.lock | grep version
version = "2512.3.2"
version = "2606.0.0"
```

## Root cause

Both crates inherit the umbrella from the workspace, and the workspace is on 2606:

```toml
# evm/rust/Cargo.toml (ismp-abi)
# modules/consensus/beefy/primitives/Cargo.toml (beefy-verifier-primitives)
[dependencies.polkadot-sdk]
workspace = true
```

```toml
# Cargo.toml:163
polkadot-sdk = { version = "=2606.0.0", default-features = false }
```

That line landed in `eb654713` ("polkadot 2606.0.0 upgrade", #1062) on 2026-07-23. Both crates were published on 2026-05-30, ~8 weeks earlier, and have not been republished since, so the manifests advanced while the released version numbers did not.

| Crate | Published | `polkadot-sdk` on crates.io | This repo |
|---|---|---|---|
| `ismp-abi` | 0.1.1 (only) | `=2512.3.2` | `=2606.0.0` |
| `beefy-verifier-primitives` | 0.1.1 (only) | `=2512.3.2` | `=2606.0.0` |

Resolution chain:

```
ismp 2606.0.0
  └── ismp-abi ^0.1.1
        └── polkadot-sdk =2512.3.2

grandpa-verifier 2606.0.0        → polkadot-sdk =2606.0.0
substrate-state-machine 2606.0.0 → polkadot-sdk =2606.0.0
```

Both requirements are exact, so no resolver can reconcile them onto one version.

## Impact

**Consuming everything from crates.io:** resolves and builds, but carries a duplicated SDK tree.

**Mixing a git-sourced SDK with the registry ISMP crates**, relevant to the [solochain guide][docs], which has solochains build against `polkadot-sdk` directly splits every SDK crate across two sources:

```
frame-support    ['git', 'reg']
sp-io            ['git', 'reg']
sp-runtime       ['git', 'reg']
```

A small crate tolerates that. A FRAME runtime does not: building one this way fails on duplicated types, e.g.

```
error[E0152]: duplicate lang item in crate `sp_io`: `panic_impl`
error[E0308]: expected `sp_application_crypto::KeyTypeId`,
                 found `sp_core::crypto::KeyTypeId`
```

`[patch.crates-io]` does not resolve it: a patch has to satisfy the original requirement, and `=2512.3.2` admits exactly one version that no published tag provides (`polkadot-stable2512-2` and `-3` both ship `2512.3.3`; the `stable2512` branch head is `2512.4.0`).

## Suggested fix

1. Publish `ismp-abi 0.2.0` and `beefy-verifier-primitives 0.2.0` from the current workspace, so the released artifacts carry `=2606.0.0`.
2. Consider relaxing `=` to `^` on the umbrella across the crates that pin it, so the mismatch does not recur when one crate is published out of step.

## Notes

We hit this while integrating a sovereign GRANDPA solochain. We worked around it by staying on the 2512 line, where the pins are internally consistent, so it is not blocking us. Filing because it looked like it would bite anyone adopting 2606.

Happy to test a republish.

[docs]: https://docs.hyperbridge.network/developers/polkadot/solochains

seunlanlege:
Fixes #1172.

Two commits. The first is the republish the issue asks for. The second removes the reason it spread so far.

## The bug

`ismp-abi` and `beefy-verifier-primitives` are each published at exactly one version, `0.1.1`, and both pin `polkadot-sdk = "=2512.3.2"`. The workspace moved to `=2606.0.0` in #1062, but neither crate was republished, so the artifacts on crates.io still carry the pre-upgrade pin.

`ismp 2606.0.0` depends on `ismp-abi ^0.1.1`, so every consumer of the 2606 line pulls two SDK umbrellas into one graph:

```console
$ cargo add pallet-ismp@2606.0.0 ismp-grandpa@2606.0.0 && cargo generate-lockfile
$ grep -A1 'name = "polkadot-sdk"' Cargo.lock | grep version
version = "2512.3.2"
version = "2606.0.0"
```

Both requirements are exact, so nothing reconciles them. It also doesn't stop at the facade crate — the components double up too, which is what produces the `duplicate lang item in sp_io` error the reporter hit:

```
sp-core       39.0.0  + 43.0.0
sp-runtime    45.0.0  + 48.0.0
sp-io         44.0.0  + 48.0.0
frame-support 45.1.3  + 48.0.0
```

## Commit 1 — put both crates back on the release line

`scripts/release-crates.sh` drives `cargo release` over 13 crates, and neither of these is in it. Walking the workspace dependency closure from the released set, they're the only two crates reachable from a released crate that aren't released themselves.

The version scheme is what hid it. Every released crate that pins the umbrella is on CalVer; the two that don't pin it (`serde-hex-utils`, `crypto-utils`) are on semver. These two pin `=2606.0.0` but sit on `0.1.1` — the only exceptions — so their version numbers gave no signal that the pin had moved underneath them.

- `beefy-verifier-primitives` `0.1.1` → `2606.0.0`
- `ismp-abi` `0.1.1` → `2606.0.0`
- `ismp` `2606.0.0` → `2606.1.0`
- both added to `scripts/release-crates.sh`

Two things worth a look here.

**`ismp` has to be bumped as well.** Publishing `ismp-abi` on its own fixes nothing, because the already-published `ismp 2606.0.0` is immutable and requires `ismp-abi ^0.1.1`. Bumping `ismp` is what carries the fix to consumers. Nothing downstream needs republishing — `pallet-ismp`, `ismp-grandpa` and the rest require `ismp ^2606.0.0`, so they pick up `2606.1.0` on their own.

**Deliberately not a patch bump.** `ismp-abi 0.1.2` would satisfy the existing `^0.1.1` and fix the 2606 line without touching `ismp`, which is exactly why it's wrong: the 2512 line resolves to a single `polkadot-sdk 2512.3.2` today, and a patch bump would drag the 2606 pin onto it. That moves the breakage rather than fixing it. The CalVer bump is chosen so `^0.1.1` does *not* match.

## Commit 2 — keep the umbrella out of `ismp`

`ismp` sits at the root of the published tree, so an SDK pin on it reaches everything. That's why a stale pin on one small ABI crate turned into a broken FRAME runtime.

It doesn't need the umbrella. It uses `sp-weights` directly, which is fine — `Weight` is in its public API and the crate is small and standalone. The heavy dependencies all arrive through `ismp-abi`, which pulls `frame-support`, `sp-runtime`, `sp-consensus-beefy` and `sp-mmr-primitives` for two modules `ismp` never touches: the SCALE `HostParam` mirrors in `host_params`, and the BEEFY conversions in `conversions`. What `ismp` actually imports is the generated `sol!` bindings and `ToU256`, all plain alloy.

So both modules now sit behind a `substrate` feature, with `polkadot-sdk`, `beefy-verifier-primitives`, `codec` and `scale-info` optional on it. The feature is on by default, so external consumers of `ismp-abi` see no change. `ismp` turns it off.

Four workspace crates use the gated surface while inheriting `default-features = false`, so they opt in explicitly — `pallet-ismp-host-executive`, `pallet-beefy-consensus-proofs`, `pallet-ismp-testsuite` and `tesseract-beefy`. A workspace build would have covered them through feature unification, but `cargo check -p <one>` wouldn't, so the opt-in is written out.

Result: `ismp`'s dependency tree goes from 420 crates to 270, with no `polkadot-sdk`, `frame-support`, `sp-io`, `sp-core` or `sp-runtime` left in it. The 2606 pin now lives only on crates that are genuinely SDK-bound, where consumers already expect it.

## Publishing

Order matters — `beefy-verifier-primitives`, then `ismp-abi`, then `ismp`. `cargo release` takes that from the script ordering. Until they're published, `cargo package` on `ismp-abi`/`ismp` can't resolve the new requirement from the registry, which is expected. `beefy-verifier-primitives` still has to be published even though `ismp` no longer uses it, since cargo requires optional dependencies to resolve at publish time.

## Verification

- `cargo check` passes for: `ismp-abi` with the feature off (both `no_std` and `std`), `ismp-abi` with it on, `ismp` with it off, and the two pallets that opt in
- `cargo tree -p ismp` shows no polkadot-sdk, frame-support, sp-io, sp-core or sp-runtime
- workspace lockfile updates cleanly, still a single `polkadot-sdk 2606.0.0`
- reproduced the dual-SDK resolution on the 2606 line and confirmed the 2512 line is clean, both against the live index

Full workspace build is left to CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2947,7 +2947,7 @@ dependencies = [
 
 [[package]]
 name = "beefy-verifier-primitives"
-version = "0.1.1"
+version = "2606.0.0"
 dependencies = [
  "derive_more 1.0.0",
  "parity-scale-codec",
@@ -10495,7 +10495,7 @@ checksum = "a6cb138bb79a146c1bd460005623e142ef0181e3d0219cb493e02f7d08a35695"
 
 [[package]]
 name = "ismp"
-version = "2606.0.0"
+version = "2606.1.0"
 dependencies = [
  "alloy-primitives 1.5.7",
  "alloy-sol-types 1.5.7",
@@ -10515,7 +10515,7 @@ dependencies = [
 
 [[package]]
 name = "ismp-abi"
-version = "0.1.1"
+version = "2606.0.0"
 dependencies = [
  "alloy-contract",
  "alloy-network",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -248,7 +248,7 @@ ibc-core-host = { version = "0.57.0", default-features = false }
 ibc-proto = { version = "0.52.0", default-features = false, package = "ibc-proto"}
 
 # published crates
-ismp = { version = "2606.0.0", path = "./modules/ismp/core", default-features = false }
+ismp = { version = "2606.1.0", path = "./modules/ismp/core", default-features = false }
 serde-hex-utils = { version = "0.2.0", path = "modules/utils/serde", default-features = false }
 crypto-utils = { version = "0.3.0", path = "modules/utils/crypto", default-features = false }
 bls-utils = { path = "modules/utils/bls-utils", default-features = false }
@@ -268,14 +268,14 @@ substrate-state-machine = { version = "2606.0.0", path = "modules/ismp/state-mac
 
 # local crates
 ismp-testsuite = { path = "./modules/ismp/testsuite" }
-ismp-abi = { version = "0.1.1", path = "./evm/rust", default-features = false }
+ismp-abi = { version = "2606.0.0", path = "./evm/rust", default-features = false }
 simnode-tests = { path = "parachain/simtests" }
 integration-tests = { path = "parachain/integration-tests" }
 hyperclient = { path = "modules/hyperclient", default-features = false }
 subxt-utils = { path = "modules/utils/subxt", default-features = false }
 
 # consensus provers & verifiers
-beefy-verifier-primitives = { version = "0.1.1", path = "./modules/consensus/beefy/primitives", default-features = false }
+beefy-verifier-primitives = { version = "2606.0.0", path = "./modules/consensus/beefy/primitives", default-features = false }
 beefy-prover = { path = "./modules/consensus/beefy/prover" }
 beefy-verifier = { path = "./modules/consensus/beefy/verifier", default-features = false }
 bsc-prover = { path = "./modules/consensus/bsc/prover" }
--- a/evm/rust/Cargo.toml
+++ b/evm/rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ismp-abi"
-version = "0.1.1"
+version = "2606.0.0"
 edition = "2021"
 authors = ["Polytope Labs <hello@polytope.technology>"]
 description = "Generated rust types for the ISMP solidity ABI"
--- a/evm/rust/Cargo.toml
+++ b/evm/rust/Cargo.toml
@@ -22,20 +22,33 @@ alloy-provider = { workspace = true, optional = true }
 alloy-network = { workspace = true, optional = true }
 alloy-transport = { workspace = true, optional = true }
 
-beefy-verifier-primitives = { workspace = true, default-features = false }
+beefy-verifier-primitives = { workspace = true, default-features = false, optional = true }
 
 # SCALE codec for the host-params SCALE representation that lives alongside the
 # ABI types so the cross-language conversion impls have a single home.
-codec = { workspace = true }
-scale-info = { workspace = true }
+codec = { workspace = true, optional = true }
+scale-info = { workspace = true, optional = true }
 
 [dependencies.polkadot-sdk]
 workspace = true
 default-features = false
+optional = true
 features = ["sp-consensus-beefy", "sp-mmr-primitives", "frame-support", "sp-runtime"]
 
 [features]
-default = ["std"]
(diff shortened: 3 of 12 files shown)


Wizdave97 (approved):

royvardhan (approved):

nol4lej:
Thank you for implementing it. Best regards!

PR #1175, opened 2026-08-27
[simplex]: Pace Hyperbridge HTTP reads and halve the phantom poll's per-block cost

Wizdave97:
Fillers were logging `[429]: Too Many Requests` from the Hyperbridge HTTP endpoint against a 10 req/s limit, on a poll whose interval is 15s. The interval was never the request rate.

A tick costs one head read plus the cost of every block in its range, and `getPhantomOrdersInBlock` was costing four RPCs per block, not one: `api.at(hash)` with no version to go on resolves a registry by fetching `chain_getHeader(hash)` and then `state_getRuntimeVersion(parentHash)` on every block, because the `getUpgradeVersion` shortcut that would skip it only covers chains hardcoded in `@polkadot/types-known`. They all go out back-to-back, so a tick averaging 0.4 req/s arrives at ~33 req/s. On a phantom order interval, simplex's `handlePhantomOrders` fans out one concurrent `offchain_localStorageGet` per configured chain (up to 16) on top of that.

- `http()` now builds a `RateLimitedHttpProvider`, whose `send` waits on a `TokenBucket`. Buckets are keyed by endpoint origin in a module-level map, so every coprocessor in a process pointed at the same node shares one budget — the limit counts requests per address, not per connection. Default 8 req/s, `HYPERBRIDGE_RPC_MAX_RPS` to override.

- `getPhantomOrdersInBlock` takes an optional `knownVersion` and passes it to `api.at`, which drops both hidden reads. `pollPhantomOrders` establishes one per tick via `confirmedRuntimeVersion()`: read after the head, used only when it matches the previous reading. `specVersion` only increases, so two equal readings prove no upgrade landed in the range about to be scanned; a reading that differs falls back to exact per-block resolution for that tick. A stale pin fails silently — events decode into a shape the scan does not recognise, the block reads as empty, and the cursor advances past it — which is the miss the block cursor exists to prevent. The read is skipped when there is nothing to scan, so a quiet tick still costs exactly one request.

- `maxBlocksPerPoll` defaults to 20 rather than 500, and a 429 backs the poll off for a doubling number of ticks (capped at 8), reset by any tick that gets through. Ordinary failures still retry on the very next tick. The cap is not optional once the bucket exists: a 500-block catch-up would queue ~1000 requests ahead of a bid submission that is worth nothing after five blocks.

Net per tick: `1 + 4n` requests becomes `1 + 1 + 2n`, with the burst paced.


seunlanlege:
Can probably leverage batched rpc method calls here

https://www.json-rpc.dev/learn/examples/batch-requests

seunlanlege:
merge conflicts

seunlanlege (approved):

PR #1176, opened 2026-08-27
[evm, sdk, simplex]: Give solver bids a signed expiry

Wizdave97:
A signed bid was valid forever. SolverAccount.validateUserOp returned SIG_VALIDATION_SUCCESS, which packs validUntil = 0, and ERC4337Utils expands 0 to BLOCK_RANGE_MASK. Nothing else bounded it: order.deadline is chosen by the order placer with no contract-side ceiling, and retract_bid on Hyperbridge does not reach the destination chain. Since the placer holds the session key, they alone decided when to exercise — a free option on solver inventory, taken up only once the rate had moved against the solver.

The selection signature is now 168 bytes
(commitment | validUntil | solverSig | sessionSig) and the solver signs an EIP-712 BidValidity(bytes32 userOpHash,uint48 validUntil) digest. The expiry must be inside the signed digest because userOpHash does not cover op.signature — an expiry carried alongside it could be rewritten by whoever replays the bid. validUntil == 0 is rejected, since zero is how the old unbounded behaviour is spelled.

The account binds the expiry but cannot judge it: ERC-7562 forbids the TIMESTAMP opcode during validation. Capping the tenor is therefore the signer's job, so simplex stamps every bid via bidValiditySeconds (default 300).

### DIFF
--- a/evm/src/apps/intentsv2/SolverAccount.sol
+++ b/evm/src/apps/intentsv2/SolverAccount.sol
@@ -40,9 +40,35 @@ contract SolverAccount is Account, ERC7821, IERC1271 {
 
     /**
      * @notice Expected signature length for intent solver selection
-     * @dev abi.encodePacked(commitment, solverSignature, sessionSignature) = 32 + 65 + 65 = 162 bytes
+     * @dev abi.encodePacked(commitment, validUntil, solverSignature, sessionSignature)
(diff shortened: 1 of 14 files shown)


seunlanlege:
Prefer we add `validUntil` to `FillerOptions` so we don't re-introduce blind signing of UserOps

PR #1177, opened 2026-08-27
[simplex]: Refuse phantom orders that are not already expired

Wizdave97:
Being expired is the whole reason a phantom order is safe to quote. The pallet builds one with the chain's latest confirmed height as its deadline so it "can never be executed for real", and fillOrder reverts Expired() once deadline < block.number. That invariant is what lets quotePhantomFill skip every ceiling a real bid gets: no budget, no wallet-balance read, neither profit gate.

Nothing upstream established it. The order body is read from a single Hyperbridge node's offchain storage, and fetchPhantomOrder assigns the commitment from the event rather than re-deriving it from the bytes, so the body is unauthenticated. A body with a live deadline turned an unbounded quote into a signed, executable authorization to fill an order of someone else's choosing, paying out to a beneficiary named in that same body.

preparePhantomBid now reads the destination head and refuses anything still fillable, before quoting a leg. A failed head read is also a refusal: without a height there is no way to tell an expired order from a live one, and quoting is the unsafe branch.


seunlanlege:
Prefer we add `validUntil` to `FillerOptions` so we don't re-introduce blind signing of UserOps

seunlanlege (approved):

PR #1180, opened 2026-08-27
[evm, sdk, simplex]: Bound a bid's life with FillOptions.validUntil

Wizdave97:
A signed bid stayed executable forever. The order's deadline is chosen by the placer with no contract-side ceiling, retracting the bid on Hyperbridge does not reach the destination chain, and the placer holds the session key — so they alone decided when to exercise, and would only do so once the rate had moved against the solver. A free option on solver inventory.

FillOptions gains validUntil, a block number that fillOrder enforces with a new FillExpired error. Zero means unbounded, which is the right default for a solver filling directly and keeps existing callers working. It rides in calldata, which userOpHash already covers, so it is tamper-proof without touching the bid signature format or redeploying SolverAccount.

Adding a struct field changes fillOrder's selector (0x5cfb1ea5 -> 0xa5470064), so both shapes are on the wire while gateways upgrade per chain. The new fillOrderCodec probes fillOptionsVersion() once per deployment, encodes the matching shape, and decodes either — so bids built against an older gateway are still priced into aggregation rather than dropped. Only a contract answer is cached; a transport error is rethrown, since caching it would silently strip validUntil from every later fill on that chain.

simplex sets it from bidValiditySeconds (default 1800), converted to blocks per destination chain, and warns once per chain where the gateway is too old to carry it.

Found by the scheduled IntentGateway/Simplex security audit.


evm/src/apps/IntentGatewayV2.sol
```diff
@@ -449,8 +467,9 @@ contract IntentGatewayV2 is IntrinsicIntents, ExtrinsicIntents, ReentrancyGuardT
         }
 
         if (_params.priceOracle != address(0)) {
-            IIntentPriceOracle(_params.priceOracle)
-                .recordSpread(commitment, order.source, order.inputs, options.outputs);
+            IIntentPriceOracle(_params.priceOracle).recordSpread(
+                commitment, order.source, order.inputs, options.outputs
+            );
```
seunlanlege: we can drop this

sdk/packages/simplex/filler-config-example.toml
```diff
@@ -42,6 +42,22 @@ substratePrivateKey = ""
 # Hyperbridge WebSocket URL
 hyperbridgeWsUrl = ""
 
+# Bid validity (optional)
+# How long a signed bid stays executable, in seconds. Default: 1800 (30 minutes).
```
seunlanlege: default should be like 5mins max

sdk/packages/simplex/src/services/ContractInteractionService.ts
```diff
@@ -37,6 +37,16 @@ Decimal.config({ precision: 28, rounding: 4 })
...
  */
+/**
+ * Slack added to every bid's `validUntil`, in blocks.
+ *
+ * The head is read before the bid is built, signed, submitted to Hyperbridge and finally
+ * selected, so it is already behind by the time the fill lands — and `Chain.blockTime` is a
+ * nominal figure, not a guarantee. The asymmetry decides the direction: overshooting costs a
+ * marginally staler quote, undershooting silently throws away bids that were about to be won.
+ */
+const BID_VALIDITY_PAD_BLOCKS = 5n
```
seunlanlege: Should probably pad by seconds based on blocktime, since every chain has different blocktime, 30s is good padding time

Wizdave97: Yeah, that's been done. I'll update the comments

seunlanlege (approved):

PR #1183, opened 2026-08-27
[tesseract]: Use http for substrate state queries

### LINKED ISSUE #626: Tesseract-Substrate: Use http requests for state queries
To improve reliability of tesseract messaging relayer we can  use http requests when making queries and reserve websocket only for transactions submission.

dharjeezy:
Substrate chains now take an optional `rpc_http` endpoint alongside `rpc_ws`. When it is set, plain rpc requests are answered over http and the websocket is left to carry subscriptions alone, which is all that extrinsic submission and the block streams actually need, so a dropped socket costs us pending subscriptions instead of every query in flight. Leaving the field out keeps the current behaviour of running everything over the websocket.

Closes #626.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27902,6 +27902,7 @@ dependencies = [
  "futures",
  "ismp",
  "ismp-parachain",
+ "jsonrpsee 0.24.11",
  "pallet-ismp-demo",
  "pallet-ismp-host-executive",
  "pallet-ismp-relayer",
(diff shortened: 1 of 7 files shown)


Wizdave97:
There's no need for an additional endpoint, we can convert the ws endpoint to http like its done in the sdk

seunlanlege (commented):
amazing how easy it is to do this in rust

seunlanlege (approved):

PR #1185, opened 2026-08-28
[indexer]: index the gateway's OrderCancelled event

Wizdave97:
Indexes Intent Gateway `OrderCancelled(bytes32 indexed commitment, address canceller)` events and exposes cancellation initiation through the order's cancellation relation and status metadata. Cancellation never writes `IOrderV3.status`: a locally guarded full-row save can still overwrite a refund committed by an independent chain indexer with a different cached snapshot.

## Behavior

- Adds `IOrderV3Cancellation`, the `IOrderV3.cancellations` relation, datasource wiring, and the event handler. Event records retain the initiating chain, caller, timestamp, block number, and transaction hash. The caller can differ from the order creator on the permissionless destination route.
- Appends `CANCELLED` after existing enum values so the schema migration remains additive. The current gateway ABI already contains the event; it retains exactly one matching declaration.
- Writes `CANCELLED` status metadata for existing orders, including already-settled orders. Missing parents use the established pending-metadata workflow. Cancellation alone leaves the parent status unchanged; parent filters for `CANCELLED` will not identify cancellation initiation.
- Preserves `EscrowRefunded` as the `REFUNDED` transition. Existing lifecycle writers, pending-status flushing, and their accounting side effects remain unchanged.
- Adds SDK `OrderStatus.CANCELLED`. The status stream keeps polling after cancellation, gives existing terminal metadata precedence over a later cancellation entry, and avoids repeated emission of the same status.

## Validation

- Eight indexer service tests pass: metadata-only writes, all existing terminal states, cancellation before placement, pending-metadata materialization, duplicate/repeated logs, and same-chain cancellation followed by refund.
- Six SDK stream tests pass: cancellation stays intermediate, repeated polling emits it once, and a refund wins over cancellation with later or equal timestamps.
- A checked-in runtime regression script uses two independent SubQuery caches and Postgres connections. It reproduces the stale whole-row regression, verifies the actual service preserves `REFUNDED` in both cache flush orders, and checks the missing-parent path with multichain timestamp history. Reproduction commands are in the metadata-only decision document.
- Indexer `ENV=mainnet npm run build:release` and SDK browser/node/type builds pass.
- The indexer cancellation ABI declaration matches the compiled current `IntentGatewayV2` ABI.
- The prior refresh validated current-main-to-PR additive schema migration against SubQuery/Postgres, including preservation of an existing order and an unchanged restart. This follow-up does not change the schema.

## Known pre-existing SDK limitation

The order query requests `orders`, while its parser reads `orderPlaceds`; that legacy query shape also differs from `IOrderV3`. The focused stream tests supply the parser's expected response and do not establish live GraphQL compatibility. Migrating that query is separate from this cancellation persistence fix.


Wizdave97:
@royvardhan  please approve this

royvardhan (approved):

PR #1188, opened 2026-08-31
[simplex]: retry MPCVault execute on the create/execute race and name failing RPCs

royvardhan:
MPCVault intermittently rejects a signing-request uuid its own createSigningRequest just returned (`3 INVALID_ARGUMENT: Invalid uuid`), before the callback co-signer is contacted. Execute now retries INVALID_ARGUMENT/NOT_FOUND twice with short backoff. gRPC failures are rethrown naming the RPC, the signing-request uuid and MPCVault's x-request-id, and the app-level error guard also trips on code-only errors (MPCVault sends empty messages with only a code set).

## Assumptions

- The bug is a race between create and execute: MPCVault's backend has a replication lag, so execute can run before the request it is executing is visible. This is inferred from logs, not confirmed by MPCVault.
- The failing call is execute, not create. Create's only uuid is the vault uuid, which never changes and worked minutes before and after every failure.
- The failures left no callback-server logs because MPCVault only contacts the callback during execute, and the rejection happens before that step.
- Retrying is safe: INVALID_ARGUMENT and NOT_FOUND mean MPCVault did nothing, so nothing can be signed twice.

### DIFF
--- a/sdk/packages/simplex/CHANGELOG.md
+++ b/sdk/packages/simplex/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @hyperbridge/filler
 
+## 0.12.1
+
+### Patch Changes
+
+- MPCVault signing failures now name the failing RPC, the signing-request uuid and MPCVault's x-request-id, and the error guard also trips on code-only errors (MPCVault sends empty messages with only a code set). `executeSigningRequests` retries INVALID_ARGUMENT/NOT_FOUND twice with short backoff: MPCVault intermittently rejects a uuid its own createSigningRequest just returned, which failed live fills with `3 INVALID_ARGUMENT: Invalid uuid` before the callback co-signer was contacted.
+
 ## 0.12.0
 
 ### Minor Changes
--- a/sdk/packages/simplex/CHANGELOG.md
+++ b/sdk/packages/simplex/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Patch Changes
 
-- MPCVault signing failures now name the failing RPC, the signing-request uuid and MPCVault's x-request-id, and the error guard also trips on code-only errors (MPCVault sends empty messages with only a code set). `executeSigningRequests` retries INVALID_ARGUMENT/NOT_FOUND twice with short backoff: MPCVault intermittently rejects a uuid its own createSigningRequest just returned, which failed live fills with `3 INVALID_ARGUMENT: Invalid uuid` before the callback co-signer was contacted.
+- MPCVault signing failures now name the failing RPC, the signing-request uuid and MPCVault's x-request-id, and the error guard defensively trips on errors that carry only a non-zero code with an empty message. `executeSigningRequests` retries INVALID_ARGUMENT/NOT_FOUND twice with short backoff: MPCVault intermittently rejects a uuid its own createSigningRequest just returned, which failed live fills with `3 INVALID_ARGUMENT: Invalid uuid` before the callback co-signer was contacted.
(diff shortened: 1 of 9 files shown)


seunlanlege (approved):

Wizdave97:
[[polytope-labs/hyperbridge#1188](https://github.com/polytope-labs/hyperbridge/pull/1188)](https://github.com/polytope-labs/hyperbridge/pull/1188) — 134/-18 across `mpcvault.ts` plus docs. The shape of the change is right: naming the RPC, the uuid and the x-request-id is the piece that was actually blocking diagnosis, the retry is correctly scoped to codes that mean the server did nothing, the loop terminates, and `docs/ai/{ChangeLog,Decisions,Flow}.md` are all updated per CLAUDE.md. Five findings:

- [[mpcvault.ts:52](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/wallet/mpcvault.ts:52)](sdk/packages/simplex/src/services/wallet/mpcvault.ts:52) — `apiErrorText` treats the UNSPECIFIED (0) enum as an error, so an `Error` carrying the zero code throws where the old message-only guard succeeded; that would stop all signing. Also, the cited evidence for code-only errors (`{"code":16,"message":""}`) is a gRPC status envelope, not the in-band `Error`.
- [[mpcvault.ts:86](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/wallet/mpcvault.ts:86)](sdk/packages/simplex/src/services/wallet/mpcvault.ts:86) — `getLogger("mpc-vault")` binds to the process-wide context, which `Logger.ts` keeps sink-less on purpose for library use, so the retry warning and uuid line are invisible to embedded consumers who pass `SimplexOptions.logger`. Every other service takes a `LoggerContext`.
- [[mpcvault.ts:163](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/wallet/mpcvault.ts:163)](sdk/packages/simplex/src/services/wallet/mpcvault.ts:163) — exhausted retries leave the created signing request PENDING in the vault forever; `Decisions.md` reasons about avoiding exactly this orphaning, but the failure path still leaks one per failed signature.
- [[mpcvault.ts:113](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/wallet/mpcvault.ts:113)](sdk/packages/simplex/src/services/wallet/mpcvault.ts:113) — create gets no retry. The timing (failures at 0.6–1.7s vs callback at 1.4–2.1s in healthy runs) does not rule create out, so if the assumption is wrong the outage continues and only the error text improves.
- [[mpcvault.ts:142](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/wallet/mpcvault.ts:142)](sdk/packages/simplex/src/services/wallet/mpcvault.ts:142) — no automated coverage: the only mpcvault test is skipped without credentials, and the constructor builds a real gRPC client with no seam, so CI never exercises the loop.

The first two are worth fixing before merge; the last three are cheap follow-ups. Nothing here blocks the diagnostic value of shipping it — once a failure names its RPC, the create-vs-execute question settles itself.

royvardhan:
All addressed in 369f6fc5 except the create retry. The retry exists to absorb a read-after-write race on a freshly issued uuid, and create references nothing freshly issued — its only uuid is the constant vault uuid, so an INVALID_ARGUMENT there is deterministic and retries would just triple the latency of a real config error, while ambiguous failures would leak one pending request per attempt. Since failures now name their RPC, a live failure in create would identify itself and we can revisit with evidence.

Wizdave97 (approved):

Issue #1159, opened 2026-08-21
[indexer]: update solver balances after every order, not only on phantom snapshots


seunlanlege:
Solver (liquidity provider) balances in the IntentGatewayV3 indexer (`sdk/packages/indexer`) are only refreshed by the phantom order sweep. Every fill moves a solver's balances, and the events that record the fill already identify the solver and the tokens, so the indexer can refresh the affected balances per order as well.

## Today

`handlePhantomOrderPrices` (`src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts`) runs on `PhantomBidWindowExhausted`. The SDK's `aggregatePhantomBids` (`sdk/packages/sdk/src/protocols/intents/phantom-aggregation.ts`) sweeps every bidding solver's balance of every configured token on every configured chain — `balanceOf` plus `maxWithdraw` on each ERC-4626 vault in `yieldVaults`, plus the Uniswap V4 positions declared in the bid on the bid's own chain — and the handler writes one `LiquidityProviderBalanceV2` row per (solver, chain, token). `updateLiquidityPools` (`src/services/liquidityPool.service.ts`) derives `PoolBidder.liquidity`, `PoolChainLiquidity.depth`, `PoolRoute.depth` and `LiquidityPool.sellDepth`/`buyDepth` from the same reads.

On mainnet the pallet generates phantom orders every `PhantomOrderInterval = 300` Hyperbridge blocks and closes the bid window 25 blocks later, so balances are sampled roughly every 34 minutes (generations at blocks 11569292 → 11569592 were 19:12:00 → 19:45:36 UTC on 2026-08-21). Everything in between is invisible until the next sweep: a solver that fills a large order has its destination-chain inventory overstated for up to half an hour, and `queryAvailableLiquidity` (`LiquidityEngine`, reading `PoolChainLiquidity.depth` / `PoolRoute.depth`) keeps advertising inventory that has already been spent. Overstating is the direction that hurts — a consumer sizes an order against depth that no longer exists. Mainnet currently sees ~36 fills/day with bursts of several per hour, so the gap is exercised routinely.

## What the events already tell us

Every balance-moving step of an order emits an event the indexer already handles in `src/handlers/events/intentGatewayV3/`, and the yield-vault handlers add a third trigger:

[...]

PR #1194, opened 2026-09-01
[indexer]: Refresh solver liquidity on every event that moves it

Wizdave97:
Closes the rest of #1159. #1192 landed the first half: `OrderFilled` / `PartialFill`
re-read the LPs backing the pools a fill traded through. This adds the events, the
inventory, and the replayability that were still missing.

### Uniswap V4 positions are now part of the inventory

A bid's `paymasterAndData` is the only place a position is ever named, so a balance
re-read could not see that inventory at all — and carrying the last sweep's *value*
forward would be worse, because simplex funds fills out of these positions: such a
fill drains the position inside the fill transaction while wallet and vault balances
barely move, so a carried value keeps advertising exactly what was just spent.

`aggregatePhantomBids` now reports the tokenIds it verified, and the phantom handler
records them in **`SolverV4Positions`** — one row per solver, keyed by its address,
holding the tokenIds from its latest bid and the chain that bid was for, replaced
wholesale each time it bids. The refresh reads that row with a single keyed `get` and
re-values each position on-chain at the event's block.

- Bids without declaring → row emptied (that is a withdrawal of the declaration).
- Skips a window → row kept; silence is not a withdrawal.
- tokenIds are stored, never the value — the value is what a fill drains.

### Reads are pinned to the event's block

`getTotalSolverBalance` is exported from the SDK with a `blockTag`, and
`memoizedSolverBalance` takes a per-chain map of them. A re-read now returns the same
value on a replay as it did live, instead of stamping today's balance onto a
historical event. Only the event's own chain is pinned — a refresh reaches across
every chain a pool is quoted on, and a block number means nothing on any other.

### Two more triggers

- **`EscrowReleased`** (source chain): the solver was just paid the order's inputs
  back, so its inventory there *rose*. The event names no filler, so the handler reads
  the gateway's `_filled(commitment)` at that block — `_withdraw` writes the
  beneficiary in the same call that emits the event, so this never waits on the
  destination chain's node.
- **Vault `Deposit` / `Withdraw`**: from the end of `recordLedger`, where the
  known-solver gate and duplicate-log guard already live. An LP's own principal only
  shifts between the raw and vault halves of one total, but the total moves when the
  counterparty is someone else, and no order event reports that.

Both name a solver and a token but no pool, so they enter through a new
`refreshProviderLiquidity` alongside the pool-scoped entry point.

### Worth a careful look in review

- **Depths are re-summed over every stored bidder row of a `(pool, chain)`, read
  before the writes.** Summing only the re-read subset was correct solely because the
  pool path re-reads everyone; with one solver's rows in hand it would erase the
  others. Reading back *after* the writes would instead rest on the store reflecting
  an in-block removal.
- **`SolverV4Positions` keyed by solver alone assumes a solver declares on one
  chain**, which holds while Uniswap V4 is configured for Base only. The writer
  refuses to overwrite a row recorded on another chain and warns, rather than silently
  deleting real inventory; the key grows to `(chain, solver)` the day a second V4
  chain is configured.
- **The declaration is a pointer, not a claim.** Positions are recorded after the
  aggregation's ownership check, and the owner is checked again on every re-read — a
  row from last window cannot know the position has since been sold.

### Schema

One new entity, no existing entity altered — `LiquidityProvider` gains no
`@derivedFrom` field. #1159 §3 (the `trigger` enum, `transactionHash`, the
event-row id shape, ordering current liquidity by `snapshotTime`) is deliberately
deferred for that reason: all four change `LiquidityProviderBalanceV2`, which is live.
Event-triggered rows keep borrowing Hyperbridge's head block, and the design is left
as a comment on `recordProviderBalances` for whenever a migration is on the table.

### Tests

- SDK `phantomAggregation`: 44 (2 new — verified positions reported, unowned ones not).
- `liquidityPoolRefresh`: 28 (declaration read once per solver, other-chain rows
  ignored, drained / burned / transferred positions, unreadable position, the
  provider-scoped path).
- `phantomOrder.handlers`: 21 (5 new on recording — wholesale replace, emptying,
  keeping an absent solver's row, the cross-chain warning).

`docs/ai/{ChangeLog,Decisions,Flow}.md` updated in both packages.


sdk/packages/indexer/src/handlers/events/substrateChains/handlePhantomOrderPrices.handler.ts
```diff
@@ -166,6 +167,17 @@ export const handlePhantomOrderPrices = wrap(async (event: SubstrateEvent): Prom
 		}).save()
 	}
 
+	// What this window's bidders declared, recorded so a later fill can re-read those positions: the
+	// bid is the only place they are named. Every verified bid sweeps balances, so lpBalances names
+	// the bidders — except one holding nothing anywhere, whose only trace is its declaration.
+	await recordDeclaredPositions({
+		chain: phantom.chain,
+		bidders: [...aggregate.lpBalances.map((lp) => lp.solver), ...aggregate.positions.map((p) => p.solver)],
```
royvardhan:
```ts
bidders: [...aggregate.lpBalances.map((lp) => lp.solver), ...aggregate.positions.map((p) => p.solver)],
```
A solver that bids without declaring positions and sweeps zero balance everywhere lands in neither `lpBalances` nor `positions`, so it is missing from `bidders` and its old declaration row is never emptied. The refresh then keeps valuing positions it stopped offering, and a V4-funded solver with a near-empty wallet is exactly the profile this PR targets. Better to have `aggregatePhantomBids` return the verified solver set (it already tracks `countedSolvers`) and use that here.

sdk/packages/indexer/src/services/intentGatewayV3.service.ts
```diff
@@ -783,12 +782,77 @@ export class IntentGatewayV3Service {
...
+		})
+	}
+
+	/**
+	 * The same refresh for a vault deposit or withdrawal. An LP moving its own principal only
+	 * shifts inventory between the raw and vault halves of one total, which the re-read confirms
+	 * rather than changes; the total does move when the counterparty is someone else (a treasury
+	 * funding the solver, or inventory leaving it), and no order event reports that at all.
+	 */
+	static async refreshLiquidityAfterVaultEvent(params: {
```
royvardhan:
```ts
static async refreshLiquidityAfterVaultEvent(params: {
```
Never called. `YieldVaultService.recordLedger` calls `refreshProviderLiquidity` with `liquidityRefreshContext` directly, so this can be dropped, or the vault refresh should go through it.

royvardhan (approved):

PR #1196, opened 2026-09-02
[simplex]: prefer Simplex paymaster over Circle in selection

royvardhan:
Prefer the Simplex paymaster first, falling back to Circle. Deposit gates unchanged; tests updated; filler bumped to 0.12.3.

A Simplex builder failure (RPC error, bootstrap approve failure, missing native dust) now demotes Simplex to a skip reason and falls through to Circle instead of aborting selection. The token permit probe no longer treats a transport error as "no permit".

Removes `skipPermit`. Delegation ops were forced past the Simplex paymaster's EIP-2612 PERMIT mode into the PERMIT2/APPROVE branch, whose bootstrap needs a native-funded `approve`. A solver holding no native could not send it, so sponsored delegation failed and the direct EIP-7702 fallback failed too (seen on Base and Arbitrum). The permit runs in the paymaster frame under its own limit, so the fixed account-side gas limits the flag was protecting were never at risk. Delegation on a permit-capable token now needs no native at all.


Wizdave97:
## Review: [[polytope-labs/hyperbridge#1196](https://github.com/polytope-labs/hyperbridge/pull/1196)](https://github.com/polytope-labs/hyperbridge/pull/1196) — prefer Simplex paymaster over Circle

The reorder itself is clean — gate semantics, headroom, fail-open reads and the `type: "none"` accumulation are all preserved, and the tests were properly inverted rather than just re-labelled. The problems are all consequences of *what now runs first*: Simplex's builder is far more side-effecting and far more throw-prone than Circle's, and it was previously shielded by Circle winning on most chains.

**Findings**

- [[src/services/paymaster/index.ts:61](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/paymaster/index.ts:61)](sdk/packages/simplex/src/services/paymaster/index.ts:61) — Dropping the Circle-only `paymasterVerificationGasLimit` override likely trips rundler's 0.4 verification-efficiency floor for the three fixed-gas senders; by the model in `VaultFundingPlanner.ts:354-357` the vault sweep goes from 99k/200k = 0.50 to 99k/310k = 0.32, and `trySendSponsored` swallows the rejection, so those ops silently start paying native gas.
- [[src/services/paymaster/index.ts:71](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/paymaster/index.ts:71)](sdk/packages/simplex/src/services/paymaster/index.ts:71) — Only a `null` return falls through to Circle; every throwing path in `buildSimplexPaymasterData` (transport error in `selectToken`, the deliberate rethrow in `paymasterSupportsPermit2`, the native-dust `throw` in `sendFundedApprove`) now aborts selection before Circle is evaluated — contradicting the fail-open intent of the adjacent `depositShortfall`.
- [[src/services/paymaster/index.ts:71](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/paymaster/index.ts:71)](sdk/packages/simplex/src/services/paymaster/index.ts:71) — The native-funded bootstrap approve (up to two txs, each awaiting two confirmations) now runs inside `prepareBidUserOp`; neither it nor `fx.ts` `submitBid` has a try/catch, so a slow or reverting approve costs the bid.
- [[src/services/paymaster/index.ts:71](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/paymaster/index.ts:71)](sdk/packages/simplex/src/services/paymaster/index.ts:71) — `tokenSupportsPermit`'s bare `catch { return false }` turns an RPC blip into "token has no permit", newly reachable on every Circle-configured chain, producing an unnecessary `approve(Permit2, max)` from the solver EOA.

The first two are worth blocking on. #2 has a small fix — wrap the Simplex block in try/catch and push the error into `skipReasons` — which also softens #3 and #4. #1 needs either extending the override to Simplex's modes or measuring the modes against the floor before this ships; the Decisions.md entry treats it as a wait-and-see, but the arithmetic already in the tree suggests it fails immediately.

royvardhan:
Verified all four against the code — pushed fixes for #2 and #4 (ef69f2a):

- **2**: confirmed. The Simplex builder call is now wrapped in try/catch — a throw warns, joins `skipReasons` as `simplex: <message>`, and falls through to Circle. The deposit gate stays outside the catch (it already fails open). Added selection-level tests for a throwing builder.
- **4**: confirmed. `tokenSupportsPermit` now uses the same revert-vs-transport discrimination as `paymasterSupportsPermit2`; a transport error propagates into the new catch and demotes to Circle instead of reading as "no permit".
- **1**: the mechanism is real but the 0.32 arithmetic holds the usage numerator constant across paymasters — ~75k of the 99k *is* Circle's own validation. Simplex PERMIT mode executes the permit during validation (~113k+), so the ratio lands near 137k/310k ≈ 0.44, above the floor. The APPROVE mode is the unmeasured one that could sit low; a floor rejection degrades to a native tx with the existing "Bundler rejected UserOp" warn (all three fixed-gas senders handle the `null`), so we're monitoring rather than retuning pre-emptively. Recorded in Decisions.md.
- **3**: accepted — the bootstrap is once per fresh solver, only for a non-permit token, and with #2's fix its failure modes now demote to Circle instead of costing the bid.

---

Wizdave97 (approved):

PR #1198, opened 2026-09-03
[simplex]: Stop bidding on phantom orders whose window has closed

Wizdave97:
A mainnet filler spent nine hours bidding on orders that had expired hours earlier. Its bids landed, nothing errored, and it backed no pool the whole time: 0 PoolBidder rows against 692 balance rows whose newest stops at 03:03, while healthy fillers kept writing them.

The lag behind the order it bid on grew monotonically - 4 blocks, then 147, 1611, 2228, 2765, 3448 - the signature of a cursor that cannot catch up rather than a stall. pollPhantomOrders advances by at most maxBlocksPerPoll (10) a tick and never skips ahead, so a deficit accumulated while ticks were lost (a scan overrunning the 15s interval, or the rate-limit backoff sitting out up to 8 ticks) is only repaid if the sustained rate beats the chain's. Below that it stays behind forever, and every order it then delivers is already dead, because the bid window is tens of blocks.

Fixed at both places the delay comes from:

- SDK: new maxLagBlocks option. Past that, the cursor jumps to the head and reports the abandoned range through onSkip. Off by default - advancing only past blocks really read is what makes an outage delay orders instead of dropping them, which a consumer reading this feed as history wants. Only a bidder knows the backlog is worthless, so only a bidder asks. Also adds latestBlockNumber(), since an order carries its registration block and no clock.
- simplex: the scanner opts in at 60 blocks, and handlePhantomOrders drops events older than 40 blocks against one head read per batch. That covers the delay the scanner cannot see - the global queue, and the quoting in preparePhantomBid - and turns a silent outage into a warning naming the lag. A failed head read keeps every event: one flaky endpoint must not stop a filler bidding.

Bidding late is not a harmless no-op: the extrinsic is accepted and reserves a deposit, while the aggregation read that order's bids when its window closed. That is what made this invisible - the filler's own logs said "Phantom bids submitted" throughout.


royvardhan (commented):
Two things, both verified against the pallet and the live Nexus chain. The mechanics themselves (cold-start guard, skipped range via `onSkip`, fail-open head read) check out and both test files pass locally.

**`sdk/packages/simplex/src/core/filler.ts:42`**
```ts
const MAX_PHANTOM_ORDER_AGE_BLOCKS = 40
```
The bid window is chain-owned and governance-set. Nexus currently has `PhantomBidWindow` = 15 (the runtime constant is 25, Gargantua's is 5), so this gate still lets through orders 16 to 40 blocks old. With the generation interval at 55 those orders are still current and the pallet rejects the bid with `PhantomOrderBidWindowClosed`, so it burns fees and error logs rather than deposits, but it is roughly 25 blocks of bids the gate exists to prevent. If governance raises the window past 40 this constant starts dropping live orders instead.

Suggest reading the effective window from `api.query.intentsCoprocessor.phantomBidWindow()` (falling back to `api.consts.intentsCoprocessor.phantomOrderBidWindowBlocks` when zero), exposed from the SDK next to `latestBlockNumber()`, and sizing the gate as window plus a small fixed margin. The same applies to `MAX_LAG_BLOCKS` in `intentsCoprocessor.ts:102`, which could be derived rather than fixed at 60.

**`sdk/packages/simplex/docs/ai/Decisions.md:9`** and **`sdk/packages/sdk/docs/ai/Flow.md:23`**
```
Chosen: two independent guards — the poll skips a backlog it cannot bid on (`maxLagBlocks`), and the filler
```
```
2. On the first successful head read the cursor is set to `head - 1 - lookbackBlocks`; ...
```
The final commit replaced the `maxLagBlocks` option with the `MAX_LAG_BLOCKS` constant and removed `lookbackBlocks`, but the Decisions entry still names the option, the SDK Flow doc still describes the lookback cursor and does not mention the skip-ahead step, and the PR description still says the scanner "opts in at 60 blocks". Flow.md is a documented control-flow change so it needs the update; the others just need the stale names fixed.

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -31,6 +31,16 @@ import { Decimal } from "decimal.js"
...
 
+/**
+ * How old a phantom order may be, in Hyperbridge blocks, and still be worth bidding on.
+ *
+ * The pallet closes an order's bid window tens of blocks after registering it, so this is that
+ * window plus room for the delay between the poll seeing the block and this bid being signed. Set
+ * it too high and the filler burns deposits on windows that already closed; too low and a slow but
+ * still useful bid is thrown away before it is tried.
+ */
+const MAX_PHANTOM_ORDER_AGE_BLOCKS = 40
```
royvardhan: The window this is sized off is governance-set, and it has already moved. Nexus has `PhantomBidWindow` at 15 right now (the runtime constant says 25, gargantua's is 5), so with 40 here we still bid on orders 16 to 40 blocks old. The interval is 55 so those orders are still current and the pallet bounces them with `PhantomOrderBidWindowClosed`. That is fees and error logs rather than reserved deposits, but it is 25 blocks of the exact bids this is meant to stop. And if governance ever raises the window past 40 we start dropping live orders.

Can we read it from chain instead? `api.query.intentsCoprocessor.phantomBidWindow()`, falling back to `api.consts.intentsCoprocessor.phantomOrderBidWindowBlocks` when it is zero, and then window plus a small margin here.

sdk/packages/sdk/src/chains/intentsCoprocessor.ts
```diff
@@ -85,6 +85,22 @@ const DEFAULT_RPC_MAX_RPS = 8
...
+ * a deposit for nothing. The lag is also self-sustaining: the cursor gains at most
+ * `maxBlocksPerPoll` per tick, so a deficit accumulated while ticks were lost (a scan overrunning
+ * the interval, or a rate-limit backoff sitting them out) is never repaid unless the sustained rate
+ * exceeds the chain's. A filler can sit hours behind indefinitely while looking healthy — which is
+ * exactly what happened on mainnet.
+ *
+ * Sized at a few bid windows, so an ordinary hiccup still recovers every order it can and only a lag
+ * no bid could survive is skipped.
+ */
+const MAX_LAG_BLOCKS = 60
```
royvardhan: Same thought as the filler gate: this could be derived from the on-chain bid window rather than fixed at 60.

sdk/packages/simplex/docs/ai/Decisions.md
```diff
@@ -4,6 +4,31 @@ AI-maintained record of non-obvious choices made in `sdk/packages/simplex`: what
 
 Entry format: heading with the decision, then alternatives considered and the reasoning. Newest first.
 
+## 2026-09-03 — A stale phantom order is dropped, not bid on, and the drop is loud
+
+Chosen: two independent guards — the poll skips a backlog it cannot bid on (`maxLagBlocks`), and the filler
```
royvardhan: `maxLagBlocks` no longer exists after the last commit, it is the `MAX_LAG_BLOCKS` constant now. The PR description also still says the scanner opts in at 60 blocks.

sdk/packages/sdk/docs/ai/ChangeLog.md
```diff
@@ -12,6 +12,32 @@ Files: list of files touched.
...
+walking that backlog is the right answer. The check runs only once the cursor is established, so a cold start —
+exactly one block behind — is never read as a backlog.
+
+That bounds the property the cursor was built for. It still advances only past blocks whose events were really
+read, so an outage delays orders rather than dropping them — up to the point where the delayed orders are dead
+anyway.
+
+`lookbackBlocks` is gone with it. It existed so a restarting process could reach back and still bid on a window
+already open, which is the same late bid this change is removing: by the time the process is up, that window has
+all but closed. Nothing passed it, and a cold start now begins at the head.
```
royvardhan, line 35: Flow.md still has the old cursor in step 2 (`head - 1 - lookbackBlocks`) and nothing about the skip ahead. Needs updating with this.

sdk/packages/simplex/src/core/filler.ts
Wizdave97: Yeah

royvardhan (approved):
Verified df6028e against the pallet and locally.

- `phantomTimings()` reads `PhantomBidWindow` / `PhantomOrderInterval` and falls back to the `PhantomOrderBidWindowBlocks` constant when the stored window is zero, which is exactly what the pallet's `phantom_bid_window()` does. Cached once per instance, and a failed read clears the cache so it retries.
- Poll skip threshold is `window + max(interval, window)` and the jump lands at `head - 1 - window`, so the first block scanned after a skip is `head - window`, whose orders the pallet still accepts (`block_number <= created_at + window`). The cursor is back inside the threshold after a jump, so it does not skip again next tick; the new test pins that.
- Filler gate is `window + 2`, permissive on the edge as intended, with the pallet as the authority there.
- Flow.md step 2 now describes the cold-start cursor and the skip; the stale `maxLagBlocks` / `lookbackBlocks` references in Decisions are gone.

`pollPhantomOrders.test.ts` (37) and `phantom-bid-staleness.test.ts` (7) pass locally on the head commit; both CI jobs are green.

One non-blocking nit: the PR description still describes the first commit (`maxLagBlocks` option, opt-in at 60, drop at 40). Worth refreshing before merge since it becomes the merge commit message.

---

PR #1199, opened 2026-09-03
[evm, sdk, host-executive]: gate cross-chain deliveries behind a whitelisted relayer and fix HostManager rotation addressing

royvardhan:
Adds a single authorised relayer to IntentGatewayV2, BridgeToken and HostManager. Their inbound callbacks revert unless the message was submitted by that relayer, so a forged consensus alone cannot land a message. The host records a refused delivery as undelivered and the relayer can resubmit it.

- Gateway: `_relayer` appended to storage (slot 13, packed behind `_paused`). `setRelayer` is callable by `_owner`, or as `UpgradeContract` migration calldata so the relayer is armed in the upgrade transaction. Zero fails closed. The unused `_paused` getter is dropped to stay under EIP-170.
- HyperFungibleToken: opt-in gate with an owner-only `setRelayer`. BridgeToken overrides `_checkRelayer` to fail closed; its deploy script sets the relayer from `BRIDGE_RELAYER` before `configure`.
- HostManager: same gate on governance deliveries, set by the host admin. This closes the route where a forged `SetHostParam` swaps the host's handler for one that reports the whitelisted relayer on any message. The HostManager carries no user traffic, so permissionless relaying is unaffected.
- Host executive: `update_host_params` addressed a HostManager rotation to the new manager, which the host does not yet trust, so the rotation could never land while Hyperbridge recorded it as done and every later update and withdrawal for that chain was sent to the wrong contract. The request is now addressed to the current manager; the payload still installs the new one. Covered by a pallet test that pins the recipient and Foundry tests that play both addressings through the real host.
- Tests include upgrading the live mainnet proxy on a fork with state compared before and after, and the forged handler swap played through the real host.

Rollout: deploy the new gateway implementation and `upgrade_gateway` per chain with `init_data = cast calldata "setRelayer(address)" <relayer>`; deploy a new HostManager per chain with `DeployHostManager.s.sol`, set its relayer, then point the host at it with `SetHostParam`. The rotation needs the runtime fix released first, since the current runtime addresses it to the new manager.

### DIFF
--- a/evm/script/DeployBridgeToken.s.sol
+++ b/evm/script/DeployBridgeToken.s.sol
@@ -9,14 +9,21 @@ import {BaseScript} from "./BaseScript.sol";
 contract DeployBridgeToken is BaseScript {
     function deploy() internal override {
         address dispatcher = config.get("CALL_DISPATCHER").toAddress();
+        // The only account allowed to deliver messages to the token, on every chain. Read here rather
+        // than in BaseScript so the other scripts do not require it.
+        address relayer = vm.envAddress("BRIDGE_RELAYER");
         BridgeToken bridge = new BridgeToken{salt: salt}(admin);
 
+        // Set before `configure`: until the host is set nothing can reach `onAccept`, so the token
+        // is never live without its relayer.
+        bridge.setRelayer(relayer);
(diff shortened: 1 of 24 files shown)


Wizdave97:
Please bump nexus runtime version

Wizdave97 (approved):

PR #1210, opened 2026-09-06
[simplex]: widen the dashboard container to 150rem

seunlanlege:
The operator dashboard and setup wizard were capped at 80rem (1280px) by `.app-container`, which left the order history roughly 900px of usable width on wide screens once the 16rem sidebar and column padding were taken out. The cap is now 150rem (2400px).

## Test plan
- [x] `pnpm ui:check` (biome, tsc, vite) passes


seunlanlege:
Superseded by #1212, which carries this commit along with the rest of the dashboard work.

PR #1213, opened 2026-09-07
[evm, sdk, runtime]: gate SimplexPaymaster governance and remove the approve mode

royvardhan:
`SimplexPaymaster.onAccept` now accepts governance deliveries only from the authorised relayer, armed by `initialize` on a bare proxy or `migrate(relayer)` as the upgrade init data on the live proxies, and rotated by the new `SetRelayer` request kind and `set_paymaster_relayer` extrinsic. Mode byte `0x01` is refused and the simplex client no longer reads or creates an allowance to the paymaster. Publish simplex 0.14.0 only after the live proxies are upgraded.


evm/script/DeploySimplexPaymaster.s.sol
```diff
@@ -19,6 +19,8 @@ contract DeployScript is BaseScript {
         // 24h avoids transient StaleOraclePrice reverts on late pushes.
         uint256 maxOracleAge = vm.envOr("MAX_ORACLE_AGE", uint256(90_000));
         uint256 swapSlippageBps = vm.envOr("SWAP_SLIPPAGE_BPS", uint256(200)); // default 2%
+        address relayer = vm.envAddress("PAYMASTER_RELAYER");
```
seunlanlege:
```suggestion
        address relayer = vm.envAddress("GOVERNANCE_RELAYER");
```

evm/src/utils/SimplexPaymaster.sol
```diff
@@ -191,14 +206,22 @@ contract SimplexPaymaster is Initializable, HyperApp, PaymasterERC20 {
 
     uint256 public swapSlippageBps;
 
-    uint256[49] private __gap;
+    /// @dev The only relayer whose `onAccept` deliveries are accepted; zero means every relayer.
+    address private _relayer;
+
+    uint256[48] private __gap;
```
seunlanlege, line 212: y tho

sdk/packages/simplex/src/services/paymaster/provider/simplex.ts
```diff
@@ -40,15 +38,15 @@ const APPROVE_TX_GAS = 60_000n
...
  * - PERMIT (0x00) when the token supports EIP-2612: a permit executed during validation
- * - PERMIT2 (0x02) when the token is already approved to Permit2: a per-op, single-use
- *   Permit2 signature; nothing is exposed to the paymaster at rest
- * - APPROVE (0x01) while a legacy allowance to the paymaster is still in place
- * - otherwise one funded bootstrap tx: approve(Permit2, max) where Permit2 exists and
- *   the paymaster deployment supports it (then PERMIT2 for the account's lifetime),
- *   else a capped approve(paymaster)
+ * - PERMIT2 (0x02) otherwise: a per-op, single-use Permit2 signature; nothing is exposed
+ *   to the paymaster at rest. A token not yet approved to Permit2 costs one funded
+ *   bootstrap tx, approve(Permit2, max), for the account's lifetime.
```
seunlanlege, line 45: chains that don't have eip-2612, can only delegate via native right? So maybe add the permit2 approval to ride along with the 7702 delegation?

evm/src/utils/SimplexPaymaster.sol
royvardhan: `_relayer` took the slot just before the gap, so the gap gives up one to keep the contract's footprint at the same 57 slots. Just the usual convention when a reserved slot gets used; nothing inherits from this contract so it is not load-bearing. Happy to leave it at 49 if you prefer the smaller diff.

sdk/packages/simplex/src/services/paymaster/provider/simplex.ts
royvardhan: Already there on the native delegation path: `resolvePendingPermit2Approval` plus `sendDelegationTransaction` batch `approve(Permit2, max)` into the set-code tx. It can't happen on the sponsored path for the first op, since the paymaster prefunds during validation, before the op's calldata runs. Only improvement left is going straight to the batched native tx when a Permit2 approval is pending instead of trying the sponsored path first (two txs today). Can do that here or as a follow-up.

evm/src/utils/SimplexPaymaster.sol
seunlanlege: oh i was wondering why the gap variable, but iiuc, we're using it to reserve storage slots?

royvardhan: Yes

sdk/packages/simplex/src/services/paymaster/provider/simplex.ts
seunlanlege: yeah best to go straight to batched native tx

seunlanlege:
conflicts

seunlanlege (approved):

PR #1217, opened 2026-09-07
[simplex]: remote access over ssh

seunlanlege:
Simplex's web UI is loopback-only and most operators run it behind NAT, so there was no way to check on a filler from a phone. This adds an opt-in remote access tunnel, plus the wallet and docs work that came out of testing it.

## Remote access

`[simplex.tunnel]`, off unless `enabled = true`. Simplex keeps an outbound SSH session to a relay ([polytope-labs/simplex-tunnel](https://github.com/polytope-labs/simplex-tunnel), hosted at `simplex.tunnel.polytope.technology`) which leases it a stable public port. A device's SSH client connects there with a local port forward and opens the dashboard at `localhost`.

The relay never sees dashboard traffic. The device's SSH session terminates in an SSH server embedded in this process, not in sshd and not in the relay, so the relay only moves ciphertext. That embedded server is the most exposed code here — anyone scanning the relay can reach it — so it accepts public-key auth against paired devices and `direct-tcpip` to the UI bind, and nothing else: no shell, no exec, no other destination. It also enforces a pre-auth timeout, three failures per connection, and ten per source in ten minutes.

- **Four keys, deliberately separate.** Operator key (identifies this filler to the relay, decides its port), simplex host key (pinned by the device), one key per device, and the relay's host key (pinned by simplex — the hosted relay's fingerprint ships as the default, so first contact is verified rather than trusted). Losing a device revokes one line and changes nothing else.
- **Pairing pastes the device's public key** by default, so the private half never leaves it. Generating the pair in simplex stays available for apps that cannot make their own, and shows the key once with a QR code.
- **Best-effort.** It starts after the filler is up, never during the setup wizard, reconnects with backoff, and a tunnel failure never touches filling.

Keys live in `<data-dir>/tunnel/` in plain OpenSSH formats.

## Wallet

- **Send funds now opens a review dialog** instead of a `window.confirm` string. Token logo with the network badged on it, the amount, the full recipient, balances, and the split the sender will actually make between the wallet and the vault. That vault line is a real check: `TokenSender` withdraws from one vault rather than several, so when the shortfall exceeds the largest single vault the dialog says the transfer will fail before it is submitted.
- **The ledger gained action filters and pagination.** `Pager` moved out of `Orders.tsx` into a shared component so the two pages cannot drift.

## Docs

A new "Accessing the dashboard" section on the operator page: the loopback default and why the UI is unauthenticated, the Docker port prefix, then SSH port forwarding for a host you can already reach and the tunnel for one behind NAT. It sits in the existing page so the "keep it on loopback" warning and the ways around it are read together.

## Testing

- 24 new tests. `tunnel.test.ts` runs a fake relay in-process and drives a real `ssh2` client through it; `ui-server-tunnel.test.ts` covers the routes. Full suite passes.
- End to end against the deployed relay: an OpenSSH operator leased a port and kept it across a reconnect, simplex's own client connected with the fingerprint pinned, and a device reached the dashboard through `simplex.tunnel.polytope.technology` with strict host-key checking. Shell attempts, other destinations and unpaired keys were all refused.
- Paired and used from Termius on iOS.

## Notes for review

- `ssh2` is CommonJS, so the tunnel modules use default imports. Named imports type-check and pass under vitest but throw at load in the shipped ESM binary, which the smoke test caught.
- `sdk/pnpm-workspace.yaml` declines the optional `ssh2` and `cpu-features` build scripts; ssh2 works in pure JS. A clean `--frozen-lockfile` install passes.
- Simplex bumped to **0.15.0**. Main released 0.14.0 for the paymaster work while this branch was also sitting on 0.14.0; both sides wrote the same string so git merged it silently, and the collision was only visible against main. The sdk is untouched.


sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -281,6 +291,17 @@ program
...
 				runtime = simplex.internals
+				// Remote access rides on the operator-mode UI only: it starts here,
+				// after the filler is up, never while the setup wizard holds secrets.
+				// A wildcard UI bind is reached on loopback; a specific address as-is.
+				const uiHost =
+					isLoopbackHost(uiBind.host) || uiBind.host === "0.0.0.0" || uiBind.host === "::" ? "127.0.0.1" : uiBind.host
+				tunnel = new TunnelService({
+					dataDir: resolveDataDir(options.dataDir),
+					config: config.simplex.tunnel,
+					uiTarget: () => ({ host: uiHost, port: uiBoundPort }),
```
ddboy19912: We should only start the tunnel only after the UI binds successfully. Otherwise, `--no-ui` or a port conflict could expose another localhost service through the tunnel.


Potential Fix:

```
 const startFiller = async (config: FillerConfigFile, path: string) => {
   // Start filler only.
   simplex = await Simplex.start(/* ... */)
   runtime = simplex.internals
-
-  tunnel = new TunnelService({
-    dataDir: resolveDataDir(options.dataDir),
-    config: config.simplex.tunnel,
-    uiTarget: () => ({ host: uiHost, port: uiBoundPort }),
-  })
-  tunnel.start()
   return simplex
 }

 if (uiEnabled) {
+  try {
+    tunnel = new TunnelService({
+      dataDir: resolveDataDir(options.dataDir),
+      config: config.simplex.tunnel,
+      uiTarget: () => ({ host: uiHost, port: uiBoundPort }),
+    })
+  } catch (err) {
+    logger.error({ err }, "Remote access unavailable")
+  }
+
   uiServer = new UiServer({
     mode: "operator",
     uiDistDir: resolveUiDistDir(),
     operator: await operatorContextFrom(simplex!, () => shutdown("UI"), tunnel),
   })

   try {
     uiBoundPort = await uiServer.start(uiBind.port, uiBind.host)
+    tunnel?.start()
   } catch (err) {
+    await tunnel?.stop()
+    tunnel = undefined
     logger.error({ err }, "UI server failed to start")
     uiServer = undefined
   }
 }
```

For the wizard transition:

```
 onSaveAndStart: async (config, _toml, path) => {
   await startFiller(config, path)
+  tunnel = createTunnel(config)
   server.enterOperatorMode(
     await operatorContextFrom(simplex!, () => shutdown("UI"), tunnel),
   )
+  tunnel?.start()
 },
```

sdk/packages/simplex/src/bin/simplex.ts
```diff
@@ -281,6 +291,17 @@ program
 					watchOnly: options.watchOnly,
 				})
 				runtime = simplex.internals
+				// Remote access rides on the operator-mode UI only: it starts here,
+				// after the filler is up, never while the setup wizard holds secrets.
+				// A wildcard UI bind is reached on loopback; a specific address as-is.
+				const uiHost =
+					isLoopbackHost(uiBind.host) || uiBind.host === "0.0.0.0" || uiBind.host === "::" ? "127.0.0.1" : uiBind.host
+				tunnel = new TunnelService({
```
ddboy19912: The tunnel initialization could terminate filling. We should catch tunnel startup failures and let filling continue without remote access.

Potential Fix:
```
- tunnel = new TunnelService(options)
- tunnel.start()
+ try {
+   tunnel = new TunnelService(options)
+   tunnel.start()
+ } catch (err) {
+   tunnel = undefined
+   logger.error(
+     { err },
+     "Remote access failed to start; filling will continue",
+   )
+ }
```

sdk/packages/simplex/src/services/tunnel/EmbeddedSshServer.ts
```diff
@@ -0,0 +1,190 @@
...
+		let failures = 0
+		let authenticated = false
+		const authTimer = setTimeout(() => {
+			if (!authenticated) {
+				this.logger.debug({ origin }, "Dropping tunnel connection: not authenticated in time")
+				conn.end()
+			}
+		}, this.authTimeoutMs)
+
+		conn.on("authentication", (ctx) => {
```
ddboy19912, line 293: Revoked devices retain active access. We should also invalidate existing SSH sessions, not only block new logins. Close active connections or reject future channels.

Later channels do not recheck authorization: 
https://github.com/polytope-labs/hyperbridge/blob/f53f397c122eeb2f0c7255ae8ab0b7c5004665ba/sdk/packages/simplex/src/services/tunnel/EmbeddedSshServer.ts#L135

Potential Fix:
```
 export class EmbeddedSshServer {
+  private readonly connectionsByFingerprint =
+    new Map<string, Set<Connection>>()

+  disconnectDevice(fingerprint: string): void {
+    const connections = this.connectionsByFingerprint.get(fingerprint)
+    if (!connections) return
+
+    for (const connection of connections) connection.end()
+    this.connectionsByFingerprint.delete(fingerprint)
+  }

   private onConnection(conn: Connection, ip: string, port: number): void {
+    let authenticatedFingerprint: string | undefined

     conn.on("authentication", (ctx) => {
       // ...
       const fingerprint = fingerprintOf(ctx.key.data)
       if (!this.opts.isAuthorized(fingerprint)) {
         return fail(`unknown device key ${fingerprint}`)
       }

       if (ctx.signature === undefined || ctx.blob === undefined) {
         return ctx.accept()
       }

       if (!key.verify(ctx.blob, ctx.signature, ctx.hashAlgo)) {
         return fail("bad signature")
       }

+      authenticatedFingerprint = fingerprint
+      const connections =
+        this.connectionsByFingerprint.get(fingerprint) ?? new Set()
+      connections.add(conn)
+      this.connectionsByFingerprint.set(fingerprint, connections)
+
       authenticated = true
       ctx.accept()
     })

     conn.on("ready", () => {
       conn.on("tcpip", (accept, reject, info) => {
+        if (
+          !authenticatedFingerprint ||
+          !this.opts.isAuthorized(authenticatedFingerprint)
+        ) {
+          return reject()
+        }
+
         // Existing target validation...
       })
     })

     conn.on("close", () => {
+      if (authenticatedFingerprint) {
+        const connections =
+          this.connectionsByFingerprint.get(authenticatedFingerprint)
+        connections?.delete(conn)
+        if (connections?.size === 0) {
+          this.connectionsByFingerprint.delete(authenticatedFingerprint)
+        }
+      }
       // Existing cleanup...
     })
   }
 }
```

sdk/packages/simplex/src/services/tunnel/EmbeddedSshServer.ts
```diff
@@ -0,0 +1,190 @@
...
+			const fail = (reason: string) => {
```
ddboy19912, line 294: Non-public-key attempts bypass `fail()`, so password or keyboard-interactive attempts do not count toward either advertised failure limit. 

We shoud route all rejected authentication methods through the same accounting path.

Potential Fix:
```
- if (ctx.method !== "publickey") return ctx.reject(["publickey"])
+ if (ctx.method !== "publickey") {
+   return fail(`unsupported authentication method: ${ctx.method}`)
+ }
```

sdk/packages/simplex/src/services/tunnel/EmbeddedSshServer.ts
```diff
@@ -0,0 +1,190 @@
...
+ * reach this code, so it has to be that narrow.
+ */
+export class EmbeddedSshServer {
+	private readonly server: SshServerType
+	private readonly logger = getLogger("tunnel")
+	private readonly authTimeoutMs: number
+	private readonly maxAuthFailures: number
+	private readonly maxFailuresPerSource: number
+	private readonly sourceWindowMs: number
+	private readonly failuresBySource = new Map<string, number[]>()
```
ddboy19912, line 115: Expired entries in `failuresBySource` are only removed when the same source reconnects, so one-off scanner addresses can remain indefinitely. Please add periodic pruning and a maximum size for `failuresBySource` to prevent unbounded memory growth.

Potential Fix:
```
+const MAX_TRACKED_SOURCES = 10_000
+const FAILURE_SWEEP_INTERVAL_MS = 60_000

 export class EmbeddedSshServer {
+  private lastFailureSweep = 0

   private recordFailure(ip: string): void {
+    const now = Date.now()
+
+    if (now - this.lastFailureSweep >= FAILURE_SWEEP_INTERVAL_MS) {
+      this.pruneFailures(now)
+      this.lastFailureSweep = now
+    }
+
+    if (
+      !this.failuresBySource.has(ip) &&
+      this.failuresBySource.size >= MAX_TRACKED_SOURCES
+    ) {
+      const oldest = this.failuresBySource.keys().next().value
+      if (oldest !== undefined) this.failuresBySource.delete(oldest)
+    }
+
     const hits = this.failuresBySource.get(ip) ?? []
-    hits.push(Date.now())
+    hits.push(now)
+
+    // Reinsert so Map order represents recent activity.
+    this.failuresBySource.delete(ip)
     this.failuresBySource.set(ip, hits)
   }

+  private pruneFailures(now: number): void {
+    const cutoff = now - this.sourceWindowMs
+
+    for (const [ip, hits] of this.failuresBySource) {
+      const active = hits.filter((timestamp) => timestamp > cutoff)
+      if (active.length === 0) this.failuresBySource.delete(ip)
+      else this.failuresBySource.set(ip, active)
+    }
+  }
 }
```

sdk/packages/simplex/src/services/tunnel/keys.ts
```diff
@@ -0,0 +1,215 @@
...
+
+	removeDevice(fingerprint: string): boolean {
+		const before = this.devices()
+		const kept = before.filter((d) => d.fingerprint !== fingerprint)
+		if (kept.length === before.length) return false
+		this.writePrivate("authorized_keys", kept.length ? `${kept.map(deviceLine).join("\n")}\n` : "")
+		return true
+	}
+
+	knownRelay(): KnownRelay | undefined {
```
ddboy19912: `known_relay` stores only one relay fingerprint. Switching from A to B overwrites A’s pin, so switching back to A trusts its next key again. We should store fingerprints per relay address, or require an explicit `relayHostKey` when changing relays.

Potential Fix:
```
- knownRelay(): KnownRelay | undefined {
+ knownRelay(relay: string): KnownRelay | undefined {
   const path = join(this.dir, "known_relay")
   if (!existsSync(path)) return undefined
-  const [relay, fingerprint] =
-    readFileSync(path, "utf8").trim().split(/\s+/)
-  return relay && fingerprint ? { relay, fingerprint } : undefined
+
+  for (const line of readFileSync(path, "utf8").split("\n")) {
+    const [storedRelay, fingerprint] = line.trim().split(/\s+/)
+    if (storedRelay === relay && fingerprint) {
+      return { relay: storedRelay, fingerprint }
+    }
+  }
+
+  return undefined
 }

 rememberRelay(relay: string, fingerprint: string): void {
-  this.writePrivate("known_relay", `${relay} ${fingerprint}\n`)
+  const path = join(this.dir, "known_relay")
+  const entries = existsSync(path)
+    ? readFileSync(path, "utf8")
+        .split("\n")
+        .filter(Boolean)
+        .filter((line) => line.split(/\s+/)[0] !== relay)
+    : []
+
+  entries.push(`${relay} ${fingerprint}`)
+  this.writePrivate("known_relay", `${entries.join("\n")}\n`)
 }
```

Usage:
```
- this.keys.knownRelay()
+ this.keys.knownRelay(this.relay)
```

sdk/packages/simplex/src/bin/simplex.ts
seunlanlege: @/tmp/claude-1001/-home-seun-Projects-hyperbridge--claude-worktrees-simplex-port-tunnel-e38419/b5d17bae-74cf-45a5-a232-d3a3a03bb8f9/scratchpad/r1.md

sdk/packages/simplex/src/bin/simplex.ts
seunlanlege: @/tmp/claude-1001/-home-seun-Projects-hyperbridge--claude-worktrees-simplex-port-tunnel-e38419/b5d17bae-74cf-45a5-a232-d3a3a03bb8f9/scratchpad/r2.md

sdk/packages/simplex/src/services/tunnel/EmbeddedSshServer.ts
seunlanlege: @/tmp/claude-1001/-home-seun-Projects-hyperbridge--claude-worktrees-simplex-port-tunnel-e38419/b5d17bae-74cf-45a5-a232-d3a3a03bb8f9/scratchpad/r3.md

sdk/packages/simplex/src/services/tunnel/EmbeddedSshServer.ts
seunlanlege: @/tmp/claude-1001/-home-seun-Projects-hyperbridge--claude-worktrees-simplex-port-tunnel-e38419/b5d17bae-74cf-45a5-a232-d3a3a03bb8f9/scratchpad/r4.md

sdk/packages/simplex/src/services/tunnel/EmbeddedSshServer.ts
seunlanlege: @/tmp/claude-1001/-home-seun-Projects-hyperbridge--claude-worktrees-simplex-port-tunnel-e38419/b5d17bae-74cf-45a5-a232-d3a3a03bb8f9/scratchpad/r5.md

sdk/packages/simplex/src/services/tunnel/keys.ts
seunlanlege: @/tmp/claude-1001/-home-seun-Projects-hyperbridge--claude-worktrees-simplex-port-tunnel-e38419/b5d17bae-74cf-45a5-a232-d3a3a03bb8f9/scratchpad/r6.md

seunlanlege:
@/tmp/claude-1001/-home-seun-Projects-hyperbridge--claude-worktrees-simplex-port-tunnel-e38419/b5d17bae-74cf-45a5-a232-d3a3a03bb8f9/scratchpad/summary.md

ddboy19912 (approved):

Wizdave97:
No major issues outside these

```
**High**

1. **A paired device is a full operator** (PR 1217, design). Tunnel requests reach the UI on loopback and are indistinguishable from the operator's browser. A device holder can pair new keys (persisting past revocation), re-point the relay with trust-on-first-use, stop the filler, and send funds. Fix: tag tunnel-originated connections and refuse the tunnel-admin and money routes for them.

**Medium**

2. **Silent connections are never timed out by simplex** (PR 1217, confirmed at runtime). ssh2 fires the connection callback only after the client banner, and the auth timer starts inside that callback. A forwarded stream that sends nothing is held forever and never counted. Fix: arm the deadline in `inject()`.

3. **Idle phone connections are never timed out by the relay** (relay, PoC). No first-byte deadline, and the per-source cap keys on /64. Together with finding 2, 32 silent connections from eight sources lock an operator out of their port permanently. Fix: first-byte deadline plus a per-source connection rate limit on operator ports.

4. **Pre-auth slots outlive the handshake timeout, and the hard global cap makes lockout cheap** (relay, PoC). The banner read happens before the watchdog is spawned, so a silent socket holds a slot for 75 s. 512 such sockets refuse every operator connection and reconnection. Fix: wrap the handshake in the timeout, and replace the hard cap with random early drop or raise it.
```

Wizdave97:
7. High (design): a paired device is indistinguishable from the operator and holds every privilege
The embedded server dials the UI on 127.0.0.1 (EmbeddedSshServer.ts:192), so a request arriving through the tunnel is, to UiServer.handle, the same as one from the operator's browser. Its only guards are the Host header and the constant X-Simplex-UI: 1 header, both of which any client sets. A device holder therefore has:

POST /api/tunnel/devices: pair an additional key. A lost unlocked phone, or malware with the device key, gains persistence that survives revoking the lost device. The PR's "hangs up on a live session when its device is revoked" is defeated in one request.
PUT /api/tunnel: switch the relay to one the attacker runs. With no relayHostKey configured for the new address the pin is trust-on-first-use, and the change is persisted to the config file.
POST /api/send, /api/stop, /api/config, /api/vault/*, /api/log-level: move funds, halt the filler, rewrite strategy config.
None of this is an auth bypass, but it is the largest gap against the PR's own threat model, which treats the phone as the thing most likely to be lost.

Fix: tag tunnel-originated connections and give them a reduced route set. The cleanest tagging is in-process. Both servers live in the same Node process, so hand the SSH channel straight to the HTTP server with httpServer.emit("connection", channel) after setting a viaTunnel marker on it, instead of dialing loopback. UiServer.handle can then read req.socket.viaTunnel and refuse /api/tunnel* outright for tunnel requests, and, as a product decision, the money and lifecycle routes as well. This also removes the loopback TCP hop. An alternative is dialing from a distinct loopback source (localAddress: "127.0.0.2"), but that only works on Linux without extra setup. A per-device second factor on money routes is the other option if remote sends must stay.

PR #1222, opened 2026-09-08
[tesseract]: Deliver modules named in the module filter alongside the incentivized ones

dharjeezy:
The outbound task drops hyperbridge-originated requests whose source module has no `OutboundRequestDeliveryReward` registered, so a module with no reward configured is never delivered at all. `pallet-hyper-fungible-token` sits in that position, which leaves its transfers out of hyperbridge undeliverable.

`module_filter` could not be used to unblock it either, since a non empty filter is exclusive and naming one module there silences every other module the relayer was being paid for.

The two lists are now additive: a request is delivered when its source module appears in the reward allowlist or in the operator's `module_filter`, and an empty filter keeps the current behaviour. Claims are still only filed for modules on the reward allowlist, since the pallet rejects the rest with `OutboundRequestNoRewardConfigured`.

### DIFF
--- a/tesseract/messaging/messaging/src/outbound.rs
+++ b/tesseract/messaging/messaging/src/outbound.rs
@@ -347,15 +347,7 @@ async fn submit_for_dest(
 	// so the filter keeps every event routed to this destination.
 	let mut events = events
 		.into_iter()
-		.filter(|ev| {
-			filter_events(
-				&relayer_config,
-				incentivized.as_deref(),
-				dest_state_machine,
-				coprocessor,
-				ev,
-			)
-		})
+		.filter(|ev| filter_events(&relayer_config, dest_state_machine, coprocessor, ev))
 		.collect::<Vec<_>>();
 
 	retain_incentivized_requests(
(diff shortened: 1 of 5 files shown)


Wizdave97:
Bump tesseract

tesseract/messaging/messaging/src/outbound.rs
```diff
@@ -533,18 +548,29 @@ async fn submit_for_dest(
...
+/// everything as a no-op fallback.
+///
+/// The reward allowlist decides what the relayer gets *paid* for; the
+/// `module_filter` decides what the operator is *willing* to deliver. A module
+/// on neither list is dropped, which is what kept hyperbridge-originated
+/// transfers from `pallet-hyper-fungible-token` from being relayed at all
+/// while no reward was configured for it.
 fn retain_incentivized_requests(
 	events: &mut Vec<Event>,
+	config: &RelayerConfig,
```
Wizdave97, line 555: This is the only place we should make the change, the others were fine.

Wizdave97 (approved):

PR #1234, opened 2026-09-09
[simplex]: move operator state into bids.db and retire runtime-state.json

seunlanlege:
Simplex keeps two pieces of operator state across restarts:

- `paused` — whether the operator has stopped the filler
- `phantomBids` — the live phantom bid on each chain, so the next run can retract it and get the 0.01 BRIDGE deposit back

Both lived in a `runtime-state.json` file next to `bids.db` and `activity.db`. This PR moves them into `bids.db` and deletes the file.

## Why

The file was never a deliberate choice. It arrived with pause/resume back in July, before the `SimplexDataStore` interface existed, and nobody revisited it when simplex became a library.

It had two bugs.

**A crash mid-write lost everything.** `writeFileSync` truncates the file before writing it. If the process died in between, the file no longer parsed, and the reader fell back to an empty record. A paused filler came back up unpaused, and the phantom deposit was stranded — the two failures the file existed to prevent, in the exact situation it was meant to cover.

**Two writers could overwrite each other.** Saving one key meant reading the whole record, merging, and writing it all back. The phantom batch and an operator pause could interleave and drop one another's key.

SQLite gives us both properties for free, and the process already has `bids.db` open.

## What changed

- `SqliteStateStore` keeps one row per key in `bids.db`, values JSON-encoded. Adding a key later needs no schema change.
- `StateStore` gains an optional `patch`. The SQLite store implements it as an upsert of just the named keys, in a single transaction.
- `patchRuntimeState` uses `patch` when the store has one, and keeps the old read-merge-write for stores that don't — which is what `MemoryDataStore` uses.

`patch` is optional because `StateStore` is public API. Someone backing it with Postgres or an HTTP service shouldn't have to implement two writers just to keep compiling.

Error handling keeps its old intent. Writes log instead of throwing, so a pause that can't be saved still pauses the filler. Reads now throw, so an unreadable database fails boot rather than quietly starting a filler the operator had paused.

## Upgrading

The first time a filler starts on this version, the store looks for the old file — first in the data directory, then in `.filler-data/` — imports whatever it finds, and deletes both copies.

It only does this when the table is empty, so it can't overwrite state the database already holds. Deleting both copies matters too: otherwise a fresh database would re-import a stale file and re-apply a pause the operator had already lifted.

One consequence worth knowing: **downgrading to 0.16.x loses the state and starts unpaused.** The version that reads the file is the one that deletes it.

## Breaking

`FileStateStore` is no longer exported from `@hyperbridge/simplex/sqlite`, and the on-disk state format changed. No version bump in this PR, so whichever release picks it up owes at least a minor.

## Testing

New tests cover the round trip, reopening a data directory, two concurrent patches, deleting a key, whole-record `set`, writing to a closed database, and all three migration cases.

The concurrent-patch test is a real regression test — it fails if you disable `patch` and let the old read-merge-write run.

CI is green. The first run hit an unrelated failure in `fx.testnet.test.ts`, where a Hyperbridge extrinsic couldn't replace an identical one already in the pool. It passed on re-run with no code change.

### DIFF
--- a/docs/content/developers/sdk/api/simplex.mdx
+++ b/docs/content/developers/sdk/api/simplex.mdx
@@ -668,9 +668,16 @@ else, so an implementation may prune, sample or cap history freely.
 ```typescript lineNumbers
 get(): Promise<RuntimeState>
 set(state: RuntimeState): Promise<void>
+patch?(patch: Partial<RuntimeState>): Promise<RuntimeState>
 ```
 
-`RuntimeState` is `{ paused?: boolean }`.
+`RuntimeState` is `{ paused?: boolean, phantomBids?: Record<string, string> }`.
+
+`set` replaces the whole record. `patch` is optional and merges, leaving keys it does not name
+untouched — implement it if your backend can do that atomically. The fallback when you do not is a
+read, a merge and a write back, which loses a key another writer set in between. That race is real
+on a running filler: `paused` is written by the operator and `phantomBids` by the bid batch, from
+different places at unrelated times.
 
 ### Implementations
 
@@ -679,8 +686,10 @@ set(state: RuntimeState): Promise<void>
 | `MemoryDataStore` | `@hyperbridge/simplex` | Default. Process-local, row-capped, lost on exit |
 | `SqliteDataStore` | `@hyperbridge/simplex/sqlite` | Durable. Built on `node:sqlite`, so nothing to install |
 
-`SqliteDataStore(dataDir)` keeps `bids.db`, `activity.db` and `runtime-state.json` in the given
-directory, so a data directory written by the binary is picked up unchanged. It is built on Node's
+`SqliteDataStore(dataDir)` keeps `bids.db` and `activity.db` in the given directory, so a data
+directory written by the binary is picked up unchanged. Operator state rides in `bids.db` beside
+the bids, and implements `patch`; a `runtime-state.json` from an earlier version is imported once
+and then removed. It is built on Node's
 built-in `node:sqlite`, so there is no native module to install or compile — which is why the
 package requires Node 22.16 or newer.
 
--- a/sdk/packages/simplex/docs/ai/changelog/2026-09-10-operator-state-moved-into-bids-db-runtime-state-json-retired.md
+++ b/sdk/packages/simplex/docs/ai/changelog/2026-09-10-operator-state-moved-into-bids-db-runtime-state-json-retired.md
@@ -1,0 +1,34 @@
+# Operator state moved into `bids.db`; `runtime-state.json` retired
+
+`FileStateStore` is now `SqliteStateStore`, backed by a `runtime_state` table in `bids.db` — the
+database `SqliteBidStore` already holds open — with one row per `RuntimeState` key and the value
+JSON-encoded. `SqliteDataStore` passes it the bids connection and the data directory.
+
+Two bugs go with the file. Its `writeFileSync` truncated before writing, so a crash mid-write left
+JSON that would not parse and `get()` fell back to `{}`, silently resuming a paused filler and
+stranding the phantom deposit the file existed to reclaim. And every writer went through
+`patchRuntimeState`'s read-merge-write, so the phantom batch and an operator pause could drop one
+another's key. `StateStore` gained an optional `patch`, which the SQLite store implements as an
+upsert of only the named keys inside one transaction; `patchRuntimeState` uses it when present and
+otherwise keeps the old get-then-set, which is what `MemoryDataStore` gets. A key set to
+`undefined` deletes its row. `node:sqlite` has no transaction helper, so `write` brackets its
+statements with `BEGIN`/`COMMIT` and rolls back on a throw.
+
+Existing data directories are migrated on open: if `runtime_state` is empty the store reads
+`runtime-state.json` from the data directory, then from `.filler-data/` relative to cwd, writes
+what it finds, and unlinks both copies. Guarding on an empty table means it can never overwrite
+state the database already holds; deleting both copies means an empty database cannot resurrect a
+pause the operator has since lifted.
+
+Writes still swallow their errors (logged now, which the file store could not do) so a pause that
+cannot be persisted still pauses the filler; reads no longer do, so an unreadable database fails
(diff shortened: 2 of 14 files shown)


royvardhan:
The migration deletes the legacy state file even when it failed to read or import it:

```ts
try {
    state = JSON.parse(readFileSync(path, "utf-8")) as RuntimeState
} catch {
    continue
}

// Later, even if nothing was imported:
for (const path of paths) {
    try {
        unlinkSync(path)
    } catch {}
}
```

I reproduced this with a valid file containing `paused: true` and phantom-bid records, but without read permission. Startup succeeded with empty state and deleted the file, making those records unrecoverable.

Could we distinguish missing files from read/parse failures, and preserve the legacy files unless migration succeeds?

[Code affected: state.ts:60–75](https://github.com/polytope-labs/hyperbridge/blob/9f8a27ba6a67ff074b17c6f9f037ee9172a0dc07/sdk/packages/simplex/src/data/sqlite/state.ts#L60)

seunlanlege:
Good catch, and thank you for reproducing it — that's a real data-loss bug and your reading of it is exactly right. Fixed in 8d01c9a.

The delete loop now runs over the files that were actually read back, not over both candidate paths:

- **Missing file** — ignored silently. That's the normal case and says nothing.
- **Any other read failure** (no permission, I/O error, malformed JSON) — logged at warn, and the file stays where it is. It may still hold the pause, which is the state the import exists to rescue.
- **Read and parsed** — imported if it's the first one, and deleted either way.

I also treat a parsed value that isn't a plain object (`null`, an array, a bare string) as a failed read, for the same reason.

Deleting every *readable* copy rather than only the imported one is deliberate: a copy left behind gets re-imported by the next empty database and re-applies a pause the operator has since lifted.

Three tests, each of which fails against the previous behaviour: an unparseable file, a file holding a non-object, and an unreadable one. The unreadable case skips when running as root, since root reads a mode-000 file regardless.

seunlanlege:
Please check again @royvardhan

Wizdave97:
Three findings:

- **[[src/data/sqlite/state.ts:141](https://claude.ai/epitaxy/sdk/packages/simplex/src/data/sqlite/state.ts:141)](sdk/packages/simplex/src/data/sqlite/state.ts:141)** — `patch()` guards the write with `persist()` but then calls the unguarded `this.read()`, so a database failure still rejects. This is the *only* path production uses (`Simplex.pause`, `bin/simplex.ts:177` → `patchRuntimeState` → `store.patch`); nothing ever calls `set()`. On an unreadable `bids.db` — closed store, SQLITE_IOERR, or SQLITE_BUSY from a second `SqliteDataStore` (bids.db has no WAL) — `UiServer` has already called `filler.pause()` at line 461, then line 462 rejects and the handler returns a 500. The operator is told the pause failed while the filler is actually paused, which is exactly what `persist()`'s doc comment promises won't happen. The test that looks like it covers this exercises `set()`, so it passes regardless.
- **[[docs/content/developers/sdk/api/simplex.mdx:682](https://claude.ai/epitaxy/docs/content/developers/sdk/api/simplex.mdx:682)](docs/content/developers/sdk/api/simplex.mdx:682)** — public API docs still say `SqliteDataStore` keeps `runtime-state.json`, and the `StateStore` snippet at :668 omits the new optional `patch`. The PR's own rationale for making `patch` optional is third-party implementers — who read this page and will ship a store that silently keeps the read-merge-write race. `docs/ai/` was updated; this wasn't.
- **[[src/data/sqlite/state.ts:57](https://claude.ai/epitaxy/sdk/packages/simplex/src/data/sqlite/state.ts:57)](sdk/packages/simplex/src/data/sqlite/state.ts:57)** — the two candidate paths aren't de-duplicated. With `--data-dir .filler-data`, one file lands in `recovered` twice and is unlinked twice; the second throws ENOENT and logs "Could not delete the retired state file", telling the operator a stale file survived when it didn't.

Things I checked and found fine: the empty-table migration guard, whole-record `set` semantics, `undefined`-deletes-a-row, the `ENOENT`-vs-everything-else split in the read error handler, no lingering `FileStateStore` importers, and `boot.ts:649` letting `get()` throw (deliberate, and it unwinds cleanly).

The concurrent-patch test is a real regression test, as the description claims — worth keeping.

The worktree is left on the `pr-1234` branch.

seunlanlege:
All three confirmed and fixed in d0e080a. Thanks — the first one is the serious one and I'd have missed it.

**`patch()` reading outside the guard.** Correct, and correct about why the existing test didn't catch it: it exercises `set()`, which never reads. Production only ever reaches the store through `patch` — `Simplex.pause/resume`, `bin/simplex.ts` `setPaused`, and the phantom batch all route through `patchRuntimeState`, and nothing outside tests calls `set()`.

The write and the read-back now share one guard, and on failure `patch` returns the requested patch rather than throwing. With the database unreadable that is the most that can honestly be said about the state; the failure is already logged, and no production caller reads the return value. `persist` is generic over its return type to allow it.

One small correction to the trace, which doesn't change the finding: the handler is at `UiServer.ts:777`/`:779`, not 461/462 — `filler.pause()` then `await setPaused(pause)`. Same mechanism, same 500.

New test: `patch` on a closed database resolves instead of rejecting. It fails against the previous code.

**Duplicate candidate paths.** Confirmed — `join(".filler-data", …)` gives a relative string and `resolve(cwd, ".filler-data", …)` an absolute one, so the same file lands in `recovered` twice and the second `unlinkSync` throws ENOENT. The misleading log line is the real damage. Both paths are now `resolve`d and de-duplicated through a `Set`, with a test that asserts no "could not delete" warning when the data directory *is* the legacy directory.

**Stale API docs.** Fair, and the reasoning is right — that page is what third-party implementers read, and they're the whole reason `patch` is optional. `docs/content/developers/sdk/api/simplex.mdx` now shows `patch?` in the `StateStore` snippet, explains when to implement it and what the fallback costs, and no longer claims `SqliteDataStore` keeps `runtime-state.json`. `RuntimeState` was also still documented as `{ paused?: boolean }`, from before `phantomBids` existed; that's updated too.

43 tests pass across `src/tests/data`, five runs clean.

royvardhan (approved):

Wizdave97 (approved):

PR #1246, opened 2026-09-09
[simplex]: Read the filler's logs from the dashboard, with level pills and search

seunlanlege:
The filler's logs were only readable on the console of whatever terminal or container it was started from. An operator on the dashboard had no way to see why a fill was skipped. This adds a Logs page to the operator sidebar: everything the filler has logged since launch, searchable, with a level control that decides what it records.

## Where the records live

`LogStore` is fed by registering its `sink()` on a `LoggerContext`, exactly as the console sink is. It keeps two tiers, because "all of it" and "fast" pull in opposite directions:

- **The file is the history.** Every record is appended to `<dataDir>/logs/simplex-<launch>.log` as NDJSON, bounded only by the disk. Older launches are pruned, keeping five.
- **The ring is a cache in front of it.** The last 5000 records are also held in memory. That is what the live stream broadcasts, and what answers a query outright as long as nothing has been evicted — the common case for a filler at `info`.

When a query does reach past the ring's oldest record, the older half is read back off the file. Splitting at that seq also sidesteps the write stream's buffer: the records not yet flushed to disk are exactly the ones the ring still holds, so the two halves cannot miss one between them.

The store is constructed and registered at module load, before anything can log, and handed to `Simplex.start` inside a fan-out sink rather than attached afterwards — `start()` builds the filler's context and runs all of boot inside itself, so a sink added after it resolves misses every boot record. `openLaunchFile` runs once `--data-dir` is parsed and writes out whatever the ring already holds, so the file genuinely begins at launch.

Two routes: `GET /api/logs` (backfill, plus the level, the ring capacity, how many records have been captured, and where the history is being written) and `GET /api/logs/stream` (SSE: replay from `after`, then live). The stream subscribes *before* awaiting the replay, since reading history can hit the file, and orders the two sources by seq on the way out.

## The level control

The pills set the filler's log level, in both directions. Raising records more; lowering stops recording it at all, which is how an operator keeps a long-running filler from writing a log file they do not want. The PUT is awaited rather than optimistic, because changing the level re-runs the feed and its GET reads the level back out of the config — fired concurrently, that read can answer before the write lands and show the old level back.

`setLogLevel` now moves the process-wide context as well as the filler's. The page merges records from both and reports one level for the result, so leaving the process context pinned at its default made that a lie in both directions.

## Desktop only

A log line is a wide, dense, monospace record, and reading logs means scanning and comparing them — the thing a 390px column is worst at. It is also the only page holding an open stream and thousands of rows, which is real battery and memory on a device that opened the dashboard to check a balance.

The gate is `(max-width: 600px), (pointer: coarse) and (max-height: 600px)`. Width alone was wrong in both directions: a phone in landscape is 700-950px wide and read as a desktop. The guard is in the component rather than in CSS, so the page never mounts and no stream is opened even for the frame before the redirect.

## Fixes from an adversarial audit

The first revision of this branch was audited across ten dimensions with every finding judged by a three-lens panel — 43 raised, 26 upheld. The ones that mattered:

- **`parseRecord` dropped any record whose payload carried a `level` field.** pino writes its own numeric level first and the caller's merge object verbatim after, so `warn({ level }, …)` emits two `level` keys and `JSON.parse` keeps the caller's. That silently swallowed `handleLogLevel`'s own "Log level changed from the UI" — the one line recording the change, missing from the page it was logged for. The numeric level now comes from the line prefix, and the payload's value is kept in `detail`.
- **Eviction was `Array.splice`,** which shifts the whole buffer on every record once full. Measured 9.89µs/record versus 1.43µs for a head-index ring: ~86% of the per-record cost, on the filler's synchronous logging path. Now O(1).
- **`detail.slice()` pinned the whole parent string.** V8 returns a SlicedString, so the 4000-char cap bounded what the page showed but not what the process held. Forced to a flat copy.
- **A stale `after` emptied the page permanently.** seq restarts at 1 each launch, so Clear's mark from a previous process filtered out everything. Now ignored when it exceeds this launch's counter, on both sides.
- **Dropped SSE frames were silent.** The 1MB backlog guard is still right — a reader that has stopped reading must not buffer the firehose inside the filler — but it now emits `event: gap` on recovery so the client re-reads instead of splicing a hole shut.
- **The sixth sidebar item clipped the footer** on a short viewport; the sidebar can now shrink and its nav scrolls.
- **`LogRow` is memoized** (13-22ms per flush at 2000 rows, versus 2.6ms), and sticky-scroll re-anchors whenever the list is replaced wholesale.

## On exposure

Log records can carry provider URLs with API keys: `compactError` records the failing endpoint, and `QuorumError` messages embed each provider. This is not new exposure — the pre-existing `GET /api/chains` already returns `rpcUrls` and `bundlerUrl` unmasked on the same listener, with a comment saying it does so deliberately because the chain editor round-trips them. Every reader of `/api/logs` is already a reader of `/api/chains`. (An earlier revision of this description claimed no log line carries a secret-bearing field; that was wrong, though the conclusion stands.)

## Testing

- `src/tests/log-store.test.ts` (29): capture through a real `LoggerContext`, the `level` collision, ring eviction, the query predicates, subscriptions, and persistence end to end — writing, reading history back after eviction, filtering it, backfilling on a late `openLaunchFile`, pruning launches, and falling back to memory when the directory is unusable.
- The `logs` block in `src/tests/ui-server.test.ts` (14): both routes and their filters, the limit cap, method and availability gates, the `Content-Type`, the stale-`after` case, records logged during the replay, bidirectional `PUT /api/log-level` with persistence, and that `stop()` ends open tails. Its SSE helper is now event-driven rather than deadline-based, which also made the file faster.
- 132 tests pass across `log-store`, `ui-server`, `logger` and `setup-api`.
- Driven in Chrome against a live `UiServer`: history reaching back past the ring (264 captured, 200 in memory, all 264 returned contiguously from seq 1), lowering the level cutting recording from ~13 records/12s to 2, raising it back, the landscape-phone gate with zero `/api/logs` requests, a tablet in landscape correctly keeping the page, and the sidebar at a 380px viewport.

Version bumped to 0.17.0.


Wizdave97:
Four findings:

- **[[src/services/server/UiServer.ts:1521](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/server/UiServer.ts:1521)](sdk/packages/simplex/src/services/server/UiServer.ts:1521)** — the replay writes up to 2000 records in one synchronous loop with no drain, and `res.writableLength` is `outputSize + socket.writableLength`, so it accumulates across the whole loop. With `detail` capped at 4000 chars, 2000 records is routinely 1–2MB — past the 1MB guard at :1488. A *healthy* operator on the remote-access tunnel gets `dropped = true`, then `event: gap`; [[Logs.tsx:194](https://claude.ai/epitaxy/sdk/packages/simplex/ui/src/operator/Logs.tsx:194)](sdk/packages/simplex/ui/src/operator/Logs.tsx:194) responds with `setAttempt(n => n+1)`, which re-runs the feed effect, replays the same 2000 records, and trips the guard again. The page never settles, and each cycle re-triggers the file scan below. The guard is the right idea — it just needs to await drain (or chunk across ticks) so it fires on a reader that stopped reading rather than one that was never given a chance.
- **[[src/services/server/LogStore.ts:197](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/server/LogStore.ts:197)](sdk/packages/simplex/src/services/server/LogStore.ts:197)** — `needsHistory` is decided before `fromRing` is measured. On any filler whose ring has wrapped, the unresumed initial GET falls through to `scanFile`, which parses every line below `oldestInRing` — on a day at `debug`, several GB — and then `slice(-limit)` throws all of it away because the ring alone already filled the 2000-record page. The 250ms-debounced search box repeats it per keystroke. `if (fromRing.length >= limit) return fromRing.slice(-limit)` before the scan costs nothing in the cases where history is genuinely needed.
- **[[src/services/server/LogStore.ts:279](https://claude.ai/epitaxy/sdk/packages/simplex/src/services/server/LogStore.ts:279)](sdk/packages/simplex/src/services/server/LogStore.ts:279)** — the launch stamp has second resolution and the stream opens `flags: "a"`. A sub-second restart (systemd `RestartSec=0`) appends a fresh seq-1..N run after the dead launch's, and pruning can't help because the same-named file sorts newest. `scanFile` then breaks on the old launch's records and returns them as this launch's history; the UI keys rows by seq, so they collide. Milliseconds in the stamp, or `flags: "wx"` with a fallback, fixes it.
- **[[ui/src/operator/Operator.tsx:83](https://claude.ai/epitaxy/sdk/packages/simplex/ui/src/operator/Operator.tsx:83)](sdk/packages/simplex/ui/src/operator/Operator.tsx:83)** — `const mobile = useIsMobile()` is never read; `handheld` drives everything. Leaves a live `matchMedia` subscription re-rendering the shell for nothing. Biome only lints `src/`, so CI won't catch it.

seunlanlege:
Thanks — all four were real. Fixed in c5541db0 (rebased onto the docs split from #1253).

**The backpressure guard.** You're right that it fired on a healthy reader, and it was worse than a false positive: the `gap` it emitted made the page re-run the feed, which replayed the same page and tripped it again. A livelock, not a hiccup. Awaiting drain is the fix, so every frame now leaves through one queue drained by an async pump that awaits `drain`, and the bound is a record count above `MAX_LOG_PAGE` rather than a byte threshold — a replay can no longer reach it, so the guard means what it says. Regression test asserts a 1200-record, ~2.4MB replay reports no gap; I checked it fails against the old guard rather than passing vacuously.

**The wasted scan.** Exactly as described — `slice(-limit)` threw away everything the scan had parsed. `if (fromRing.length >= limit) return fromRing.slice(-limit)` now runs before `needsHistory` is even computed, since the scan is the expensive half.

**Same-second launches.** Took both halves of your suggestion: the stamp carries milliseconds *and* the stream opens `wx`, so a collision is refused rather than appended into. One thing your note surfaced that I hadn't seen — the error handler cleared `stream` but left `file` set, so a store that fell back to memory still advertised a path in the DTO for a file nothing was writing. That's now cleared together. I watched it happen live, incidentally: this box's disk hit 100% mid-session, the write stream took an ENOSPC, and the dashboard footer flipped to "memory only" while still reporting the path.

**The dead `useIsMobile()`.** Gone. And you're right that CI wouldn't have caught it — `lint` only covers `src/`. Worth pointing `biome check ui/src` at CI separately; happy to do that here or leave it for its own change.

Also in this push, unrelated to the review: the Logs page now takes a 1.25rem gutter instead of the shell's up-to-4rem page padding, which moves ~90px from margin into message text at 1440.

135 tests pass.

Wizdave97 (approved):

PR #1247, opened 2026-09-09
[simplex]: Replace better-sqlite3 with node:sqlite

### LINKED ISSUE #1236: [simplex] Replace better-sqlite3 with node:sqlite
Part of #1235

## Why

`better-sqlite3` is C++ compiled to a `.node` binary. It works today because `npm install` compiles it on each user's machine. A desktop app has no compile step, so every supported platform needs a prebuilt copy. Under the model settled in #1237 the solver runs in a bundled Node binary, not Electron, so there is no Electron ABI to rebuild against. The per-platform prebuild matrix is what remains, and removing the native module is what deletes it.

Node ships SQLite in the runtime as `node:sqlite`. Removing the native dependency is the single highest-leverage change for every packaging option, and it pays off before any desktop work ships.

Verified locally on Node v24.12.0: `node:sqlite` is available without a flag (exports `DatabaseSync`, `StatementSync`, `Session`, `constants`, `backup`), emitting only an `ExperimentalWarning`.

## Immediate wins, independent of desktop

- `sdk/packages/simplex/scripts/Dockerfile` installs `python3 make g++` solely as the node-gyp fallback. Delete them; the builder image shrinks and the build gets faster.
- `optionalDependencies` currently pins `@types/better-sqlite3: ^7.6.13` while `devDependencies` pins `^9.6.0`. Both disappear.
- `openDataStore()` in `src/bin/simplex.ts` exists mostly to turn a failed native build into a readable error. That whole failure mode goes away.

## Surface area

Small and fully enumerated. Counting call sites under `src/data/sqlite/`:

| Call | Count | Notes |
|---|---|---|
| `.all(` | 21 | direct equivalent |
| `.run(` | 10 | direct equivalent |
| `.prepare(` | 10 | direct equivalent |
| `.get(` | 9 | direct equivalent |
| `.exec(` | 5 | direct equivalent |
| `.pragma(` | 1 | **no equivalent** |
| `.transaction(` | 1 | **no equivalent** |

`src/data/sqlite/state.ts` is JSON-on-disk (`runtime-state.json`) and is not affected.

## The two real gaps

**`db.pragma()`** — `src/data/sqlite/bids.ts:65` reads `db.pragma("table_info(bids)")` for its column-add migration. `node:sqlite` has no `.pragma()`. Use `db.prepare("PRAGMA table_info(bids)").all()`, which is exactly what `src/data/sqlite/activity.ts:84` already does. The other pragma is `activityDb.pragma("journal_mode = WAL")` in `src/data/sqlite/index.ts`, which becomes `db.exec("PRAGMA journal_mode = WAL")`.

**`db.transaction(fn)`** — `src/data/sqlite/activity.ts:178` uses better-sqlite3's wrapper. `node:sqlite` has none. Write `BEGIN` / `COMMIT` / `ROLLBACK` explicitly with a `try`/`catch`. One call site.

## Tasks

- [ ] Swap `new Database(path)` for `new DatabaseSync(path)` in `src/data/sqlite/index.ts`.
- [ ] Replace the `import type { Database as DatabaseType } from "better-sqlite3"` type imports in `bids.ts:1` and `activity.ts:1` with `DatabaseSync`.
- [ ] Convert the two pragma call sites as above.
- [ ] Convert `activity.ts:178`'s transaction to explicit BEGIN/COMMIT/ROLLBACK.
- [ ] Audit return-shape differences. `node:sqlite` returns plain objects and uses `BigInt` for INTEGER in some paths; the existing `as any[]` casts around `PRAGMA table_info` will hide breakage, so type them properly while in there.
- [ ] Drop `better-sqlite3` and `@types/better-sqlite3` from `dependencies`, `devDependencies` and `optionalDependencies`.
- [ ] Remove `better-sqlite3` from `ALWAYS_EXTERNAL` in `tsup.config.ts`.
- [ ] Simplify or delete `openDataStore()`'s native-module error path.
- [ ] Remove `python3 make g++` from `scripts/Dockerfile`.
- [ ] Decide the `engines` floor — see below.

## The engines question

`package.json` says `node: >=22`. `node:sqlite` landed in 22.5 behind `--experimental-sqlite` and was unflagged later. Two options:

- Raise the floor to the first version where it is unflagged. Cleanest. The Dockerfile is already on 24.19.0, so the container is unaffected.
- Keep `>=22` and add `--experimental-sqlite` to the shebang. Not a new pattern — the bin already ships `#!/usr/bin/env -S node --enable-source-maps` — but it does not help library consumers who import `@hyperbridge/simplex/sqlite` from their own process.

Recommendation: raise the floor, and say so in the release notes.

## Compatibility requirement

Operators have existing `bids.db` and `activity.db` files, and bid records are how locked deposits are found again for retraction. The new store must open those files unchanged and run the same in-place column migrations.

## Acceptance criteria

- [ ] `pnpm --filter @hyperbridge/simplex test` passes.
- [ ] A data directory written by the current release opens, reads and writes correctly under the new store. Add a fixture test for this, not a manual check.
- [ ] `simplex` starts, records a bid and streams activity with no native module present anywhere in `node_modules`.
- [ ] Docker image builds without a compiler toolchain.
- [ ] `docs/ai/ChangeLog.md` and `docs/ai/Decisions.md` updated.

seunlanlege:
Closes #1236.

Swaps `better-sqlite3` for Node's built-in `node:sqlite`. The package now has no native dependency.

## Why

`better-sqlite3` is a C++ module compiled during `npm install`. That's fine on a dev machine with a compiler. It's a problem anywhere without one — a desktop app has no compile step, so we'd need a prebuilt binary for every platform we support.

Node ships SQLite in the runtime. Using it makes the whole problem go away.

Gone with it: `@types/better-sqlite3`, the `optionalDependencies` block, `python3`/`make`/`g++` from the Docker builder, and three stale config entries.

## The two APIs with no drop-in replacement

**`db.pragma()`** read a table's columns before a migration. Replaced by a small `columnNames()` helper running the same query. One of the three call sites already did it this way by hand.

**`db.transaction(fn)`** has no equivalent, so `attachOrder` now writes `BEGIN` / `COMMIT` / `ROLLBACK` itself.

That transaction earns its keep. `attachOrder` finds rows missing an order summary, updates them, then tells the caller which rows changed. Without one boundary around the read and the write, another writer could fill a row in between. The caller would get back IDs of rows it never touched.

## Node version floor

`engines.node` goes from `>=22` to `>=22.16.0`.

`node:sqlite` stopped needing a flag in 22.13. But two APIs here landed later — `isOpen` in 22.15, `isTransaction` in 22.16. So 22.16 is the first version where everything we call actually exists.

Node 22 is in maintenance LTS until April 2027, so this cuts nobody off. Docker is on 24 either way.

## One bug worth knowing about

tsup rewrites `from "node:sqlite"` to `from "sqlite"` by default.

For `fs` or `path` that's harmless, since both spellings work. `sqlite` has no unprefixed version, so it just fails. Both the CLI and the library entry point were dead on startup with `Cannot find package 'sqlite'`.

Every test passed the whole time. Tests import from source, and nothing rewrites imports there. Only the built bundle was broken.

Fixed by turning the rewrite off. `build.sh` now greps the bundle to catch it coming back, since no test can.

## Not breaking existing installs

Operators already have `bids.db` and `activity.db` on disk. A bid record is how a locked deposit gets found again for retraction. Lose one and you strand money.

`sqlite-compat.test.ts` opens real database files that **better-sqlite3 wrote**, at the old schema, committed under `src/tests/data/fixtures/legacy-v0`.

They're committed as real files deliberately. A database the test created through `node:sqlite` would have been written by the same library reading it back, so it couldn't catch a cross-library problem at all.

The suite checks old rows read back, migrations apply, a won bid's deposit is still findable, WAL survives, and new writes land. Breaking the migration fails 5 of its 11 cases, so it isn't vacuous.

## Second commit: what an audit turned up

I put the change through an adversarial audit afterwards. It found one real bug.

### The busy timeout was silently lost

better-sqlite3 waits 5 seconds for a locked database. `node:sqlite` waits 0. Nothing passed a timeout, so the wait quietly became zero.

| | before | after |
|---|---|---|
| contended write | fails after 28ms | waits 5023ms |

So anything holding the file for a moment would kill the write racing it — an operator running `sqlite3 bids.db`, a backup, a second process on the same data dir. A dropped bid write is a deposit the sweep can't find later.

Now fixed, with a test that holds a lock from a second connection.

The lesson: swapping libraries swaps defaults. The first pass checked that the APIs and return values matched. It never checked constructor defaults.

### Two guards were backwards

`if (!db.isOpen)` and `if (db.isTransaction)` both take the wrong branch when the property doesn't exist.

A missing `isOpen` skips closing every database. A missing `isTransaction` skips the rollback and leaves the connection stuck in an open transaction for good. Both now compare against `false`.

`engines` is advisory — npm and pnpm warn and install anyway — so it can't be trusted to keep old versions out.

### CI wasn't running the new tests

The simplex CI step runs `test:filler`, which names four specific files. None are in `src/tests/data`. The compatibility suite would never have run on a PR.

Added a `test:data` script and a step for it. No network, no secrets, runs in seconds.

### Smaller things

- The CLI started printing Node's SQLite experimental warning on `--help` and `--version`.
- The fixture generator deleted the `.gitignore` that keeps the fixtures committable.
- A docs line still said "Node.js 22 or later".
- A `Decisions.md` note claimed tsup 9 exists. It doesn't.
- A `Flow.md` step described a shutdown path that doesn't happen.
- A CI comment still mentioned the native module.

## How it was checked

- 28/28 storage tests. Full suite matches the pre-change baseline.
- Docker image builds with no compiler in the builder, and contains no `better-sqlite3`.
- `pnpm install --frozen-lockfile` passes.

Then ran the **shipped bundle** inside the container:

- `simplex --version` starts
- a bid is recorded and read back
- activity streams
- a legacy data directory migrates, deposit still reclaimable
- a contended write waits its 5 seconds


royvardhan:
**[P2] Preserve the busy timeout on Node 23**

`engines.node: ">=22.16.0"` includes Node 23, but `DatabaseSync`’s `timeout` option was added in Node 22.16 and 24; Node 23 silently ignores it.

I reproduced this with the PR’s `sqlite-compat.test.ts`: the contended write waits approximately five seconds on Node 22.16.0, but fails after **0.45 ms** on Node 23.11.1. The previous `better-sqlite3` implementation defaults to a five-second wait.

This can lose bid records under brief lock contention. Since the filler persists the record after submitting the bid, the retraction sweep could miss a deposit already locked on-chain.

Could we set `PRAGMA busy_timeout = 5000` explicitly on both connections, or exclude Node 23 from the supported engine range? Adding runtime-matrix coverage would catch this regression.

royvardhan (approved):

Issue #1237, opened 2026-09-09
[simplex] Electron shell: process model, app skeleton, first-run


seunlanlege:
Part of #1235

> **Daemon side is done and merged.** The `@hyperbridge/simplex` changes this issue needed shipped in #1245 (Unix socket listen mode, provenance tagging, stale-socket recovery, path-length guard), #1247 (`node:sqlite`) and #1248 (`--no-open`, `--log-format`), plus follow-ups #1249 and #1250. Checked-off tasks below are already on `main` — **do not redo them**. What remains here is Electron work only.
>
> One finding changed an assumption: Windows named pipes are **not** owner-only. See #1242.

## Decision: detached child process over a Unix domain socket

**Settled.** Electron spawns the existing `simplex` binary detached and lets it go. The app is a viewer that attaches over a `0600` Unix socket (named pipe on Windows). There is no TCP listener by default.

### Why not in-process

| Event | How often | In-process | Detached |
|---|---|---|---|
| Close window | daily | fills (tray) | fills |
| **Quit app** | **daily** | **stops** | **fills** |
| App crashes | rare | stops | fills |
| Auto-update | occasional | stops | stops (see #1240) |
| Logout / reboot | daily | stops | stops |

In-process ties the solver's life to the app's, turning every "I'm done looking at this" into a trading interruption. The update row is a tie — the solver must stop to replace its own binary either way.

In-process does have one genuine security edge: no socket at all. It is small in practice. The only attacker it additionally stops is malware running as the operator's own user, and that process reads `filler-config.toml` and takes the private key without touching any API — no process model prevents that, and #1241 established that the OS keychain does not either on Windows or Linux.

### Why the solver binary, not an embedded runtime

The child is the same artifact npm and Docker ship, so it is the thing we already test. No Electron ABI coupling, no `electron-rebuild` on every Electron bump, and a solver crash does not take the window down — the UI can report it and offer a restart.

### Both transports, selected by context

`UiServer` supports **both** a TCP port and a Unix socket. Neither replaces the other:

| Context | Listens on |
|---|---|
| CLI (`simplex`) | TCP `127.0.0.1:8686`, unchanged |
| Desktop app | a Unix socket, and nothing else |

`start(port, host)` widens to take a target — `{ host, port }` or `{ socketPath }`. Everything else in `UiServer` is untouched: same route table, same handlers, same SSE.

[...]

PR #1248, opened 2026-09-09
[simplex]: add --no-open and --log-format json to the CLI

seunlanlege:
Part of #1237.

The desktop app will run `simplex` as a background process and save its output to a log file. Two
things the CLI always did got in the way. Each now has a flag on `run`.

Just the two flags. No Electron code, no changes to the server transport.

## The flags

**`--no-open`** — start the setup wizard, but don't launch a browser.

The app shows the wizard in its own window, so a second browser opening is wrong. Everything else is
the same: the wizard still starts, and still prints its URL.

Not to be confused with `--no-ui`, which switches the web UI off entirely.

**`--log-format json`** — print logs as one JSON object per line.

Colour codes are unreadable in a log file. `json` gives you plain NDJSON instead. It defaults to
`pretty`, so nothing changes for anyone running simplex in a terminal.

```bash
simplex run --no-open          # wizard starts, no browser
simplex run --log-format json  # {"level":30,"time":...,"msg":"..."}
```

## Nothing changes if you don't pass them

This was the main thing to get right, so it was checked by running the built binary and diffing the
output byte for byte.

| Command | What comes out on stdout | Browser opens? |
|---|---|---|
| `simplex run` | banner, colourised logs, wizard URL | yes |
| `simplex run --log-format pretty` | identical to the above | yes |
| `simplex run --no-open` | identical to the above | **no** |
| `simplex run --no-open --log-format json` | 2 lines, both valid JSON, no colour codes | no |

A bad value is rejected up front: `--log-format ndjson` exits 1 and says
`Allowed choices are pretty, json`.

## A bug was found in review, and fixed

The first version of `--log-format json` **crashed the solver** when whatever was reading its output
went away — which is the exact thing the flag was added for.

The short version: pino-pretty was quietly handling stream errors for us. Passing `process.stdout`
straight through removed pino-pretty, and took that error handling with it. A broken pipe then killed
the process instead of being ignored.

Reproduced by starting the solver in the background and closing the pipe, 3 runs each way:

| | before the fix | after |
|---|---|---|
| `--log-format json` | **died, exit 1** | keeps running |
| no flags | keeps running | keeps running |

Fixed in 93900ca6. The reasoning is written up in `docs/ai/Decisions.md`.

## CI wasn't running the new tests

CI ran simplex's `test:filler` script, which only covers four network-backed files. The 15 tests added
here were never run on a PR — you could delete the feature and CI would still go green.

e0b86680 adds a `test:unit` script and a CI step for it. Takes about 20 seconds, makes no network calls.

## Rebased onto #1249

#1249 moved `run`'s option declarations out of the bin into `addRunOptions`
(`src/cli/run-options.ts`). Both flags moved there with them, and `open` joined its `RunOptions`
interface. `logFormat` deliberately did not — see the implementation notes.

That turned out to be a real improvement, not just a merge. The tests can now parse the **actual**
declarations instead of a hand-copied mirror that could silently drift. Which in turn made this
testable:

```
simplex run --ui --no-open
```

That used to swallow `--no-open` and open a browser anyway, because the old `--ui [<[host:]port>]`
placeholder made commander treat the option as required. #1249 fixed it; there is now a regression
test pinning it. The test file went from 15 tests to 22, including the first coverage `--no-open` has
had.

## Known and not fixed here

- `--log-format` works on `run` but not on `paymaster-keeper`, which is also a long-running daemon.
- Whoever merges `simplex-ui-socket-listen-mode`: it adds an unguarded `console.log` on the same
  wizard path, which will break the "every line is JSON" promise. The two branches merge cleanly, so
  nothing will warn you.

## Commits

| | |
|---|---|
| 7070bca5 | the two flags |
| c5f4dbfc | the crash fix |
| 165271b4 | the CI gate |
| e6c1d8cf | drop the version bump |
| 8e8231b5 | docs, after the #1249 rebase |

Also included: the `docs/ai/` entries the package convention asks for, and a README note.

**No version bump.** Four simplex branches are open against #1237 at once, and a bump in each would
collide on the same field for nothing. Publishing is tag-driven, so the version can be set once on
whatever lands.

<details>
<summary>Implementation notes, if you're reviewing the code</summary>

**Why the flag is parsed twice.** The log sink has to exist before commander runs — it's registered
while `bin/simplex.ts` is still being evaluated. So `logFormatFromArgv` in `src/cli/log-format.ts`
scans raw argv for `--log-format`, and commander declares it separately for `--help` and validation.

The action deliberately ignores commander's parsed value. One reader, one answer. Otherwise a
command line the two could read differently — `simplex run -c --log-format json`, where commander
binds `--log-format` to `-c` — could put the ASCII banner in the middle of the NDJSON.

**Two existing properties that had to survive.** Both are documented in comments around
`consoleSink()`, and both still hold:

1. Formatting happens in-process, not through pino's worker-thread transport. json mode removes a
   transform rather than adding one.
2. Each writer gets its own pino-pretty transform, because two pino instances sharing one interleave
   their output. In json mode both writers share `process.stdout` instead, which is safe for the
   opposite reason: there's no transform in the middle reassembling anything. Checked by driving
   4000 records of 4 KB through a shared stdout — no partial or interleaved lines.

**Where stdout writes come from.** The ASCII banner and the wizard's URL line were the only two
non-logger writers. In json mode the banner is skipped and the URL becomes a log record with a `url`
field, which is more useful to a supervisor than prose it would have to scrape.

One caveat found in review: that covers every writer simplex controls, but not the whole process.
`@polkadot/api` prints a plain-text line to stdout if the Hyperbridge runtime upgrades mid-run. Rare,
and impossible on the wizard path, but json consumers should parse defensively. The docs say so now.

**Local test state.** The full suite doesn't pass cleanly in my worktree, and didn't before this
change either — 15 files fail on environment gaps (stale `@uniswap/sdk-core`, missing
`BINANCE_API_KEY` and RPC URLs, network-dependent simnode suites). I ran it on the pristine tree and
again with the change: the failing-file set is identical. `tsc --noEmit` is unchanged apart from a
pre-existing error shifting line number.

</details>

### DIFF
--- a/.github/workflows/test-sdk.yml
+++ b/.github/workflows/test-sdk.yml
@@ -141,6 +141,11 @@ jobs:
             - name: Run simplex data-store tests
               run: pnpm --filter="@hyperbridge/simplex" test:data
 
+            # test:filler covers the network-backed suites; the CLI and logger units
+            # were gated by nothing until 2026-09-10, so a broken flag stayed green.
+            - name: Run simplex test - cli and logger units
+              run: pnpm --filter="@hyperbridge/simplex" test:unit
+
             - name: Run simplex test
               run: pnpm --filter="@hyperbridge/simplex" test:filler
 
--- a/sdk/packages/simplex/README.md
+++ b/sdk/packages/simplex/README.md
@@ -92,7 +92,9 @@ simplex run --log-format json                # NDJSON on stdout instead of colou
 
 `--no-open` and `--log-format json` are for running the solver under a supervisor. The wizard still
 starts and still reports its URL under `--no-open`; only the browser launch is skipped. `json` makes
-every line of stdout one JSON object, with no ANSI escapes, which is what a captured log file needs.
+every line simplex writes to stdout one JSON object, with no ANSI escapes, which is what a captured log
+file needs. (One caveat: `@polkadot/api` prints a plain-text line to stdout if the Hyperbridge runtime
+upgrades while the solver is running, so parse defensively.)
 
 The curve-update API:
 
--- a/sdk/packages/simplex/README.md
+++ b/sdk/packages/simplex/README.md
@@ -86,7 +86,13 @@ Flags:
 simplex run -c filler-config.toml            # UI on 127.0.0.1:8686
 simplex run -c filler-config.toml --ui 9000  # custom port
 simplex run -c filler-config.toml --no-ui    # headless
+simplex run --no-open                        # start the wizard, don't launch a browser
+simplex run --log-format json                # NDJSON on stdout instead of colourised lines
 ```
+
+`--no-open` and `--log-format json` are for running the solver under a supervisor. The wizard still
+starts and still reports its URL under `--no-open`; only the browser launch is skipped. `json` makes
+every line of stdout one JSON object, with no ANSI escapes, which is what a captured log file needs.
 
 The curve-update API:
 
--- a/sdk/packages/simplex/docs/ai/ChangeLog.md
+++ b/sdk/packages/simplex/docs/ai/ChangeLog.md
@@ -333,8 +333,14 @@ byte-identical to a default run and launched no browser (checked with a stub `xd
 `--log-format pretty` was identical to passing nothing. A separate check ran two `LoggerContext`s into
 one shared `process.stdout` for 4000 records and found no interleaved or partial lines.
 
-Files: src/bin/simplex.ts, src/cli/log-format.ts (new), src/tests/cli/log-format.test.ts (new),
-README.md.
+Rebased onto #1249, which split `run`'s option declarations into `addRunOptions`
+(`src/cli/run-options.ts`). Both flags moved there, and `open` joined its `RunOptions` interface;
+`logFormat` deliberately did not. The test now parses the real builder instead of a hand-copied mirror,
+which is what makes the `--ui`-swallows-the-next-flag case (fixed by #1249) testable here at all — the
+suite went from 15 to 22 tests, including the first coverage `--no-open` has had.
+
+Files: src/bin/simplex.ts, src/cli/run-options.ts, src/cli/log-format.ts (new),
+src/tests/cli/log-format.test.ts (new), README.md.
 
 No version bump. Four simplex branches were open against #1237 at once and a bump in each would have
 collided on the same field for nothing — publishing is tag-driven, so the version can be set once on
--- a/sdk/packages/simplex/docs/ai/ChangeLog.md
+++ b/sdk/packages/simplex/docs/ai/ChangeLog.md
@@ -261,6 +261,37 @@ listened for. Each fix was mutation-checked: reverting it fails its test.
 Files: src/services/server/UiServer.ts, src/services/server/http-util.ts,
 src/services/tunnel/EmbeddedSshServer.ts, src/bin/simplex.ts, src/tests/ui-server-socket.test.ts,
 docs/ai/{ChangeLog,Decisions,Flow}.md.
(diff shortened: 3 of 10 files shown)


seunlanlege:
Audited this branch before merge. One real bug, in the new flag, now fixed on the branch.

### `--log-format json` crashed the process on a broken stdout pipe

`consoleSink()` returns a bare `process.stdout` in json mode. What that also removed, without my
noticing, was error handling. pino-pretty's `build()` ends in `pump(source, stream, destination)`, and
pump attaches `error` listeners to the destination — `process.stdout.listenerCount("error")` goes from
0 to 2 the moment `prettyStream({destination: process.stdout})` is constructed. A stream with no
listener turns a failed write into an unhandled `error` event, which Node turns into an uncaught
exception.

That lands exactly on the case the flag exists for. Spawning the built CLI detached with
`stdio: ["ignore","pipe","pipe"]` and then destroying the read end — the desktop app quitting while the
filler it spawned keeps running:

| invocation | before | after |
|---|---|---|
| `--log-format json` | exit 1, `Unhandled 'error' event`, 3/3 | survives 3/3 |
| default (pretty) | survives 3/3 | survives 3/3 |

`simplex run --log-format json \| head` was the same crash at a terminal.

Fixed in 93900ca6 with `if (logFormat === "json") process.stdout.on("error", () => {})`, at module
scope because `consoleSink()` is called per writer. Swallowing matches the pretty path, where the same
error tears down pump's chain and logging goes quiet while the filler keeps filling — `LoggerContext`
already treats a broken sink as the host's problem rather than a reason to fail a fill. All five
acceptance criteria re-verified against a fresh build afterwards.

### CI was not running these tests

`test-sdk.yml` gated simplex on `test:filler`, which pins four network-backed files. `src/tests/cli`
and `src/tests/logger.test.ts` were gated by nothing, so the 15 tests added with these flags never ran
on a PR — deleting the json branch left the suite green. e0b86680 adds a `test:unit` script over those
two paths and a CI step. Network-free, ~20s.

### Corrected in the docs

Three claims in the `docs/ai/` prose added with the flags were wrong:

- `init` and `validateConfig` were named as consumers of the process-wide sink. Neither resolves a
  logger. The real wizard-path consumers are `UiServer`'s `getLogger("ui")` and `setup-api`'s
  `getLogger("setup")`.
- The Flow.md walk placed the filler's second sink before the config branch. `startFiller` is only
  *defined* there and called inside `if (configPath)`, so on the no-config path there is no second
  sink until the operator saves the wizard. The `--no-ui` + no-config exit was missing entirely.
- "every line of stdout is one JSON object" holds for every writer this package controls, but not for
  the process: `@polkadot/util`'s logger routes `log` to `console.log`, and `bin/quiet.ts` patches only
  `console.warn`, so a Hyperbridge runtime upgrade mid-run would print one plain-text line. The claims
  are now scoped to what simplex itself writes, and the README says to parse defensively.

### Known, not fixed here

- **`simplex run --ui --no-open` launches a browser anyway.** `--ui [<[host:]port>]` contains both `<`
  and `[`, so commander 12 marks the option required *and* optional and unconditionally shifts the next
  token: `{ui: "--no-open", open: true}`. Pre-existing — it already swallows `--watch-only` and `-c` on
  `main` — but `--no-open` is a new casualty. Being handled separately.
- **`--log-format` is process-wide but declared only on `run`**, so `paymaster-keeper`, itself a
  supervised daemon, rejects it.
- **Heads-up for whoever merges `simplex-ui-socket-listen-mode`:** it adds an unguarded `console.log`
  on this same wizard path. Once both land, `--log-format json` emits non-JSON and nothing in the suite
  notices. It merges cleanly ~200 lines from the guarded site, so the resolver gets no signal.

royvardhan (approved):

PR #1255, opened 2026-09-10
[tesseract]: Retry submitting messages from hyperbridge that were cancelled

### LINKED ISSUE #1252: [tesseract]: Retry submitting messages from hyperbridge that were cancelled
Since messages from hyperbridge are now gated to a whitelisted relayer, tesseract should retry messages from hyperbridge that did not deliver successfully. Check request receipt first before retrying.

Refactor the retry unprofitable task to do this

dharjeezy:
## What this does

Deliveries out of hyperbridge are gated to a whitelisted relayer, so a batch that never landed was quietly lost: the fan out logged the messages it skipped and dropped them, and a failed submission took the whole batch with it. Requests addressed to the modules an operator lists are now parked in the same table the retry loop reads, and the outbound pipeline spawns that loop per EVM destination so something finally drains it. Every pass asks the destination for a request receipt before doing any proof or gas work, so anything already delivered, or past its timeout, is dropped rather than resubmitted into a revert.

## Configuration

`retry_modules` is an optional list of hex encoded module ids, matched against a request's `to` field. Listing at least one module is what switches parking and the retry loop on; absent or empty parks nothing and spawns no loop. `retry_frequency` (formerly `unprofitable_retry_frequency`, the old key is still accepted) only paces the loop, in seconds, and defaults to five minutes when unset.

## Notes

A parked request is proved at the height the destination is already on whenever that is at or past the height it was parked with, and no consensus message is added. Only when the destination has not reached that height does the batch carry the beefy proof for it, so it verifies on arrival instead of waiting for the next consensus update. Rows are deduplicated by commitment, keeping the lowest height a request was seen at.

Only post requests are parked. The consensus message in a failed batch is superseded by the next proof that comes along, and a get response is paid for on delivery rather than claimed.

This only covers deliveries to EVM chains. The inbound pipeline no longer parks cancelled or unprofitable messages for substrate destinations, since nothing ever drained those rows; they are logged and dropped.

closes #1252

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28396,7 +28396,7 @@ checksum = "8f50febec83f5ee1df3015341d8bd429f2d1cc62bcba7ea2076759d315084683"
 
 [[package]]
 name = "tesseract"
-version = "2.4.11"
+version = "2.4.12"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/tesseract/messaging/messaging/src/events.rs
+++ b/tesseract/messaging/messaging/src/events.rs
@@ -588,10 +588,49 @@ fn is_allowed_module(config: &RelayerConfig, module: &[u8]) -> bool {
 /// every module. Here it names none, which is what callers that need "the
 /// operator asked for this module by name" want.
 pub fn is_explicitly_filtered(config: &RelayerConfig, module: &[u8]) -> bool {
-	config.module_filter.as_ref().map_or(false, |filters| {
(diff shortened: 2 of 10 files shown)


Wizdave97 (commented):
The retry should explicitly be for only messages from the coprocessor.
Also the retry should not wait for the next consensus update, store the beefy proof height alongside the message and refetch it if necessary

Wizdave97 (approved):

Issue #1259, opened 2026-09-12
[evm]: Partial fills at the solver's rate


seunlanlege:
## Motivation

Partial fills settle every filler at the order's own rate: escrow is released pro rata to the output delivered, so a solver with a better price than the order's floor has no way to pass it on except overpaying on a fresh full fill. #980 brings partial fills to cross-chain orders under the same pro-rata rule. The HyperFX orderbook quotes per-solver rates and combines solvers for one order, so it needs a partial fill to settle at the filler's rate, on both kinds of order.

This adds one thing: a solver may quote its own rate on any fill.

## Design

Builds on #980. Nothing here touches the escrow messages, the cancel proofs or `_cumulativeReleased`; the rate lives entirely on the fill side.

### A fill states the input it takes

`FillOptions` gains `TokenInfo[] inputs`, one entry per output, each naming the matching input token and the escrow the solver takes for the output it delivers. Empty means today's behaviour, so existing solvers are unaffected. The rate is the ratio of the two raw amounts, no fixed point.

For a rate-quoted leg:

- the take may not exceed what the order's rate releases for the output outstanding, so a solver can better the order's rate but never worsen it; `RateBelowOrder` otherwise;
- the fill is credited output at the order's rate for that take, `take × totalRequired / inputs.amount`, capped at what is outstanding, and progress in `_partialFills` advances by the credited amount, so the mapping keeps its meaning across the upgrade;
- everything delivered above the credited amount is surplus.

### Surplus stays in the output token, on the destination

Surplus is split by `surplusShareBps` exactly like today's overpayment surplus: the beneficiary's share goes out with the fill, the protocol's share stays as dust and is reported through `DustCollected`, and all of it goes to the protocol when the order carries calldata. Denominating surplus in the output token is what lets this sit on top of #980 unchanged: nothing about a rate improvement crosses chains. An earlier draft refunded input-token surplus to the user, which would have had to travel inside the redeem message. Dropped.

### Composition with #980

- Progress is credited output at the order's rate, so `_cumulativeReleased(escrowTotal, filled, totalRequired)` and the redeem slices it produces are unchanged. The rate never reaches the source chain.
- Cancel proofs and the refund of the proven-unredeemed fraction are unchanged for the same reason.
- Same-chain and cross-chain fills should share one leg loop with the rate applied inside it. If #980 has not already merged the two loops, do it here.
- Ships as a governance upgrade together with #980, per #982.

### Execution model

[...]

PR #1265, opened 2026-09-14
[evm]: split IntentGatewayV2 into delegatecall modules

### LINKED ISSUE #1262: [evm]: split IntentGatewayV2 into delegatecall modules
## Motivation

`IntentGatewayV2` is at the EIP-170 code size limit. Glamsterdam may raise it on mainnet, but the other chains the gateway is deployed to adopt that on their own schedules, and #980 and #1259 need room now. The gateway is already an ERC-1967 proxy with UUPS upgrades through governance, so the implementation can delegate its heavy branches to separately deployed modules without changing the proxy, the address, or the upgrade path.

## Design

The implementation keeps every external entry point and its modifiers, does the shared validation, and delegatecalls the branch to a module. Delegatecall runs the module's code in the proxy's storage context with `msg.sender`, `msg.value` and `address(this)` preserved, so the existing `IntrinsicIntents` and `ExtrinsicIntents` logic runs unchanged. This is a fixed split of two or three modules, not a diamond: no selector table, no loupe.

### Rules

- **One storage layout.** Each module inherits `IntentsBase` exactly as the implementation does, declares no storage of its own, and never runs an initializer. The append-only rule applies to all of them at once. The `_partialFills` slot guard test and the hardcoded `_filled` slot for cancel proofs keep working because the layout is shared.
- **Module addresses are immutables in the implementation.** Upgrading a module is a normal UUPS upgrade with new immutables. No new governance surface, and the audit unit stays "this implementation version". Storage-held module addresses are rejected.
- **Guards only at the entry.** `nonReentrant`, `onlyHost`, `_checkRelayer` and the transient-storage solver selection stay on the implementation's functions. Module functions carry no `nonReentrant`, since the lock is already held when they run.
- **Reverts bubble byte for byte.** The SDK matches custom error selectors such as `Expired` and `FillExpired`, so the delegatecall wrapper returns the module's revert data verbatim.
- **Modules refuse direct calls.** Anyone can call a module at its own address against empty storage. An `onlyDelegated` check against a self-address immutable, the same trick UUPS uses for `onlyProxy`, closes that.
- **Code check at construction.** Delegatecall to an address without code succeeds with empty return data. Modules are immutable and cannot selfdestruct post-Cancun, so a code-size check in the implementation's constructor is sufficient.
- **Events and errors stay in the shared base.** Module code emits from the proxy's address, so the indexer sees nothing new, provided every ABI still contains the declarations.

### Things that look like problems but are not

- OpenZeppelin's EIP-712 bakes the domain separator for the deploying address into immutables and recomputes it when `address(this)` differs. Under delegatecall it recomputes with the proxy's address, which is what signers sign against. This is already the path every `select` verification takes behind the proxy today.
- Transient storage is per address and shared under delegatecall, so solver selection set in `fillOrder` is visible to module code.

### Suggested split

1. **Intrinsic module.** Same-chain fills and cancels.
2. **Extrinsic module.** Cross-chain fills and cancels, including the bodies of `onAccept` and `onGetResponse`.
3. **Admin module.** Governance and paymaster administration: deployments, params, dust sweeps, upgrades. Rarely run, a lot of bytes.

The implementation keeps `placeOrder`, `select`, the entry points, and the ISMP callbacks that authenticate and route.

### Cost

One extra delegatecall per fill or cancel, a few thousand gas including the calldata copy, against fills well above a hundred thousand. Forwarding `msg.data` to a module exposing the same signature removes the re-encoding if it matters.

## Tasks

- [ ] Module contracts inheriting `IntentsBase` with `onlyDelegated`, no storage, no initializer
- [ ] Implementation: module immutables, constructor code checks, delegatecall wrapper with verbatim revert bubbling
- [ ] Storage layout test asserting the implementation and every module agree slot for slot
- [ ] Deploy script: deterministic module deployment per chain before the implementation, then the governance upgrade through `Execute`; Etherscan verification for each module
- [ ] Forge test suite passes unchanged against the split; gas snapshot for fills and cancels before and after
- [ ] Docs: architecture note on the module layout and the upgrade procedure

seunlanlege:
Closes #1262

## Why

`IntentGatewayV2` sat 209 bytes under the EIP-170 limit, with #980 and #1259 needing room. The implementation now delegatecalls its bodies to two separately deployed modules. The proxy, its address and the governance upgrade path are unchanged.

## Layout

```
                     EIP712
                       |
                  IntentsBase
               /       |        \
  IntentGatewayV2  IntrinsicIntents  ExtrinsicIntents
                       |                  |
                 IntrinsicModule    ExtrinsicModule
```

- `IntentGatewayV2` is `IntentsBase, HyperApp, ReentrancyGuardTransient, Initializable`. It keeps every entry point and its guards, `placeOrder`, `select`, the shared validation of `fillOrder` and `cancelOrder`, `initialize`, `migrate()` and the views. `fillOrder` and `cancelOrder` delegatecall typed module functions. `onAccept` and `onGetResponse` forward `msg.data` after `onlyHost`.
- `IntrinsicModule is IntrinsicIntents`: same-chain fill and cancel. `IntrinsicIntents` is untouched.
- `ExtrinsicModule is ExtrinsicIntents`: cross-chain fill, both cancel routes, and the `onAccept` and `onGetResponse` handlers with governance and `Execute`. `ExtrinsicIntents` stays a `HyperApp`; its diff is `virtual` on the two callbacks, `relayer()` and `_setRelayer` moving to `IntentsBase`, and `Execute` delegatecalling the module's own address instead of the implementation.
- Module addresses are constructor immutables, checked for code. A module upgrade is an ordinary implementation upgrade. Modules inherit `IntentsBase`, declare no storage, refuse direct calls, and their reverts bubble byte for byte through a six-line assembly wrapper.

Details and the upgrade procedure are in `evm/src/apps/intentsv2/README.md`.

## Sizes and gas

| Contract | Runtime bytes |
|---|---|
| IntentGatewayV2 (before) | 24,367 |
| IntentGatewayV2 | 14,610 |
| ExtrinsicModule | 14,535 |
| IntrinsicModule | 6,843 |

Per-test `forge snapshot` against main: fills and cancels cost 4.3k to 6.2k more gas (a cold access to the module plus re-encoding the order), settlement deliveries 2.6k to 2.9k, governance requests about 3.3k.

## What changes for operators

- `setRelayer` and `upgradeToAndCall` live only on the extrinsic module and leave the gateway ABI. Governance reaches them through `execute_on_gateway` exactly as before; the selectors are unchanged.
- A relayer rotation can no longer ride in an upgrade's init data, since that data runs against the new implementation. Upgrade and rotation are two `execute_on_gateway` calls.
- `migrate()` takes no argument and bumps `VERSION` from 2 to 3. This release's upgrade carries `migrate()` as init data; the deploy script prints the exact `data` for `execute_on_gateway`.
- The constructor is `(intrinsic, extrinsic)`; `_owner` is gone. `intrinsicModule()` and `extrinsicModule()` are new getters.
- `DeployIntentGatewayImpl` and `DeployIntentGateway` share `IntentGatewayScript`: CREATE2 modules first, reused if already at their address, then the implementation. `deploy.sh --mode full` verifies the modules too.

## Tests

- The pre-existing intent suites pass against the split with only constructor, `migrate()` and ABI adaptations.
- `IntentGatewayModulesTest` (10 tests): storage layouts read from the forge artifacts agree slot for slot (`foundry.toml` now emits `storageLayout`), modules refuse direct calls, the constructor rejects codeless modules, module reverts bubble verbatim (a custom error and DAI's string reason), either module can be swapped by an implementation upgrade, and the release's own upgrade moves a version-2 proxy to 3 with escrow and peers intact.
- The live-fork test upgrades the mainnet proxy on a fork with `migrate()` and checks every readable piece of state.

## Not in this PR

- The SDK, simplex and indexer ABI copies still show the old constructor, `_owner`, `migrate(address)`, `setRelayer` and `upgradeToAndCall`. Nothing in those packages calls them, so they stay call-compatible; regenerate them in a follow-up.
- `IIntentGatewayV2` in `sdk/packages/core` follows the gateway (`migrate()`, the two host-only functions removed) with a patch bump to 2.3.5.


royvardhan (approved):
lgtm!

royvardhan:
I would still wait for another set of 👀. cc: @Wizdave97

Wizdave97:
> I would still wait for another set of 👀. cc: @Wizdave97 

I'll check it out

seunlanlege:
Would be good to merge this asap so we can unblock @royvardhan

Wizdave97:
## Review: fund-loss focus

I found no path that lets anyone take funds from solvers, users or the protocol. The review split into four areas:
- delegatecall and storage mechanics
- governance, upgrade and migration
- fill and cancel fund flows, compared line by line against `main`
- off-chain callers, gas and the deploy scripts

Nothing reached high or critical. The findings worth acting on are ways a governance or deploy mistake could lock escrow for good.

### 1. An implementation built with the wrong module addresses locks all escrow for good (medium: large impact, but needs an operator mistake)

`evm/src/apps/IntentGatewayV2.sol`: the constructor only checks that each module has code, and `migrate()` checks nothing.

1. An implementation is deployed outside `IntentGatewayScript`, with `(intrinsic, extrinsic)` swapped or a stale module address.
2. The upgrade still succeeds, because `migrate()` runs inside the implementation and never touches a module.
3. From then on every `onAccept`/`onGetResponse` delegates to a contract without that function and reverts. That blocks RedeemEscrow, RefundEscrow, governance and `Execute`.
4. `upgradeToAndCall` is now only reachable through `Execute`, so governance can't upgrade out of it.

**Why it's new:** before the split, every implementation carried `upgradeToAndCall` itself and so could always be upgraded again.

**Mitigation:** the script computes both addresses from each module's `creationCode`, so the scripted path is correct.

**Suggested fix:** have `migrate()` (or the constructor) call a view function on each module that identifies which module it is. Or keep `upgradeToAndCall` on the implementation, which is at 14,610 bytes and has room.

### 2. Upgrading to a wrong target with empty init data also locks the gateway (low, and not new)

- If `execute_on_gateway` names `IntrinsicModule` or an unrelated contract, the proxy loses `onAccept` for good.
- This release is protected, because the printed `migrate()` init data reverts against anything that isn't the gateway.
- Later releases that ship empty init data won't have that protection. The script also now prints three addresses an operator could paste by mistake.
- **Suggested fix:** in `ExtrinsicIntents.upgradeToAndCall`, check that the target answers `extrinsicModule()`.

### 3. `migrate()` no longer arms the relayer gate (low, only matters if a proxy is still at version 1)

- Before, `migrate(relayer)` set the relayer in the upgrade transaction. Now a version-1 proxy lands at version 3 with the gate open until a separate `setRelayer` arrives through `Execute`.
- The gate is the defence against forged-consensus deliveries of RedeemEscrow messages (#1199).
- Base and Arbitrum mainnet already return `version() == 2` with the relayer set. Ethereum and the testnets weren't checked.
- **Before upgrading:** check `version()` and `relayer()` on each chain.

### 4. The deploy script always prints `migrate()` as the init data (low)

`evm/script/IntentGatewayScript.sol`: any later implementation that keeps `VERSION = 3` gets init data that reverts with `InvalidInitialization`. Nothing is lost; the upgrade just fails until it is resent with empty data.

### 5. Minor

- **Gas:** the extra hop slightly eats into the relayer's 5% gas buffer (`tesseract/messaging/evm/src/tx.rs`, `gas_with_buffer`). A delivery short on gas fails inside the host and stays retryable, so the only cost is wasted relayer gas.
- **Stale docs:** the pallet comments in `modules/pallets/intents-coprocessor/src/lib.rs` and `types.rs` still say `Execute` delegatecalls the gateway's current implementation.

### Checked and found safe

- **Storage layout:** the gateway and both modules match slot for slot (checked with `forge inspect`), and `_filled` is still at slot 2, so the cancel storage proof is unaffected. The layout matches `main`.
- **Commitment:** computed once in the gateway, and the module receives that same order and commitment. Validation and execution can't act on different orders.
- **`msg.value`:** kept through the delegatecall and spent only once. Routing between the same-chain and cross-chain modules can't be forced the wrong way.
- **Guards:** `nonReentrant` and the solver selection still apply while module code runs.
- **Direct calls to a module:** fills and cancels revert on `onlyDelegated`. The host callbacks, `setRelayer` and `upgradeToAndCall` read `host == 0` from the module's own empty storage, so everyone is refused. Modules can't hold funds.
- **Revert bubbling:** `_delegate` can't report a module revert as success. When a callback fails, `EvmHost` deletes the receipt, so a message can't be marked delivered while its escrow stays locked.
- **Signatures:** EIP-712 checks stay in the gateway and use the proxy address as before.
- **Upgrades:** the version 2→3 upgrade and the chain of later upgrades through the module both work.
- **CREATE2 module reuse:** safe, since the address is fixed by the bytecode.
- **`Execute` into fill or cancel functions:** reachable, but only by governance, which can already install any implementation.
- **Callers outside the contracts:** nothing in the SDK, simplex, tesseract or the pallets calls the removed functions, and existing gas budgets absorb the extra 3–6k.
- **Sizes:** all three contracts are well under EIP-170 (confirmed with `forge build --sizes`).
- **Timeouts:** `onPostRequestTimeout` and `onGetTimeout` behave as before.

PR #1263, opened 2026-09-14
feat(simplex): add Electron desktop shell and first-run flow

ddboy19912:
## Summary

Implements #1237 by adding a private Simplex Electron desktop shell that runs the existing Simplex artifact with a verified, staged Node 24.19.0 runtime.

The desktop app communicates over a deterministic Unix socket or Windows named pipe, serves the existing SPA through `simplex://local/`, supports first-run onboarding under Electron user data, and leaves the solver running when Electron exits.

## Highlights

- Adds the `@hyperbridge/simplex-desktop` workspace package with Electron 44.3.0.
- Verifies and stages platform-specific Node runtimes.
- Adds deterministic socket naming, health-based attach/spawn behavior, and streaming protocol proxying.
- Adds desktop config discovery fallback and Orders SSE reconnect refresh.
- Adds unit, integration, real Electron E2E, runtime-integrity, and cross-platform CI coverage.
- Documents package-local development, lifecycle behavior, verification, and deferred scope.

## Quick start

Using Node 22.16+ and pnpm 11, open a terminal in `sdk/packages/simplex-desktop` and run:

```sh
pnpm install
pnpm --filter @hyperbridge/sdk build
pnpm --filter @hyperbridge/simplex build
pnpm stage:node
pnpm dev

Issue #1243, opened 2026-09-09
[simplex] Desktop lifecycle and OS integration


seunlanlege:
Part of #1235

## Why

A solver is a background service that happens to have a window. The current split already gets this right by accident: closing the browser does not stop the filler, because the browser is only a viewer. #1237 preserves that by running the solver as a detached child.

This issue is the OS-facing half: making it obvious the solver is alive, keeping it alive, and not lying about it.

## Autostart the app, not the solver

**The app starts at login, sits in the tray, and spawns the solver if it is not already running.**

This matters beyond convenience: the app is what runs the update sequence (#1240). A solver autostarted on its own, with the app never opened, would never be updated and could drift months behind.

Detached still earns its keep — quitting the app or crashing it leaves the solver filling.

## Single instance

The Unix socket from #1237 is the lock. If it exists and answers `/health`, attach; do not spawn. If it is stale, clean up and spawn.

Two solvers against one data directory would corrupt state and double-submit bids, so this has to be correct rather than best-effort. It also makes a CLI-started solver adoptable.

## Tasks

- [x] System tray icon showing at minimum running / paused / stopped, plus a quick pause and stop.
- [x] **Closing the window must not stop the solver.** Hide to tray.
- [x] Two distinct quit actions, clearly labelled: "Quit Simplex" (solver keeps filling) and "Stop solver and quit". Do not make the operator guess which one they got.
- [x] Launch at login, opt-in, via `app.setLoginItemSettings`. Applies to the app.
- [x] Single-instance via the socket, per above. `requestSingleInstanceLock` for the app window itself.
- [x] **App supervises the solver**: health-check the socket, surface a crashed solver in the UI, offer a restart. The solver crashing should be visible, not silent.
- [x] `powerSaveBlocker` while the solver is active, so a sleeping laptop does not silently stop filling. Show it in the UI — an operator should know their machine is being held awake.
- [x] Native menus: about, check for updates, open data directory, open log, quit. Check for Updates is in the menu but hidden until the updater lands (#1240).
- [x] **Log file on disk with rotation.** A detached child's stdout goes nowhere, so this is required rather than nice-to-have.
- [x] Decide on crash reporting. If yes, opt-in, and never capture config contents or key material.
- [x] Uninstall: leave the data directory in place and say so. Silently deleting bid records would strand reclaimable deposits.

## Linux tray caveat

[...]

PR #1269, opened 2026-09-15
[simplex-desktop]: improve simplex desktop security model

### LINKED ISSUE #1242: [simplex] Desktop security model
Part of #1235

## The desktop app changes the threat model

The renderer is ours, but a compromised renderer is then inside the trust boundary. It needs hardening of its own, and the desktop path should not weaken the checks that guard the CLI, Docker and tunnel paths.

## Transport is the access control

Both transports are first-class (#1237). Neither carries an application-level credential — **authentication is a property of how the connection arrived**, not something layered on top.

| Arrived via | Access control | Reaches |
|---|---|---|
| Unix socket (macOS, Linux) | file permissions (`0600`) | the owning user only |
| Named pipe (Windows) | default pipe DACL — **not** owner-only | see below |
| Tunnel | SSH device key + pinned host key | paired devices only |
| TCP loopback | none | **any local user** |

### Windows is weaker, and it was verified rather than assumed

Recorded in `docs/ai/decisions/2026-09-09-windows-named-pipes-are-not-equivalent-to-a-0600-socket-verified.md`. libuv's `pipe_alloc_accept` calls `CreateNamedPipeW` with `lpSecurityAttributes` set to `NULL`, and no libuv or Node API can supply one. Microsoft's documented meaning of `NULL`: full control to LocalSystem, Administrators and the creator owner, plus **read access to Everyone and the anonymous account**, in the machine-global `\\.\pipe\` namespace.

What saves it in practice: driving the HTTP API requires *write* access to send a request, and Everyone does not get write. So a non-administrator local user still cannot reach `/api/send` over the pipe. Administrators can — but an administrator reads `filler-config.toml` and takes the signing key without any API, so that was never a boundary this could hold.

It cannot be tightened from Node: the only pipe-ACL knob exposed is `listen({ readableAll, writableAll })`, and it *widens* access. `restrictSocketToOwner` returns early on win32 rather than pretending otherwise.

**Net:** the practical conclusion holds on all three platforms, but on Windows it rests on Everyone lacking write rather than on the kernel enforcing ownership. An embedder needing true owner-only on Windows must create the pipe from native code.

**Unix socket for the desktop.** On macOS and Linux the kernel is the access control. Nothing to generate, store, leak or rotate, and a web page cannot open a socket file at all — structurally, not by policy.

**The desktop app opens no TCP port at all**, remote access included — the tunnel dials outbound and injects an in-memory stream, so it needs no listener (#1237). A local browser is the only thing that requires a port, and that is the CLI's job.

**TCP stays as it is for the CLI.** `127.0.0.1:8686`, unauthenticated, unchanged.

## Decision: no bearer token

**Rejected.** A token on the TCP listener would close exactly one gap — another user account on the same machine — and it costs more than it returns.

The delivery is circular: the UI is served from the origin the token protects, so either the server hands it to anyone who can `GET /`, which is precisely the attacker, or it goes in a printed URL Jupyter-style. That second option breaks the installed PWA, whose `manifest.webmanifest` declares `start_url: "./"` — a per-launch token means the PWA launches unauthenticated every time.

What is already covered without one:

| Attacker | Covered by |
|---|---|
| Remote network | loopback bind (`UiServer.ts:278,291`) |
| DNS rebinding | `hostHeaderAllowed` (`:363`) |
| Web page, mutations | `X-Simplex-UI` (`:369`) — cannot be set cross-origin without a preflight the server never grants |
| Web page, reads | no CORS headers, so responses are opaque; a page can fire a GET but cannot read it |

**Accepted limitation: on a machine with other user accounts, any of them can reach the TCP UI and call `/api/send`** (`UiServer.ts:622`), which is an arbitrary token transfer. Loopback is per-machine, not per-user, so this is more open than `filler-config.toml`, which is mode `0600`.

The answer for a shared box is an authenticated transport, not a token: `--ui-socket` plus SSH port-forwarding, or the tunnel. Both already exist.

Revisit only if a concrete multi-user deployment appears. Do not add a token without also solving PWA delivery.

## Tasks

- [x] Unix socket listen mode, `0600`, as the desktop default (implemented in #1237).
- [x] No TCP listener in the desktop app, under any setting. Enabling the tunnel must not open one.
- [x] Renderer hardening: `contextIsolation: true`, `nodeIntegration: false`, `sandbox: true`, preload exposing a minimal typed API only.
- [x] CSP for the renderer.
- [x] Block navigation to external origins; route `target=_blank` to the system browser via `setWindowOpenHandler`.
- [x] Keep the existing loopback and Host-header checks intact — they still guard the CLI, Docker and tunnel paths.
- [x] Confirm the desktop path does not implicitly enable remote access. The tunnel stays opt-in exactly as it is.
- [x] Disable devtools in production builds, or gate them behind an explicit flag.
- [x] **Document the shared-machine limitation** (#1244) and point at `--ui-socket` and the tunnel as the authenticated alternatives.
- [x] Confirm the desktop app writes the config exactly as the CLI does — mode `0600`, warning banner intact — and never copies key material anywhere else (logs, crash reports, the renderer).
- [x] Keep `--ui`'s help text saying "Unauthenticated". It is accurate, and an operator binding a UI on a shared box should read it.

## The ceiling

None of this stops malware running as the operator's own user. That process reads `filler-config.toml` and takes the private key without touching any API.

**Nothing in this milestone changes that, deliberately.** Moving keys to the OS keychain was considered and closed as not planned (#1241): it protects meaningfully only on macOS, since Windows DPAPI and Linux libsecret are per-user and decrypt for any same-user process — the same attacker that can read a `0600` file.

So the desktop app keeps the CLI's handling: `filler-config.toml` mode `0600` with the warning banner (`src/cli/init/emit-toml.ts`), in `app.getPath("userData")`. **The requirement is that the desktop path is no worse than the CLI, not better.** Documented in #1244.

## Acceptance criteria

- [x] With the desktop defaults, no TCP port is listening.
- [x] A web page cannot reach the API on any transport.
- [x] The existing `http-util` tests still pass unchanged.
- [x] The shared-machine limitation is documented, not implicit.
- [x] `/security-review` clean on the diff.

ddboy19912:
## Summary

Closes #1242.

Hardens the Simplex Electron desktop security boundary while preserving the existing CLI, Docker, socket, and tunnel behavior.

The desktop renderer is restricted to `simplex://local`, exposes no Electron/Node API, and communicates with Simplex exclusively through the Unix socket or Windows named pipe. The desktop launch path never opens a TCP listener, including when the outbound tunnel is enabled.

## Work completed

- Added hardened Electron renderer preferences:
  - `contextIsolation: true`
  - `nodeIntegration: false`
  - `sandbox: true`
  - `webSecurity: true`
  - webviews disabled
  - no preload or IPC API
- Added a desktop-only Content Security Policy.
- Denied browser permissions by default, except sanitized clipboard writes from `simplex://local`.
- Blocked external navigation, redirects, subframe navigation, webviews, and Electron child windows.
- Routed approved `target="_blank"` links to the system browser.
- Added one shared manifest for all approved HyperFX and explorer destinations.
- Added fail-closed validation requiring every external destination to be a canonical HTTPS origin.
- Disabled DevTools in packaged builds.
- Confirmed the desktop daemon always launches with `--ui-socket` and never `--ui`.
- Confirmed enabling the outbound tunnel does not create a local TCP listener.
- Preserved the existing CLI loopback, Host-header, CORS, and CSRF behavior.
- Preserved the `--ui` “Unauthenticated” warning.
- Confirmed first-run configuration uses the existing atomic CLI writer, warning banner, and Unix mode `0600`.
- Added checks that signing material is not copied into renderer storage, API responses, logs, crash-report uploads, or other Electron profile files.
- Documented Unix, Windows named-pipe, same-user, and shared-machine limitations.
- Added the required #1242 feature note.
- Updated SDK contribution guidance to reflect that Changesets are not configured and private packages do not require release metadata.

## Issue checklist

- [x] Unix socket listen mode, `0600`, as the desktop default.
- [x] No TCP listener in the desktop app under any setting.
- [x] Enabling the tunnel does not open a local TCP listener.
- [x] Renderer uses context isolation.
- [x] Renderer disables Node integration.
- [x] Renderer runs in the Electron sandbox.
- [x] Renderer has no preload, IPC, or Electron API surface.
- [x] Desktop renderer has a restrictive CSP.
- [x] External-origin navigation is blocked.
- [x] Redirects and subframe navigation outside `simplex://local` are blocked.
- [x] Webviews are disabled.
- [x] Electron child windows are denied.
- [x] Approved external links open in the system browser.
- [x] External links use an exact HTTPS allowlist.
- [x] Approved destinations have one shared source of truth.
- [x] Existing loopback and Host-header checks remain intact.
- [x] Existing CORS and `X-Simplex-UI` CSRF protections remain intact.
- [x] Desktop does not implicitly enable remote access.
- [x] Tunnel remains explicitly opt-in.
- [x] Production DevTools are disabled.
- [x] Shared-machine limitations are documented.
- [x] Unix and Windows transport differences are documented.
- [x] Desktop config uses the CLI’s atomic writer.
- [x] Desktop config is created with mode `0600` on Unix.
- [x] The private-key warning banner remains intact.
- [x] Key material is not returned by `/api/config`.
- [x] Key material is not persisted in renderer storage.
- [x] Key material is not copied elsewhere in the Electron user-data profile.
- [x] No crash-report uploader is configured.
- [x] `--ui` still describes the TCP interface as “Unauthenticated”.

## Acceptance criteria

- [x] With desktop defaults, no TCP port is listening.
- [x] A hostile web page cannot read from or mutate the desktop API.
- [x] Existing `http-util`, Host-header, socket, and tunnel behavior remains unchanged.
- [x] Shared-machine limitations and authenticated alternatives are documented.
- [x] Security review is clean.
- [x] Security-review maintainability and repository-process findings are resolved.

## Verification

- [x] 26 desktop unit tests passed.
- [x] 6 socket protocol integration tests passed.
- [x] 4 real Electron E2E scenarios passed.
- [x] 43 existing Simplex socket, tunnel, setup, and logging regression tests passed.
- [x] 11 external-link and chain-metadata regression tests passed.
- [x] Desktop typecheck passed.
- [x] Desktop lint passed.
- [x] Desktop production build passed.
- [x] Simplex production build passed.
- [x] Simplex UI check and production build passed.
- [x] Formatting checks passed.
- [x] `git diff --check` passed.

## E2E coverage

The real Electron suite verifies that:

- the detached Simplex daemon survives Electron termination;
- relaunching Electron attaches to the existing daemon;
- the daemon has no TCP listener;
- SSE reconnects after daemon restart;
- twenty renderer reloads leave only one active SSE stream;
- CSP blocks injected inline scripts and external connections;
- an untrusted page cannot read or mutate `simplex://local`;
- only approved links reach the system browser;
- external navigation cannot replace the desktop document;
- first-run onboarding writes the config under Electron user data;
- the warning banner and Unix mode `0600` are preserved;
- signing secrets appear only in the intended config file.

## Security boundary

The desktop transport authenticates through operating-system access controls rather than a bearer token:

- Unix uses an owner-only `0600` socket.
- Windows uses a named pipe and retains the documented limitations of Node/libuv’s default pipe ACL.
- Remote access uses the authenticated outbound SSH tunnel and injects its streams directly into the socket-backed UI server.
- Software already running as the same operating-system user remains inside the accepted trust boundary.


seunlanlege (commented):
Thanks for this. The window settings, navigation blocking, permission handler and exact-host link allowlist all look right. The macOS E2E run also shows the CSP header does get applied on the `simplex://` protocol.

I've left one functional problem inline: the CSP breaks toasts in the desktop app. There's also a test that doesn't cover what it claims, a couple of scope and docs notes, and some nits.

**CI note, not caused by this PR.** The Desktop Electron E2E jobs for linux-x64, linux-arm64 and win32-x64 sat at "Downloading Electron binary..." for 20 minutes and were cancelled. main at b7f9e04f (#1263) shows the same thing. So the E2E suite has only run on macOS, and "4 real Electron E2E scenarios passed" is true for macOS only. The Linux `0600` and no-listener checks and the Windows named-pipe path haven't run in CI yet.

**PR description.** It lists the same items three times (Work completed, Issue checklist, Acceptance criteria) and includes process lines like "Security review is clean". A shorter description focused on the final behavior would be easier to review.

sdk/packages/simplex-desktop/src/desktop-security.ts
```diff
@@ -0,0 +1,143 @@
...
+	"form-action 'self'",
+	"frame-ancestors 'none'",
+	"frame-src 'none'",
+	"img-src 'self' data:",
+	"manifest-src 'self'",
+	"media-src 'none'",
+	"object-src 'none'",
+	"script-src 'self'",
+	"style-src 'self'",
+	"style-src-attr 'unsafe-inline'",
```
seunlanlege: **This blocks `<style>` tags that libraries add at runtime, which breaks toasts in the desktop app.**

`style-src 'self'` also covers `<style>` elements. `style-src-attr` only unlocks `style=""` attributes.

sonner adds its whole stylesheet as a `<style>` tag when it loads (`__insertCSS` in sonner 2.0.7). Radix Select and Radix Dialog (through react-remove-scroll-bar) add `<style>` tags too.

I served this exact policy to Chromium and ran sonner's injection code. The `<style>` tag was blocked with a CSP violation, while `style={}` attributes still applied.

In the app, `<Toaster>` sits after `.app-shell` in `ui/src/App.tsx`. Without sonner's CSS the toasts lose `position: fixed` and render unstyled below the whole app. Error toasts from wallet tools, chain settings and the solver link dialog are effectively invisible. The E2E suite doesn't catch it because no test triggers a toast.

Suggested fix:

```suggestion
	"style-src 'self' 'unsafe-inline'",
```

The risk is small. `img-src`, `font-src` and `connect-src` are all limited to `simplex://local`, so injected CSS has no outside host to send data to.

It would also help to record `securitypolicyviolation` events in the E2E while the wizard and operator pages load, and assert there are none. That catches this whole class of breakage.

sdk/packages/simplex/docs/ai/decisions/2026-09-15-the-desktop-renderer-has-no-electron-bridge.md
```diff
@@ -0,0 +1,31 @@
+# 2026-09-15 — The desktop renderer has no Electron bridge
+
+Decided: the Simplex desktop renderer receives no preload or IPC API. It is the existing browser SPA,
+served only at the standard, secure `simplex://local` origin and proxied to the solver's Unix socket
+or Windows named pipe. Context isolation, Chromium sandboxing, and web security are explicit; Node
+integration and webviews are disabled. Production builds also disable DevTools.
+
+The main document receives a desktop-only Content Security Policy. Scripts, workers, styles, assets,
+and API connections are local-only; frames, objects, media, and non-local form submissions are
+disabled. React's existing element-level style properties require `style-src-attr 'unsafe-inline'`,
```
seunlanlege: If the CSP moves to `style-src 'self' 'unsafe-inline'` (see the comment on `desktop-security.ts`), this sentence needs updating. It's worth saying why inline styles are acceptable here: scripts stay locked down, and images, fonts and connections can't reach outside hosts.

sdk/packages/simplex-desktop/scripts/e2e/desktop.e2e.mjs
```diff
@@ -290,14 +454,15 @@ test("first run writes a valid private config under Electron userData", async (t
...
+	const [daemonPid] = await waitForDaemonPids(userDataDir)
 
 	const config = {
 		simplex: {
 			signer: { type: "privateKey", key: TEST_KEY },
 			maxConcurrentOrders: 5,
-			substratePrivateKey: "bottom drive obey lake curtain smoke basket hold race lonely fit walk",
+			substratePrivateKey: TEST_SEED,
 			hyperbridgeWsUrl: "ws://127.0.0.1:9",
+			tunnel: { enabled: true, relay: "127.0.0.1:9" },
```
seunlanlege: This doesn't start the tunnel, so the `assertNoTcpListener` check below doesn't test the tunnel case.

`hyperbridgeWsUrl` is `ws://127.0.0.1:9`, so `startFiller` fails inside `onSaveAndStart` before `createTunnel` runs (`src/bin/simplex.ts`, around line 490). The test confirms this later by expecting the daemon to still be in `init` mode. The security test's tunnel is a stub on an in-process `UiServer`, with no listener check.

The behavior itself is correct from reading the code. `TunnelService` never listens, and `UiServer` only binds TCP under `--ui`.

I'd either say "by construction" in the PR description instead of "confirmed", or drop the tunnel config here so the test doesn't suggest more coverage than it has.

sdk/packages/simplex/docs/ai/changelog/2026-09-15-simplex-desktop-security-model-1242.md
```diff
@@ -0,0 +1,16 @@
+# Simplex desktop security model (#1242)
```
seunlanlege, line 1: This note and the new decision doc say almost the same thing. Both also repeat the existing `2026-09-09-windows-named-pipes-are-not-equivalent-to-a-0600-socket-verified.md` decision, and the desktop README adds a third copy.

`CLAUDE.md` asks for one concise note, plus a decision doc only when it covers something the note doesn't. I'd keep the decision doc (why there's no bridge, and the CSP choices) and cut this note to a few lines that point to it.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -40,28 +42,32 @@ async function prepareDaemon(): Promise<void> {
...
 		if (mainWindow.isMinimized()) mainWindow.restore()
 		mainWindow.focus()
 		return
 	}
 	if (!daemonReady) daemonReady = prepareDaemon()
 	await daemonReady
+	if (!sessionSecured) {
+		installSessionSecurity(session.defaultSession)
+		sessionSecured = true
+	}
```
seunlanlege: Nit: Electron's security checklist recommends installing the navigation and window-open guards in `app.on("web-contents-created", ...)`. Then any future window or view gets them automatically.

Installing the session handlers once in `app.whenReady()` would also remove the need for the `sessionSecured` flag.

sdk/packages/simplex-desktop/src/tests/daemon.test.ts
```diff
@@ -36,6 +37,7 @@ describe("daemon lifecycle", () => {
 			"--data-dir",
 			"/data/simplex",
 		])
+		expect(args).not.toContain("--ui")
```
seunlanlege: Nit: the exact `toEqual` above already rules out `--ui`, so this line adds nothing.

sdk/packages/simplex-desktop/scripts/e2e/desktop.e2e.mjs
```diff
@@ -279,6 +333,116 @@ test("the custom protocol reconnects Orders SSE and releases streams across 20 r
...
+	await page.evaluate((url) => window.open(url, "_blank"), approvedLink)
+	await waitFor(
+		async () => (await electronApp.evaluate(() => globalThis.__simplexOpenedUrls ?? [])).length === 1,
+		"approved external link",
+	)
+	assert.deepEqual(await electronApp.evaluate(() => globalThis.__simplexOpenedUrls), [approvedLink])
+	assert.equal(electronApp.windows().length, 1, "target=_blank must not create an Electron window")
+
+	await page.evaluate(() => window.open("https://example.com/", "_blank"))
+	await delay(100)
```
seunlanlege: Nit: this negative check waits a fixed 100 ms, so an `openExternal` call that arrives later would still pass.

One deterministic option: open the denied link, then a second approved link. Wait until the list has two entries and assert it equals the two approved URLs.

seunlanlege (commented):
Follow-up on #1273, which merged into this branch. Its fixes resolve the earlier review, but they introduced two new problems, and part of the double-solver race is still open.

The first three comments should be fixed before this reaches `main`: the stop-and-quit timeout, the missing crash notification, and the boot-time race. The rest are nits.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -1,92 +1,426 @@
...
+		} finally {
+			startPromise = undefined
+		}
+	})()
+	return startPromise
+}
+
+async function waitForStopped(): Promise<void> {
+	if (!supervisor) return
+	const deadline = Date.now() + STOP_TIMEOUT_MS
```
seunlanlege, line 199: **"Stop solver and quit" now errors when the drain takes more than 30 seconds.**

Since the #1273 fixes, the solver holds its socket until it has finished in-flight orders and redeemed vault positions on-chain. That often takes longer than 30 seconds.

This loop still gives up at 30 seconds. So the operator sees "Simplex could not stop the solver" and the app doesn't quit, even though the stop is going fine.

Suggested fix: treat `stopping` as success here. The solver is detached and finishes draining on its own, so "Stop solver and quit" can quit as soon as it sees `stopping`.

sdk/packages/simplex-desktop/src/solver-supervisor.ts
```diff
@@ -0,0 +1,177 @@
...
+	if (response.status < 200 || response.status >= 300) {
+		throw new Error(`Simplex could not ${action}: ${response.status}${response.body ? ` ${response.body}` : ""}`)
+	}
+}
+
+export function holdsMachineAwake(status: SolverStatus): boolean {
+	return status.state === "running"
+}
+
+export function shouldNotifySolverFailure(
```
seunlanlege, line 90: **Real crashes no longer raise a notification.**

This only notifies on `unreachable`, but a crashed solver never becomes `unreachable`. When the process dies without cleaning up, its socket file stays behind. Connecting then fails with `ECONNREFUSED`, which `probeHealth` maps to `stopped`. I checked this by killing a process listening on a Unix socket with SIGKILL.

So the tray changes to Stopped, but no notification appears. #1243 asks for one.

On Unix, the probe can already tell the two cases apart:
- A clean stop deletes the socket file, so the probe fails with `ENOENT` and the status is `stopped` with reason `absent`.
- A crash leaves the file behind, so the probe fails with `ECONNREFUSED` and the reason is `stale`.

Suggested fix: notify when a live state changes to `stopped` with reason `stale`. On Windows, a named pipe disappears either way, so that case needs another approach or can stay as it is.

sdk/packages/simplex-desktop/src/daemon.ts
```diff
@@ -89,6 +88,7 @@ export async function ensureDaemon(options: {
 	const probe = options.probe ?? probeHealth
 	const initial = await probe(options.launch.socketPath)
 	if (initial.state === "ready") return { attached: true, mode: initial.mode }
+	if (initial.state === "stopping") return { attached: true, mode: initial.mode }
```
seunlanlege, line 122: **A second solver can still start while the first is booting.**

This handles a solver that is shutting down, but not one that is still starting. A configured solver starts filling before it binds its socket ([simplex.ts:440](https://github.com/polytope-labs/hyperbridge/blob/cdb0f1cf3c87acf33693f039514fe5b1f6eb23b7/sdk/packages/simplex/src/bin/simplex.ts#L440), then [:458](https://github.com/polytope-labs/hyperbridge/blob/cdb0f1cf3c87acf33693f039514fe5b1f6eb23b7/sdk/packages/simplex/src/bin/simplex.ts#L458)). If binding fails, it keeps filling anyway.

Two ways to hit it:
- The operator quits the app while the solver is booting, which takes tens of seconds, and relaunches. The probe finds no socket and spawns another solver.
- Startup takes more than 120 seconds, `ensureDaemon` gives up, and the operator clicks Restart.

Either way, the second process fills on the same signer and then fails to bind.

Suggested fix: in socket mode, bind the socket before starting the filler. The setup wizard already works this way: the server binds first, and save-and-start then boots the filler.

sdk/packages/simplex-desktop/src/tray-icon.ts
```diff
@@ -0,0 +1,26 @@
+import { join } from "node:path"
+import type { SolverStatus } from "./solver-supervisor"
+
+export const TRAY_ICON_STATES = [
+	"starting",
+	"setup",
+	"running",
+	"paused",
+	"stopping",
```
seunlanlege, line 9: Nit: `stopping.png` and its Template and `@2x` versions have exactly the same pixels as the `starting` icons. The menu label tells the two states apart, but the tray icon doesn't.

sdk/packages/simplex/src/services/server/UiServer.ts
```diff
@@ -1117,8 +1126,19 @@ export class UiServer {
 		}
 
 		if (path === "/api/stop") {
-			if (this.mode !== "operator") return sendJson(res, 409, { error: "Filler is not running" })
 			if (method !== "POST") return sendJson(res, 405, { error: "Method not allowed" })
+			if (this.stopping) return sendJson(res, 202, { stopping: true })
+			if (this.mode === "init") {
+				if (this.startState === "starting") return sendJson(res, 409, { error: "Filler startup is already in progress" })
+				if (!this.setup?.stop) return sendJson(res, 409, { error: "Filler is not running" })
+				this.beginStopping()
```
seunlanlege, line 1139: Nit: after this `beginStopping()`, a save-and-start request that arrives before the process exits is still accepted. `saveAndStart` only rejects when a start is already in progress ([setup-api.ts:366](https://github.com/polytope-labs/hyperbridge/blob/cdb0f1cf3c87acf33693f039514fe5b1f6eb23b7/sdk/packages/simplex/src/services/server/setup-api.ts#L366)), so it can boot a filler inside a process that is about to exit.

The timing window is small. Rejecting setup changes while `stopping` is set would close it.

sdk/packages/simplex/docs/ai/changelog/2026-09-15-simplex-desktop-lifecycle-and-os-integration-1243.md
```diff
@@ -0,0 +1,30 @@
...
+macOS and Windows and through per-user XDG autostart on Linux.
+
+No remote crash reporter is enabled: diagnostics remain in the existing five
+rotating launch logs. Desktop removal preserves the user-data directory by
+contract so configuration, bid records, and reclaimable-deposit history are not
+silently deleted. The documentation also makes clear that login startup and
+sleep prevention are not substitutes for running the Simplex container on a
+VPS when continuous uptime is required.
+
+Review hardening keeps the socket bound and reports `stopping` until graceful
```
seunlanlege: Nit: this "Review hardening" paragraph describes review history. CLAUDE.md asks notes to describe only the final behavior, so I'd merge these points into the paragraphs above and drop the framing.

ddboy19912:
@seunlanlege This is ready

seunlanlege (approved):
All review threads are addressed. The CSP now allows runtime styles and the E2E fails on any CSP violation. The configured solver binds its socket before filling, and stop-and-quit, crash alerts and the tray icons are fixed. Desktop E2E passes on all four platforms.

PR #1268, opened 2026-09-15
[simplex]: Remove phantom bid support from simplex

dharjeezy:
Simplex stops polling Hyperbridge for phantom orders and stops bidding on them, ahead of pricing from operator limit orders posted to the HyperFX orderbook. The real bid path, its retraction sweep and the bid store are unchanged, and `encodePhantomBidDeclaration` stays in the SDK since a limit order still declares its accepted source chains with it.

`RuntimeState` now carries only `paused`, so the filler no longer takes a state store, and `FillerConfig.uniswapV4PositionsByChain` goes with the declaration that was its only reader.

Part of #1267.

Issue #1267, opened 2026-09-15
[simplex] Post limit orders to the HyperFX orderbook and remove phantom bids


Wizdave97:
## Summary

Simplex stops bidding on coprocessor phantom orders and stops pricing from curves. Instead, the operator creates **limit orders**, and Simplex posts them to the [HyperFX orderbook](https://github.com/polytope-labs/hyperfx-orderbook).

- Simplex prices every incoming IntentGateway order against those limit orders.
- It never pays out more than the matched limit order has left.
- After a fill it resizes the limit order and reposts it.

The orderbook's design and GraphQL schema are in its [`PLAN.md`](https://github.com/polytope-labs/hyperfx-orderbook/blob/main/PLAN.md). Every schema below is copied from it. The orderbook settled its amount units in polytope-labs/hyperfx-orderbook#29, now merged: **every amount the orderbook API takes or returns is at 1e18**. Mandatory source-chain declarations landed in polytope-labs/hyperfx-orderbook#33. The ingest, cancel and heartbeat mutations are in review in polytope-labs/hyperfx-orderbook#32, which makes cancel and heartbeat messages name their solver and removes `balance` from `OrderAccepted` and `OrderRejected`. The schemas below follow #32.

### Orderbook conventions used throughout

- **Scalars.** `BigInt` is a U256 as a decimal string. `Address` is 0x-prefixed, 20 bytes, returned lowercase. `Bytes` is 0x-prefixed hex.
- **Amounts are 1e18.** Every `BigInt` amount (sizes, balances, `amountIn`/`amountOut`, dust floors) is normalized to 18 decimals, whatever the token's decimals on its chain. Raw amounts appear only inside the signed UserOp.
- **Prices are quote per 1 base, at 1e18, on both sides.** A USDC/cNGN book reads about 1,500 for bids and asks alike.
- **Base and quote come from the orderbook's pair config** (`books { base quote }`), not from Simplex's `token0`/`token1`.
- **Sides.** An order whose input (what the solver receives) is the base and output (what the solver pays) is the quote is a **BID**. Input quote with output base is an **ASK**. One order is one entry on one side; the orderbook never mirrors it.
- **Identifiers.** Chains are state machine ids (`"EVM-8453"`). Tokens are symbols in queries and addresses inside the UserOp.
[...]

PR #1270, opened 2026-09-15
[simplex]: Post operator limit orders to the HyperFX orderbook

dharjeezy:
An operator can now create, list, fetch and cancel limit orders, and simplex advertises each one on the HyperFX orderbook. A limit order is what simplex offers to pay, stored in a new `limit_orders` table in `bids.db`, and the orderbook entry is a derived copy built from a signed `fillOrder` UserOperation that nothing ever submits. Every request is validated against `serverInfo` and `books` before it is stored, so the rejection codes the orderbook keeps for bad requests should never come back.

Pricing incoming orders against these limit orders, and drawing `remaining` down on a fill, land separately.

**The create request deviates from #1267 deliberately.** The issue publishes `{ book, side, fillChain, price, size, acceptedSources, ttlSecs, expiresAt? }`; this takes `{ fillChain, tokenIn, amountIn, tokenOut, amountOut, acceptedSources, ttlSecs, expiresAt? }` and derives the book, the side and the rate from the two amounts. An operator states what they take in and what they pay out, so the direction falls out of the request instead of being a separate field they can contradict, and there is no question of which price they meant. The issue needs updating, or this noting when it merges.

Review fixes are in the last two commits. A retryable `OrderSubmissionFailed` no longer retires the order: it stays `open` with the reason on it for reconciliation to post again, where before one timeout killed it permanently. A `cancelOrder` that never got an answer reads as `failed` rather than `UNKNOWN_ORDER`, so it no longer clears a commitment whose entry may still be live. `ORDER_EXISTS` is answered by reading `order(solver, commitment)` and treating a live entry of ours as the posting it already is, instead of bumping the nonce and leaving two entries behind one liability. And `serverInfo.chains` is checked before posting, so a mistyped source chain is refused on the way in.

Two of that review's points are answered in #1271 rather than here, because they need code this PR does not have: the missing `limitOrderResized` and `limitOrderFilled` events, which come out of the fill path, and `GET /api/limit-orders/:id` carrying the fills that consumed the order, which reads the `limitOrderId` bids only carry from there on.

`docs/ai/decisions/2026-09-16-the-protocol-fee-is-already-out-of-an-order-by-the-time-we-price-it.md` records the fee basis: `placeOrder` emits `reducedInputs`, so what the matcher prices is already net, the solver gets the rate it signed, and `bookPrice` is the same trade seen from the swapper's side. The invariant worth keeping is that the event carries reduced inputs; gross ones would make every payout `1/(1-f)` too large with nothing failing.

Stacked on #1268, which it needs for `encodePhantomBidDeclaration`. 

Part of #1267.


sdk/packages/simplex/src/orderbook/limit-orders.ts
```diff
@@ -0,0 +1,420 @@
...
+				orderNonce: orderNonce.toString(),
+				status: "open",
+				lastError: null,
+			})
+			return { order: stored!, result }
+		}
+
+		const message = `${result.code}: ${result.message}`
+		this.logger.error({ id: order.id, err: message }, "Orderbook refused the limit order")
+		return { order: (await this.store.setStatus(order.id, "rejected", message))!, result }
```
Wizdave97, line 403: **A transient orderbook failure kills the limit order permanently.**

`post()` funnels `kind: "failed"` into the same branch as `rejected` and sets status `rejected`. But `live()` only covers `open` and `resizing`, so neither `reconcile()` nor `renewExpiring()` will ever look at the row again — `rejected` is terminal.

So one `DATABASE_UNAVAILABLE`, one `TIMEOUT`, or one dropped connection during `POST /api/limit-orders` and the operator's order is dead, with nothing retrying it. #1267 §2 is explicit: "`OrderSubmissionFailed` with `retryable` → Retry with backoff."

`limit-orders.test.ts` pins the current behaviour ("reports an unreachable orderbook as retryable and leaves the order rejected"), so this needs a decision rather than a quiet patch. The smallest fix that restores the spec: keep the row `open` with `lastError` set when `result.retryable`, and let reconciliation repost it — it already knows how to post a row whose commitment is missing.

sdk/packages/simplex/src/orderbook/limit-orders.ts
```diff
@@ -0,0 +1,420 @@
...
+			domain: { name: eip712DomainName, version: eip712DomainVersion },
+			types: CANCEL_ORDER_TYPES,
+			primaryType: "CancelOrder",
+			message: { commitment, timestamp },
+		})
+		try {
+			return await this.client.cancelOrder({ solver: this.signer.address, commitment, timestamp, signature })
+		} catch (err) {
+			if (err instanceof OrderbookRequestError) {
+				return { kind: "rejected", code: "UNKNOWN_ORDER", message: err.message }
```
Wizdave97: **A transport failure is reported as `UNKNOWN_ORDER`.**

`OrderbookRequestError` means we never got an answer. `UNKNOWN_ORDER` is the orderbook's considered answer that the entry is already gone. Collapsing them loses the distinction the class comment two files over goes out of its way to draw.

It has a real consequence: `cancel()` reads `UNKNOWN_ORDER` as "already gone" and clears the commitment, so a `cancelOrder` that merely timed out leaves a live entry on the book that nothing here owns any more. It also makes the duplicate-posting path in `repost()` much more likely to fire.

Suggest letting the error propagate, or a distinct `kind: "failed"` on `CancelOrderResult`, so callers can tell "gone" from "don't know".

sdk/packages/simplex/src/orderbook/limit-orders.ts
```diff
@@ -0,0 +1,420 @@
...
+	/**
+	 * Submits the posting, answering a nonce the orderbook has already seen with a
+	 * fresh one. `REPLAYED` and `ORDER_EXISTS` both mean the op hashed to something
+	 * it remembers, and it remembers every hash forever, so bumping the nonce is
+	 * the only way past. Anything else is returned as it came.
+	 */
+	private async submit(order: LimitOrder): Promise<{ result: SubmitOrderResult; orderNonce: bigint }> {
+		const orderNonce = BigInt(order.orderNonce)
+		const first = await this.buildAndSubmit(order, orderNonce)
+		if (first.kind !== "rejected" || (first.code !== "REPLAYED" && first.code !== "ORDER_EXISTS")) {
```
Wizdave97, line 421: **`ORDER_EXISTS` and `REPLAYED` are not the same thing.**

From `crates/store/src/orders.rs`:

- same `user_op_hash`, order **gone** → `Replayed` (line 269)
- same `(solver, commitment)` still **live** → `OrderExists` (line 292)

So `ORDER_EXISTS` means there is a live entry sitting at the commitment we just tried to post. Bumping the nonce puts a *second* entry up and records only the new commitment, leaving the first untracked until reconciliation happens to sweep it — the same "two live entries behind one liability" this stack is careful about everywhere else.

#1267 §2 handles it: "If `order(solver, commitment)` is ours and live, treat it as accepted. Otherwise bump `orderNonce` and resubmit once." That needs an `order(solver, commitment)` query, which the client doesn't have yet.

(For what it's worth the window is narrow today, since `repost()` always bumps the nonce and so always changes the commitment. It widens the moment anything reposts at a fixed nonce.)

sdk/packages/simplex/src/orderbook/client.ts
```diff
@@ -0,0 +1,173 @@
...
+} from "./types"
+
+const LIMITS_QUERY = `
+	query Limits {
+		serverInfo {
+			minOrderTtlSecs
+			heartbeatIntervalSecs
+			signatureSkewSecs
+			maxBatchSize
+			minOrderSizes { symbol size }
```
Wizdave97, line 19: `serverInfo.chains` is available and worth selecting here.

The schema documents it as "Every configured chain by state machine id (`EVM-8453`), sorted: the chains orders fill on and the source chains they may declare" — which is exactly half of what `UNSUPPORTED_SOURCE_CHAIN` is testing, and #1267 §2 asks `acceptedSources` to be validated against it before posting.

`types.ts` currently says of `UNSUPPORTED_SOURCE_CHAIN` that "nothing here can check ahead of asking". That's only true of the other half (whether the input symbol is registered on that chain, per `Config::registers_symbol`). The chain-is-served half is checkable, and it's the half an operator is most likely to get wrong by typo.

sdk/packages/simplex/src/orderbook/limit-orders.ts
```diff
@@ -0,0 +1,420 @@
...
+/**
+ * One limit order as the operator states it: what simplex takes in, and what it
+ * pays out for that.
+ *
+ * "10000 USDC for 139000000 CNGN" is `tokenIn: "USDC", amountIn: 10000e18,
+ * tokenOut: "CNGN", amountOut: 139000000e18`. The rate and the side of the book
+ * follow from those, so the order is directional by construction: it prices
+ * USDC to CNGN swaps and never the reverse.
+ */
+export interface CreateLimitOrderRequest {
```
Wizdave97, line 44: **This is a different HTTP contract from the one #1267 published.**

The issue specifies `POST /api/limit-orders` with `{ book, side, fillChain, price, size, acceptedSources, ttlSecs, expiresAt? }`; this takes `{ fillChain, tokenIn, amountIn, tokenOut, amountOut, ... }` and derives book, side and price from the amounts.

I think the implementation's shape is the better one — direction falls out of the request instead of being a separate field an operator can contradict, and it sidesteps open question 1 about which price to enter. But it's a published contract, so anything written against the issue breaks silently, and the deviation should be called out in the PR description rather than discovered.

Either change the issue, or note it here.

sdk/packages/simplex/src/simplex.ts
```diff
@@ -113,6 +120,9 @@ export interface SimplexEvents {
 	"order:fill-observed": { commitment: HexString; filler: string; chainId: number; txHash?: string; ours: boolean }
 	rebalance: { success: boolean; transferCount?: number; executedCount?: number; error?: string }
 	activity: ActivityEvent
+	"limit-order:posted": { order: LimitOrder }
+	"limit-order:rejected": { order: LimitOrder; code: string; message: string }
+	"limit-order:cancelled": { order: LimitOrder }
```
Wizdave97, line 125: Two of the five events #1267 §2 asks for are missing: `limitOrderResized` and `limitOrderFilled`.

All three that exist are emitted from `LimitOrderController`, which only sees operator-initiated create and cancel. A fill-driven resize runs in the filler (`settleFilledLimitOrder` → `settleFill`) and emits nothing, so an operator watching the SSE stream sees a limit order's size go stale — which is the one change they didn't initiate and most need to be told about.

The under-funded case that `reconcile()` writes to `lastError` has the same problem.

sdk/packages/simplex/src/services/server/UiServer.ts
```diff
@@ -804,6 +809,38 @@ export class UiServer {
...
+			return sendJson(res, 405, { error: "Method not allowed" })
+		}
+
+		const limitOrderMatch = path.match(/^\/api\/limit-orders\/([\w-]+)$/)
+		if (limitOrderMatch) {
+			if (this.mode !== "operator") return sendJson(res, 409, { error: "Filler is not running" })
+			const id = limitOrderMatch[1]
+			if (method === "GET") {
+				return this.handleLimitOrders(res, async () => {
+					const order = await this.operator!.limitOrders!.get(id)
```
Wizdave97: #1267 §2 specifies this route as "one limit order, **with the fills that consumed it**". It returns the order alone.

The data is there — bids carry `limitOrderId` and `reservedAmount` — so this is mostly a store read away, and it's what makes the `remaining` figure on the row explicable to an operator rather than just a number that shrank.

Wizdave97 (commented):
Reviewed against [#1267](https://github.com/polytope-labs/hyperbridge/issues/1267) and the orderbook's own source at `polytope-labs/hyperfx-orderbook@main`.

**The wire format is right.** I checked every boundary against the server rather than against the issue text:

- `tests/fixtures/orderbook-schema.graphql` is byte-identical to the orderbook's `schema.graphql`, and `orderbook-userops.json` to its `fixtures/userops.json`.
- The EIP-712 `CancelOrder` / `Heartbeat` field order and the two-field domain match `crates/core/src/messages.rs` exactly.
- The posted op satisfies `crates/core/src/verify.rs`: `session = 0`, `output.assets[0].amount = 0`, FillOptions **v2**, `source == destination == fillChain`, nonce = `bidNonceKey(commitment, 0) << 64`, commitment-prefixed 97-byte signature. The declaration round-trips through `crates/core/src/paymaster.rs`.
- `schema.test.ts` validating every sent document and pinning all three enums against the published schema is the right call — it catches exactly the class of drift nothing else here would.

Worth recording somewhere, because it isn't obvious and it *does* hold: the gateway takes `protocolFeeBps` off the **input** (`IntentGatewayV2.sol:356`, with the commitment computed over the reduced inputs) while the orderbook shades the **output** (`crates/core/src/haircut.rs`). For a linear fee the algebra coincides, so `bookPrice` and what simplex actually pays agree end to end. The "Pricing basis" paragraph in #1267 is correct.

Comments below, the first being the one I'd block on.

dharjeezy:
All seven actioned. Four here in `e434064b` and `11ea8c4c`: the retryable failure keeps the row `open` for reconciliation instead of retiring it, `CancelOrderResult` has a `failed` variant so a cancel that got no answer stops reading as `UNKNOWN_ORDER`, `ORDER_EXISTS` is answered by reading `order(solver, commitment)` and taking a live entry of ours as the posting it already is, and `serverInfo.chains` is checked before posting.

Two landed in #1271 rather than here, because they need the fill path and the `limitOrderId` on bids: the missing `limitOrderResized` and `limitOrderFilled` events, and `GET /api/limit-orders/:id` returning the fills that consumed the order.

The contract deviation is called out in the description. Your `UNKNOWN_ORDER` thread is collapsed as outdated now, since the fix replaced the lines it sat on.

On the fee basis: I could not reproduce "the algebra coincides" at first, because I assumed `inputNet` was gross. It turns on `placeOrder` emitting `inputs: reducedInputs`, so what the matcher prices is already net, the solver gets the rate it signed, and `bookPrice` is the same trade from the swapper's side. Written up in `docs/ai/decisions/2026-09-16-the-protocol-fee-is-already-out-of-an-order-by-the-time-we-price-it.md`, framed around the invariant rather than the arithmetic: gross inputs would make every payout `1/(1-f)` too large with nothing failing.

@Wizdave97

Wizdave97 (commented):
Re-reviewed at `11ea8c4c`. All seven comments are addressed, and `e434064b` is a better fix than I proposed on two of them:

- **Retryable failures** now keep the row `open` with `lastError` set, so reconciliation reposts it. A timeout no longer retires an order the operator still wants.
- **Transport failures** are their own `kind: "failed"` on `CancelOrderResult` rather than being relabelled `UNKNOWN_ORDER`, and `cancel()` now keeps the commitment on the row when it could not confirm the entry is gone — which is the part that actually prevents orphaning.
- **`ORDER_EXISTS`** looks the entry up through the new `orderAt` query and adopts it as `unchanged`, falling back to the nonce bump only when the lookup fails. Exactly what §2 asks for.
- **`serverInfo.chains`** is validated for both `fillChain` and every declared source, with the right caveat that it only answers half of `UNSUPPORTED_SOURCE_CHAIN`.
- The `:id` route returns fills via `withFills`, and `limit-order:resized` / `limit-order:filled` now exist.

The create-API shape still differs from the one #1267 published (`{tokenIn, amountIn, tokenOut, amountOut}` rather than `{book, side, price, size}`). I still think the implementation's shape is the better one — just worth a line in the PR description or an edit to the issue so the contract has one published form.

No further comments from me on this one.

sdk/packages/simplex/docs/ai/changelog/2026-09-15-limit-orders-posted-to-the-hyperfx-orderbook.md
```diff
@@ -0,0 +1,69 @@
...
+their own chains.
+
+## The API
+
+`Simplex.limitOrders` and four routes on the operator server:
+
+```
+GET    /api/limit-orders?status=&chain=&book=
+GET    /api/limit-orders/:id
+POST   /api/limit-orders     { fillChain, tokenIn, amountIn, tokenOut, amountOut, acceptedSources, ttlSecs?, expiresAt? }
```
Wizdave97: Why do we have expiresAt and ttl, expiresAt can be derived from ttl

sdk/packages/simplex/docs/ai/changelog/2026-09-15-limit-orders-posted-to-the-hyperfx-orderbook.md
```diff
@@ -0,0 +1,69 @@
...
+has accepted, so a fresh nonce is the only way past.
+
+## Config
+
+```toml
+[orderbook]
+enabled = true
+url = "https://orderbook.hyperbridge.network/graphql"
+defaultTtlSecs = 900
+renewMarginSecs = 120
```
Wizdave97: Don't renew limit orders after expiry

sdk/packages/simplex/docs/ai/changelog/2026-09-15-limit-orders-posted-to-the-hyperfx-orderbook.md
```diff
@@ -0,0 +1,69 @@
...
+The derived rate is rounded in simplex's favour, and so is the input the op is built from, so
+neither the stored rate nor a repost quotes better than the two amounts the operator gave. `REPLAYED` and `ORDER_EXISTS` are answered once by bumping
+`orderNonce`, which changes the commitment and the userOpHash; the orderbook remembers every hash it
+has accepted, so a fresh nonce is the only way past.
+
+## Config
+
+```toml
+[orderbook]
+enabled = true
```
Wizdave97: orderbook should not be optional
```suggestion
```

sdk/packages/simplex/src/config/filler-toml.ts
```diff
@@ -160,6 +161,29 @@ export interface FillerTomlConfig {
...
+	/** GraphQL endpoint. Required when `enabled`. */
+	url?: string
+	/** Off unless set. A filler with no orderbook simply posts nothing. */
+	enabled?: boolean
+	/** TTL written into each posting, in seconds. At least 900, which is the orderbook's floor. */
+	defaultTtlSecs?: number
+	/** How long before a posting expires to repost it, in seconds. */
+	renewMarginSecs?: number
+	/** How often to reconcile local limit orders against the orderbook, in seconds. */
+	reconcileIntervalSecs?: number
```
Wizdave97, line 185: What's this for?

sdk/packages/simplex/src/services/server/UiServer.ts
```diff
@@ -161,6 +164,8 @@ export interface OperatorContext {
 	stop(): Promise<void>
 	activity: Pick<ActivityRecorder, "recent" | "on" | "off" | "record" | "recordWalletTx" | "walletTxs" | "fills" | "orderHistory">
 	bids?: Pick<BidStore, "recent" | "stats" | "byCommitments">
+	/** The operator's limit orders. Absent unless `[orderbook]` is enabled. */
```
Wizdave97: orderbook should not be optional

Wizdave97 (approved):

PR #1271, opened 2026-09-15
[simplex]: Match, reserve and value against limit orders

dharjeezy:
The pieces the filler needs to price from the operator's limit orders rather than from pair curves: which limit order serves an incoming order, how much of it a bid may hold, and what an order is worth in dollars now the curves no longer say. Nothing is wired into `FXFiller` yet, so behaviour is unchanged.

`reserve` is a compare-and-set rather than a read then a write, since two chains bidding against one limit order would otherwise both see room in the gap and between them promise more output than it has. USD values come from the limit orders themselves, and a symbol with no route to a dollar stays unpriced rather than being guessed at, because the value sizes a confirmation wait.

Switching `canFill`, `calculateProfitability` and the valuation pass over follows separately. They reach the pair model through the same places, so that migration is one change rather than four, and it reviews better on its own.

Review fixes follow in the last commit here and one at the tip of the stack. `repost` now acts on the cancel it asked for instead of posting regardless, which is what kept a refused or unanswered cancel from leaving two live entries behind one liability. Cross-chain fills may be partial, since `_fillCrossChain` has kept per-token progress, cleared `_filled` on an under-fill and released escrow proportionally since #980; the comment that justified the old restriction is rewritten. `drawDown` and `release` report a guarded write that did not apply rather than losing it silently, and the draw-down now precedes the release, so a crash between them understates capacity instead of advertising output already paid.

On the matcher: dropping the `offer >= reqOut` precondition and ranking by payout rather than offer is deliberate, and it answers open question 2, since a thin order quoting a wonderful rate can no longer crowd out one that covers the swap. One swap can also draw on several limit orders now, walking levels best first until the ask is covered, because the orderbook already quotes a same-chain swapper the clearing price across every level that can fill together and honouring one of them advertises depth we then refuse to meet. That change lands at the tip of the stack, where the matcher is.

Part of #1267.


sdk/packages/simplex/src/orderbook/limit-orders.ts
```diff
@@ -321,6 +321,70 @@ export class LimitOrderService {
...
+	 * Cancels the current posting and puts the order back at its present size.
+	 *
+	 * Also how a posting is renewed before it expires and how reconciliation
+	 * restores an entry the orderbook dropped, since all three want the same
+	 * thing: whatever the order has left, live on the book again.
+	 */
+	async repost(order: LimitOrder | null): Promise<LimitOrder | null> {
+		if (!order) return null
+		// `UNKNOWN_ORDER` is as good as cancelled: the entry is already gone.
+		if (order.commitment) await this.withdraw(order.commitment as HexString)
```
Wizdave97: **`repost()` throws away the cancel result, so it can leave two live entries.**

`withdraw()` returns a `CancelOrderResult` that is discarded. On `BAD_SIGNATURE`, `SOLVER_MISMATCH`, or — far more likely — a transport error that `signAndCancel` has relabelled `UNKNOWN_ORDER`, the old entry is still live on the book and `post()` immediately adds a second.

That is precisely the failure the decision doc added in #1275 describes: "Posting from both would put two live entries behind one liability, which is the thing section 6 went out of its way to avoid." The cancel-then-post ordering is correct; it just needs to act on the answer.

Suggest: proceed only on `kind === "cancelled"` or a genuine `UNKNOWN_ORDER`, and on anything else leave the row alone with `lastError` set for the next cycle to retry. That depends on separating transport failures from `UNKNOWN_ORDER` (see my comment on #1270).

sdk/packages/simplex/src/data/sqlite/limit-orders.ts
```diff
@@ -170,4 +174,45 @@ export class SqliteLimitOrderStore implements LimitOrderStore {
...
+		const result = this.db
+			.prepare(`
+				UPDATE limit_orders SET reserved = ?, updated_at = datetime('now')
+				WHERE id = ? AND reserved = ? AND status = 'open'
+			`)
+			.run(reserved.toString(), id, order.reserved)
+		return result.changes === 1
+	}
+
+	async drawDown(id: string, amount: string): Promise<LimitOrder | null> {
```
Wizdave97, line 205: `drawDown` and `release` guard on the value they read but ignore `result.changes`, where `reserve` — the same compare-and-swap pattern, three methods up — returns a boolean the caller checks.

better-sqlite3 is synchronous and there's no `await` between the read and the write, so this can only lose a write with a second process on one `bids.db`. But when it does, it loses it silently: a dropped `drawDown` leaves `remaining` too high and simplex goes on advertising output it has already paid out, which is real money and exactly what the `LimitOrderStore` doc comment says this file exists to prevent ("a store that loses writes overcommits real money").

Returning `boolean` from both, or throwing on `changes !== 1`, costs nothing and makes the guarantee checkable.

Separately: #1267 §6 asks for steps 1–3 "in one store transaction". `settleFilledLimitOrder` does `release()` then `settleFill()` → `drawDown()` as two statements, so a crash between them releases the hold without drawing the order down.

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -403,48 +291,50 @@ export class FXFiller implements FillerStrategy {
...
+			// never offers more than it has left even when the wallet holds more.
+			const targetOutput = toRaw(match.payout, outputDecimals)
 
 			// Whether this order may be filled below what the user asked for.
 			//
 			//  - Cross-chain reverts on any under-fill: in ExtrinsicIntents.sol
 			//    `if (solverAmount < totalRequired) revert InvalidInput()` is
-			//    unconditional and per-leg, so bidding a partial there is a
-			//    guaranteed failed fill.
+			//    unconditional, so bidding a partial there is a guaranteed failed fill.
```
Wizdave97: **Partial cross-chain fills should be supported.**

This branch makes `partialEligibleCheap` require `sourceChain === destChain`, so a cross-chain order whose payout falls short of `reqOut` is skipped outright. That should become a partial fill instead — the same rule as same-chain, bounded by `min(offer, remaining − reserved)`.

Flagging the precondition so it isn't a surprise: as the gateway stands today the comment above is accurate. `ExtrinsicIntents.sol::_fillCrossChain` has an unconditional `if (solverAmount < totalRequired) revert InvalidInput()`, sets `_filled[commitment] = msg.sender` for the whole order, and keeps no `_partialFills` accounting on the cross-chain path — that mapping is only written by `IntrinsicIntents.sol`. So a partial cross-chain bid reverts, and this needs the escrow-release side to land first: proportional `RedeemEscrow` on an under-fill, plus cumulative fill tracking keyed by commitment so a second solver can take the remainder.

Once the gateway supports it, the change here is small — drop `sourceChain === destChain` from `partialEligibleCheap` and let the existing `partialEligible()` probe (output calldata, prior partial) do the rest, since both of those constraints apply cross-chain for the same reasons.

sdk/packages/simplex/src/orderbook/matching.ts
```diff
@@ -0,0 +1,103 @@
...
+ * There is no fallback price: an order matching nothing is not filled, which is
+ * the whole point of pricing from the operator's own resting orders rather than
+ * from a curve that always has an answer.
+ *
+ * When several match, the largest offer wins, and a tie goes to the one with
+ * more left. That is the order the orderbook would have served when the swapper
+ * was quoted. Exactly one limit order is ever returned, which is what keeps the
+ * draw-down on a fill a one-to-one piece of bookkeeping.
+ */
+export function matchLimitOrder(
```
Wizdave97, line 54: **One incoming order should be able to draw on several limit orders across price levels.**

The matcher returns exactly one order, and the docstring is explicit that this is to keep the draw-down "a one-to-one piece of bookkeeping". That bookkeeping convenience costs real fills: when the best-priced limit order's depth doesn't cover the swap, simplex skips (cross-chain) or under-fills (same-chain) rather than walking down to the next level, even though it is holding inventory at a price that would clear.

It also puts simplex out of step with what the orderbook already promised the swapper. From the published schema:

- `SwapQuote.rate` — "Same-chain: **the clearing price, the best at which the orders at it or better can fill the trade together**"
- `RouteKind.SAME_CHAIN` — "several solvers may fill parts of one order"
- `RouteLiquidity.maxFillableIn` — "one order cross-chain, **combined same-chain**"

So the orderbook quotes a same-chain swapper against several of our orders combined and we then honour only one of them. The quote the swapper acted on is one we won't meet.

Two shapes, and I think both are wanted:

- **Same-chain:** several bids, one per limit order, walking levels best-first until `reqOut` is covered or depth runs out. Partial fills already work here.
- **Cross-chain:** one bid whose payout aggregates several of our own limit orders. `_filled[commitment]` is per-solver, not per-order, so this needs no contract change at all and works today.

The bookkeeping follows without much trouble: `LimitOrderMatch` becomes a list, `reserve()` runs per limit order (failing atomically as a set, or reserving what it can and sizing the bid to that), and the bid row carries `{limitOrderId, amount}[]` instead of one pair — `claimReservation` and `settleFilledLimitOrder` then fan out over it. Ordering the draw-down deterministically, best price first, keeps it reproducible.

sdk/packages/simplex/src/tests/strategies/fx.payout.test.ts
```diff
@@ -1,37 +1,28 @@
 import { FXFiller, type TradingPair } from "@/strategies/fx"
```
Wizdave97, line 1: This rename breaks `pnpm test:filler` — `package.json:45` still names `src/tests/strategies/fx.curve-payout.test.ts`, which no longer exists, so the script fails outright rather than skipping.

`package.json` isn't otherwise touched in this PR, which is probably why it was missed.

sdk/packages/simplex/docs/ai/changelog/2026-09-15-price-curves-removed.md
```diff
@@ -0,0 +1,41 @@
+# 2026-09-15 — Price curves removed
```
Wizdave97, line 1: Same note as on #1268: this PR deletes no documentation, and several flows now describe subsystems it removes.

- `docs/ai/flows/phantom-probe-curve-value-published-price.md` — curves and the phantom probe, both gone
- `docs/ai/flows/market-setup-and-curve-editing.md` — the curve editor UI this PR deletes
- `docs/ai/flows/venue-pricing-uniswap-v4-funded-pairs.md` — Uniswap V4 venue quotes, removed in #1268
- `docs/ai/flows/operator-market-list-prices.md` and `docs/ai/flows/filling-how-a-fill-amount-is-sized-and-who-else-spends-the-same.md` — both still describe curve-derived pricing

The last one is the one that matters most: it's the flow describing how a fill is sized, which is exactly what this PR replaces. `CLAUDE.md` asks for flows to be updated when their behaviour changes, and this PR changes it completely.

Wizdave97 (commented):
Reviewed against [#1267](https://github.com/polytope-labs/hyperbridge/issues/1267) and `hyperfx-orderbook@main`.

One thing I want to record as **verified rather than flagged**, because it looks like a deviation and isn't: `matchLimitOrder` drops the issue's `offer >= reqOut` precondition and ranks by `payout = min(offer, available)` where §3 says rank by `offer`. Since the reduce takes the maximum payout, if any candidate can cover `reqOut` then the winner does too — no fill is lost, and `fx.ts` enforces the cross-chain all-or-nothing rule at the caller. Ranking by what is actually payable is the better rule and it answers open question 2. Worth saying so in the PR description.

The reservation lifecycle is the strongest part of this PR — `claimReservation` owning the hold from the moment a bid row exists, and the comment explaining why an overstated reservation is the safe way to be wrong, are both right.

Two feature gaps below (partial cross-chain fills, and combining levels) plus three defects.

sdk/packages/simplex/src/strategies/fx.ts
Wizdave97: **Correction: the gateway already supports this.** My precondition above is wrong — I read `ExtrinsicIntents.sol` from this stack's base (`b7f9e04f`), which predates [#980 "[evm]: cross-chain partial fills for IntentGatewayV2"](https://github.com/polytope-labs/hyperbridge/pull/980), merged as `c92b56a1`. Disregard the second paragraph; there is no contract work to wait on.

On current `main`, `_fillCrossChain` mirrors the same-chain path in full:

- cumulative progress per output token in `_partialFills[commitment][outputToken]`, written on the cross-chain path too
- `delete _filled[commitment]` on an under-fill, so the next solver continues the order
- proportional escrow via `_cumulativeReleased(escrowTotal, amountFilled, totalRequired)`, carried in a `RedeemEscrowPartial` (non-finalizing) message, with the monotonic release function guaranteeing the slices sum to exactly the escrowed amount
- `PartialFill` emitted with the same signature `ChainScanner` already consumes

The only remaining bar is output calldata — `if (order.output.call.length > 0 && !isFullyFilled) revert PartialFillNotAllowed()` — which applies to both paths for the same reason, and which `partialEligibleCheap` already tests.

So the simplex change really is just dropping `sourceChain === destChain` from line 334. Everything the existing guard does carries over unchanged:

- the output-calldata test is still correct cross-chain
- `hasExistingPartialFill` → `partialFillsFor` reads `_partialFills` on the **destination** gateway, which is exactly where the cross-chain path now writes it, so the prior-partial probe needs no change either

Two things that do need attention when this lands:

1. **The comment block above this line becomes false.** Lines 325–327 assert "Cross-chain reverts on any under-fill: in `ExtrinsicIntents.sol` `if (solverAmount < totalRequired) revert InvalidInput()` is unconditional" — that branch no longer exists. It needs rewriting when the stack rebases onto `main`, or it will read as a justification for a restriction that is no longer there.
2. **The P&L gate at line 575.** `if (!partialFill && order.fees < executionCost)` exempts partials from the fee gate because "a partial collects NO `order.fees`". Worth re-checking against #980: the full fill is what forwards accumulated fees (`RedeemEscrow` finalizing), so the reasoning likely still holds cross-chain — but it was written when cross-chain partials were impossible, so it has never been exercised on that path.

dharjeezy:
@Wizdave97 Five actioned here in `c56f8646`: `repost` acts on the cancel result, cross-chain fills may be partial now that `_fillCrossChain` keeps per-token progress and releases escrow proportionally, `drawDown` and `release` report a guarded write that did not apply, the draw-down precedes the release so a crash between them understates capacity, and `pnpm test:filler` names a test that exists. Your docs comment was already covered by an earlier pass that deleted the five stale flows.

Combining levels landed at the tip of the stack, in #1275 `c50a81c4`, because that is where the matcher is. `matchLimitOrders` walks levels best payout first and stops once the ask is covered, a bid row carries a list of holds taken all or nothing, and a fill shares out over them in the same order. One regression worth knowing: my first version skipped every candidate when the incoming order asked for zero output, which is how a quote reaches the engine, so `canFill` briefly answered false for same-chain FX. The loop always takes the best candidate now.

On the fee gate for partials: the cross-chain path uses the same `isFullyFilled` logic and only finalises on a full fill, so "a partial collects no `order.fees`" still holds. It does pay the relayer fee for `RedeemEscrowPartial` though, which the partial's profit figure does not net any more than it nets gas. Noted where the figure is computed rather than guessed at.

Wizdave97 (commented):
Re-reviewed at `c56f8646`. Both comments are addressed:

- **`repost()`** now acts on the cancel: it proceeds only on `cancelled` or a genuine `UNKNOWN_ORDER`, and otherwise leaves the row `open` with its commitment intact and the reason on it, so the next cycle retries instead of stacking a second entry. Paired with the `CancelOrderResult` change in #1270, that closes the duplicate-entry path properly.
- **`drawDown` and `release`** throw `LimitOrderWriteError` on a lost CAS rather than returning a stale read.
- `test:filler` points at `fx.payout.test.ts`, and `fbbfbd2a` removes the curve and solver-link notes.

Cross-chain partials are enabled here too, and the `fx.ts` comment block correctly describes `_partialFills` / `RedeemEscrowPartial` from #980. One leftover: the matcher's own docstring still says "a cross-chain order reverts on any under-fill" — that line moved into #1275, so I have commented on it there.

No further comments from me on this one.

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -487,417 +359,178 @@ export class FXFiller implements FillerStrategy {
...
+				if (spreadUsd) sameAssetEdgeUsd = spreadUsd
 			}
 
-			// Clamp is disabled, so a leg can never be clamped — the halt subsystem is
-			// left in place but dormant (always recorded as a clean, unclamped outcome).
-			this.recordOrderOutcome(false, order.id)
+			// What the limit order was willing to pay against what the order asked
+			// for. Handing over less than we were willing to is surplus we keep, and
+			// it is what pays for a partial fill's gas.
+			let payoutSurplusUsd = new Decimal(0)
```
Wizdave97, line 517: **"Handing over less than we were willing to is surplus we keep" is backwards — the contract takes all of it.**

`targetOutput` is not a ceiling simplex stays under. It is passed as the bid amount: `finalOutputAmount = min(targetOutput, effectiveBalance)` → `fillerOutputs` → `cacheService.setFillerOutputs` → `fillOptions.outputs` in `ContractInteractionService:723` → `options.outputs[i].amount`, which is `solverAmount` in the gateway. `buildApprovalAndFillCalldata` approves the same figure, so the transfer clears.

So on the branch this block tests for — `targetOutput > output.amount`, with `alreadyFilled == 0`, so a full fill — the gateway sets `fillAmount = totalRequired` and splits `solverAmount - totalRequired` between the beneficiary and the protocol, debiting the solver the whole `solverAmount`. The surplus goes to the swapper and the protocol. With output calldata, `_splitSurplus` sends 100% of it to the protocol.

Then line 567 does `partialEdgeUsd = sameAssetEdgeUsd.plus(payoutSurplusUsd)`, and line 617 makes that the whole of `totalProfit` for a partial fill. So the donation is booked as the profit that justifies the fill.

**When does it fire?** Whenever the matched limit order prices better than the incoming order asked — which is the normal case simplex is trying to find. `targetOutput = toRaw(payout)` and `payout = min(offer, available)`, where `offer` is what the limit order pays for the full `inputNet` at its own rate. Any time that beats `output.amount`, the difference is given away. The better the operator's price, the more is donated.

**Why nothing catches it:** the `maxOverfillBps` ceiling at line 391 is warn-only, disabled since #964, long before this stack.

The fix is one clamp — bid `min(targetOutput, output.amount)` and keep `targetOutput` for the matching and reservation decisions, where it belongs. Worth checking against #964 first, since the clamp was disabled deliberately and the reason may not be recorded here; but whatever that reason was, it predates limit-order pricing, where `targetOutput` is now a signed rate rather than a curve's upper bound.

Two smaller notes from the same read:

- The `releasedInput` comment at line 497 cites `inputs[i].amount * fillAmount / totalRequired`, which is right, but the value uses `finalOutputAmount` where the contract uses the clamped `fillAmount`. They agree today only because `hasExistingPartialFill` refuses an order someone else has already touched. Worth saying so, since the guard is what makes the arithmetic correct.
- A continuation where `solverAmount >= remaining` completes the order and sweeps the full escrow residue (`escrowedAmount = _orders[commitment][token]`), so the clamp there is favourable rather than harmful. No change needed — just the other half of the answer.

Wizdave97 (commented):
Followed the overfill question down into `IntentGatewayV2` on current `main` (post-[#980](https://github.com/polytope-labs/hyperbridge/pull/980)). Short answer: **a partial fill cannot overpay, but a full fill with `targetOutput > output.amount` donates the entire excess** — and the comment below says the opposite.

Both `IntrinsicIntents._fillSameChain` and `ExtrinsicIntents._fillCrossChain` share one branch:

```solidity
if (alreadyFilled == 0 && solverAmount > totalRequired) {
    fillAmount = totalRequired;
    (protocolShare, beneficiaryShare) = _splitSurplus(solverAmount - totalRequired, order.output.call.length > 0);
} else {
    fillAmount = solverAmount > remaining ? remaining : solverAmount;
}
```

The `else` branch is every partial fill — a first fill short of the ask (`solverAmount < totalRequired`) and any continuation (`alreadyFilled > 0`). Both clamp to `min(solverAmount, remaining)` and pull only `fillAmount`, so an over-sized `options.outputs[i].amount` on a partial costs nothing; the excess is ignored, not charged. The surplus branch is unreachable on a partial by construction, since it requires `solverAmount > totalRequired`, which is a full fill.

The other branch is the problem, and `_splitSurplus` distributes **all** of it:

```solidity
if (hasOutputCall) return (dust, 0);
protocolShare = (dust * _params.surplusShareBps) / 10_000;
beneficiaryShare = dust - protocolShare;
```

Then `safeTransferFrom(msg.sender, beneficiary, fillAmount + beneficiaryShare)` and `safeTransferFrom(msg.sender, address(this), protocolShare)` — the solver is debited exactly `solverAmount`. Nothing comes back.

dharjeezy:
The overfill fix landed at the tip, in #1275 `104c939e`, since the clamp interacts with the matcher change there. Bid is `min(offered, output.amount)` now, and `payoutSurplusUsd` describes margin actually kept rather than a donation.

Your `releasedInput` note is in as a comment: the arithmetic agrees with the contract only because the prior-partial guard keeps `alreadyFilled` at zero and the bid is never above the ask, and both halves are now stated where the value is computed.

sdk/packages/simplex/src/cli/init/steps/strategies.ts
```diff
@@ -25,9 +18,6 @@ const SYMBOL_OPTIONS = [
 const STABLE_SWAP_BID = [{ amount: "0", price: "1.001" }]
```
Wizdave97: This should be removed

sdk/packages/simplex/src/cli/init/steps/strategies.ts
```diff
@@ -25,9 +18,6 @@ const SYMBOL_OPTIONS = [
...
 const STABLE_SWAP_ASK = [{ amount: "0", price: "0.999" }]
 
-const belowParPrice = (value: number): string | undefined =>
-	value > 0 && value < 1 ? undefined : "Must be between 0 and 1 (exclusive) — the gap to 1 is your spread"
-
 export async function stepStrategies(state: WizardState, prefill?: Prefill): Promise<void> {
```
Wizdave97: For this cli, it's current architecture can't work with our limit orders paradigm, we should remove it.

Wizdave97 (approved):

PR #1273, opened 2026-09-15
feat(simplex-desktop): add lifecycle and OS integration

### LINKED ISSUE #1243: [simplex] Desktop lifecycle and OS integration

ddboy19912:
## Summary

Closes #1243.

Adds the OS-facing lifecycle layer for Simplex Desktop. The Electron app now remains accessible from the system tray, supervises the detached solver, exposes explicit pause/stop/restart controls, supports opt-in login startup, prevents sleep while actively filling, and clearly distinguishes quitting the app from stopping the solver.

This PR is stacked on the unmerged #1242 branch. After #1242 merges, this PR should be retargeted or rebased onto `main`.

## What changed

- Added native system tray support for macOS, Windows, and Linux.
- Derived tray icons from the existing Simplex PWA logo.
- Added solver states for starting, setup, running, paused, stopped, and unreachable.
- Added quick pause, resume, stop, restart, and show-window actions.
- Made closing the window hide it to the tray without stopping the solver.
- Added separate **Quit Simplex (solver keeps filling)** and **Stop solver and quit** actions.
- Added opt-in launch-at-login for installed applications.
- Added socket-based solver supervision and crash notifications.
- Added `powerSaveBlocker` while the solver is actively filling.
- Added native About, update status, data-directory, current-log, and quit menu actions.
- Added rotating on-disk solver logs.
- Added cross-platform lifecycle and Electron E2E coverage.
- Documented crash-reporting, uninstall-data, Linux tray, and continuous-uptime decisions.

## Task checklist

- [x] System tray icon shows running, paused, stopped, starting, setup, and unreachable states.
- [x] Pause, resume, stop, restart, and show-window actions are available from the tray context menu.
- [x] Closing the window hides it to the tray and does not stop the solver.
- [x] Added clearly labelled **Quit Simplex (solver keeps filling)** and **Stop solver and quit** actions.
- [x] Added opt-in launch-at-login for the installed app.
- [x] macOS and Windows use Electron login-item settings.
- [x] Linux uses an equivalent per-user XDG autostart entry.
- [x] The app uses Electron’s single-instance lock while the socket remains the solver-instance authority.
- [x] A second app launch focuses the existing window and does not spawn another solver.
- [x] The app supervises the solver through periodic private-socket health checks.
- [x] A crashed or unreachable solver is surfaced through native state, menus, window title, and notification.
- [x] Operators can restart a stopped or crashed solver.
- [x] `powerSaveBlocker` is active while the solver is filling and released while paused or stopped.
- [x] Sleep-prevention state is visible in native menus.
- [x] Native menus include About, Check for Updates, Open Data Directory, Open Current Log, and both quit actions.
- [x] Detached solver output is written to rotating on-disk logs.
- [x] Crash-reporting decision documented: no remote crash reporting is enabled.
- [x] Config contents and key material are not uploaded by the desktop shell.
- [x] The uninstall contract preserves the desktop data directory and operator records.
- [x] Every Linux action is reachable from the context menu; no action depends on tray-click support.
- [x] Desktop uptime limitations and the VPS/container path for continuous operation are documented.

## Acceptance criteria

- [x] Closing the window leaves the solver filling and the tray reflects its state.
- [x] Quitting the app leaves the solver filling; relaunching attaches to it.
- [x] Launching a second copy focuses the first and never spawns a second solver.
- [x] A crashed solver becomes visible in the desktop UI within seconds.
- [x] The machine is prevented from sleeping while the solver is actively filling.
- [x] Uninstalling preserves operator data by contract.

## Verification

- [x] Desktop typecheck passed.
- [x] Desktop lint passed.
- [x] Desktop production build passed.
- [x] 51 desktop unit and integration tests passed.
- [x] 7 real Electron E2E scenarios passed.
- [x] 32 solver log-rotation tests passed.
- [x] E2E verifies window-close and app-quit solver survival.
- [x] E2E verifies relaunch attachment and second-instance behavior.
- [x] E2E verifies pause/resume, crash visibility, and restart.
- [x] E2E verifies sleep prevention and native lifecycle state.
- [x] E2E verifies disk logging and that no desktop TCP listener is opened.

## Platform notes

On macOS, the tray appears in the menu bar. On Windows, it appears in the notification area. Linux uses Electron’s StatusNotifierItem/Gtk integration; GNOME may require an AppIndicator extension. All functionality remains accessible through the context menu.

Launch-at-login applies to the Electron app, not directly to the solver. Quitting Electron intentionally leaves the detached solver running, but it also releases Electron’s sleep blocker. Operators requiring continuous uptime should use the Simplex container on a VPS.


ddboy19912:
Screenshot 2026-09-15 at 16 25 27
Simplex in the app tray

seunlanlege (commented):
Thanks for the screenshot. The tray menu matches the code, and splitting close, quit and stop reads clearly.

Two things I'd fix before merging: the Restart race on `solver-supervisor.ts` (it can run two fillers on the same signer), and the application menu breaking copy/paste on macOS. The rest are smaller.

sdk/packages/simplex-desktop/src/solver-supervisor.ts
```diff
@@ -0,0 +1,149 @@
...
+		request.setTimeout(timeoutMs, () => request.destroy(new Error(`Simplex ${path} request timed out`)))
+		request.on("error", reject)
+		request.end()
+	})
+}
+
+/** The state the native shell displays, read only through the private socket. */
+export async function probeSolverStatus(socketPath: string): Promise<SolverStatus> {
+	const health = await probeHealth(socketPath)
+	if (health.state === "spawnable") return { state: "stopped", detail: health.reason }
```
seunlanlege, line 57: **Blocking: Restart can start a second solver while the first is still shutting down.**

The solver's shutdown path closes and deletes the socket first ([simplex.ts:411](https://github.com/polytope-labs/hyperbridge/blob/126044f0a5932fad4bf3d7754ed572d06dcf2226/sdk/packages/simplex/src/bin/simplex.ts#L411)). Only after that does it drain in-flight fills and redeem vault positions on-chain ([boot.ts:717](https://github.com/polytope-labs/hyperbridge/blob/126044f0a5932fad4bf3d7754ed572d06dcf2226/sdk/packages/simplex/src/core/boot.ts#L717)).

This line reads the missing socket as `stopped`. So within one poll of a Stop, the tray shows Stopped and enables Restart, while the old process is still sending transactions.

Restart then spawns a new solver. A configured solver starts filling before it binds the socket, and it keeps filling if the bind fails. So the socket was never a lock on filling. The result is two fillers on the same signer and data directory, with nonce collisions and duplicate bids.

The E2E fixture's `stop` just closes the server, so the test doesn't cover the drain window.

Suggested fix: keep the socket bound until the process exits, and have `/health` report `stopping` while it drains. Keep Restart disabled until the socket is gone.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -1,58 +1,278 @@
...
 		mainWindow.focus()
 		return
 	}
-	if (!daemonReady) daemonReady = prepareDaemon()
-	await daemonReady
-	if (!sessionSecured) {
-		installSessionSecurity(session.defaultSession)
-		sessionSecured = true
+	if (!supervisor) return
+	if (supervisor.status.state === "stopped" || supervisor.status.state === "unreachable") {
```
seunlanlege, line 284: Same race as the comment on `solver-supervisor.ts`. When no window exists (a hidden login launch, for example), "Show Simplex" or a dock click restarts a stopped solver here without asking.

That can hit the drain window right after a Stop. It also contradicts the README's "Restart is operator-controlled". I'd drop the auto-start here and let the window show the stopped state, with Restart in the menu.

sdk/packages/simplex-desktop/src/tray-menu.ts
```diff
@@ -0,0 +1,143 @@
...
+			click: run(actions.stopAndQuit),
+		},
+	]
+}
+
+function canStopSolver(status: SolverStatus): boolean {
+	return status.state === "running" || status.state === "paused"
+}
+
+export function buildApplicationMenuTemplate(
```
seunlanlege, line 130: **Blocking: this menu breaks copy and paste on macOS.**

Setting the application menu replaces Electron's default menu, which includes Edit and Window. On macOS, Cmd+C/V/X/A/Z only reach the web page through Edit menu items. Without them, paste stops working in the setup wizard, which is where operators paste RPC URLs and keys.

Cmd+Q, Cmd+W, Cmd+H and Cmd+M stop working too.

Suggested fix: add `{ role: "editMenu" }` and `{ role: "windowMenu" }` next to the Simplex submenu, and give the quit item `accelerator: "CmdOrCtrl+Q"`. You can confirm the bug on a Mac by pasting into a setup field.

sdk/packages/simplex-desktop/src/tray-menu.ts
```diff
@@ -0,0 +1,143 @@
...
+		{
+			id: "stop-and-quit",
+			label: "Stop solver and quit",
+			enabled: canStopSolver(model.status),
+			click: run(actions.stopAndQuit),
+		},
+	]
+}
+
+function canStopSolver(status: SolverStatus): boolean {
```
seunlanlege, line 120: **During setup, nothing in the app can stop the solver.** Your screenshot shows this state.

Both stop actions are disabled in `setup`. That matches the server, where `/api/stop` returns 409 until the solver is configured ([UiServer.ts:1119](https://github.com/polytope-labs/hyperbridge/blob/126044f0a5932fad4bf3d7754ed572d06dcf2226/sdk/packages/simplex/src/services/server/UiServer.ts#L1119)).

So someone who opens the app and quits before finishing setup leaves a background Node process running until reboot. The label "Quit Simplex (solver keeps filling)" on line 98 is also wrong in this state, since nothing is filling.

Suggested fix: allow `/api/stop` during setup, except while save-and-start is booting the filler, and enable stop for `setup` here. Use a plain "Quit Simplex" label whenever the solver isn't running or paused.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -1,58 +1,278 @@
...
+					: "The solver is no longer running. Restart it from the Simplex menu.",
+		})
+		notification.on("click", () => void safeShowWindow())
+		notification.show()
+	} catch (error) {
+		console.error(`Simplex could not display its solver notification: ${errorMessage(error)}`)
+	}
+}
+
+function onSolverStatusChanged(next: SolverStatus, previous: SolverStatus): void {
```
seunlanlege, line 166: **A deliberate stop shows a crash notification.**

This notifies on any change from a live state to `stopped`. The tray's Stop, "Stop solver and quit", and the dashboard's Stop button all go through it, so each one pops "Simplex solver stopped… Restart it from the Simplex menu."

The `stopping` state suggested on `solver-supervisor.ts` would fix this too. A change from `stopping` to `stopped` was intentional, so it can skip the notification.

sdk/packages/simplex-desktop/src/solver-supervisor.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export type SolverAction = "pause" | "resume" | "stop"
+
+const MAX_RESPONSE_BYTES = 64 * 1024
+
+function socketRequest(
+	socketPath: string,
+	path: string,
+	method: "GET" | "POST",
+	timeoutMs = 1_000,
```
seunlanlege, line 21: **One slow poll counts as a crash.**

`/health` and `/api/status` both time out after 1 second. A single slow reply switches the state to `unreachable`. That releases the sleep blocker, fires the notification and enables Restart. The next poll switches it back.

A busy solver can do this repeatedly. I'd require 2–3 failures in a row before leaving `running` or `paused`.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -69,30 +290,119 @@ async function createWindow(): Promise<void> {
...
+	} catch (error) {
+		reportActionError("Simplex could not open its window", error)
+	}
+}
+
+function createTray(): void {
+	if (!supervisor) return
+	tray =
+		process.platform === "win32"
+			? new Tray(trayImage(supervisor.status), TRAY_GUID)
```
seunlanlege: **A fixed tray GUID can stop the app from starting on Windows.**

Electron's docs say that for an unsigned executable, the GUID is tied to the executable's path, and a different path makes tray creation fail. Two dev checkouts, or an install that moves, would hit this.

Tray creation runs inside `prepareDesktop`, so the error makes the app quit with "Simplex could not start". Builds aren't signed yet, so I'd drop the GUID for now.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -1,58 +1,278 @@
...
+	}
+}
+
+function trayImage(status: SolverStatus) {
+	if (!trayIconDirectory) throw new Error("The Simplex tray icons are not initialized")
+	const path = trayIconPath(trayIconDirectory, status)
+	if (!existsSync(path)) throw new Error(`The Simplex tray icon is missing: ${path}`)
+	const image = nativeImage.createFromPath(path)
+	if (image.isEmpty()) throw new Error("Electron could not render the Simplex tray icon")
+	const resized = image.resize({ width: process.platform === "darwin" ? 18 : 24 })
```
seunlanlege: Nit: shrinking the 64px image to 18px gives macOS a single 1x image, so the icon looks blurry on Retina screens. Shipping `runningTemplate.png` at 18px plus `runningTemplate@2x.png` at 36px lets `createFromPath` pick the right one.

At 18px the status badge is also only about 5px. The `setup` ring and `running` dot look almost the same in the template set.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -1,58 +1,278 @@
...
+const menuActions: DesktopMenuActions = {
+	showWindow: () => safeShowWindow(),
+	togglePause: () => togglePause(),
+	stopSolver: async () => {
+		await stopSolver()
+	},
+	restartSolver: () => restartSolver(),
+	toggleLoginItem: () => toggleLoginItem(),
+	showAbout: () => app.showAboutPanel(),
+	checkForUpdates: async () => {
```
seunlanlege, line 112: Nit: this item is enabled but only opens a dialog saying updates aren't set up. I'd hide it until an updater exists.

sdk/packages/simplex-desktop/src/login-item.ts
```diff
@@ -0,0 +1,140 @@
...
+			const existing = readFileSync(this.autostartPath, "utf8")
+			if (!existing.includes(LINUX_AUTOSTART_MARKER)) {
+				throw new Error(`Refusing to replace an unmanaged autostart file at ${this.autostartPath}`)
+			}
+		}
+
+		mkdirSync(dirname(this.autostartPath), { recursive: true })
+		const temporary = `${this.autostartPath}.${process.pid}.tmp`
+		try {
+			writeFileSync(temporary, linuxAutostartEntry(this.options.executable), { encoding: "utf8", mode: 0o600 })
```
seunlanlege, line 130: Nit for when packaging lands: for an AppImage, `process.execPath` is a temporary mount path that changes on each launch, so this autostart entry breaks after reboot. Use `process.env.APPIMAGE` when it's set.

seunlanlege (approved):

PR #1275, opened 2026-09-16
[simplex]: Keep posted limit orders alive with heartbeats, renewal and reconciliation

dharjeezy:
Section 7 of #1267. A posting is not something the orderbook keeps for you: it suspends a solver it has not heard from, an entry expires on its own clock, and a crash between two requests leaves one side holding what the other does not. `LimitOrderLifecycle` runs the three jobs that keep the two copies together, and `bootFiller` starts and stops it with the filler.

`heartbeat()` signs `Heartbeat(solver, timestamp)` on the same domain as `CancelOrder` and goes out on half the interval `serverInfo` asks for. It stays quiet until something is posted, because the orderbook only knows a solver whose order it has accepted, and it fires immediately on a posting that comes back `surfaced: false` since that is the orderbook saying the solver is still suspended.

`renewExpiring()` replaces an entry before it expires rather than extending it, which is not possible: the op carries its own deadline and the orderbook remembers every op hash it has taken, so renewal is a fresh op on a new nonce. `reconcile()` pages through the solver's orders and cancels an entry nothing local owns, reposts an order whose entry has gone, and leaves a `resized` or `backed: false` entry alone with `UNDER_FUNDED` on the row, since reposting it would only have it cut down again. A row that went to `resizing` in the last two minutes is skipped, because that is a repost in flight rather than a crash, and posting a second entry for one liability is the thing section 6 went out of its way to avoid.

An orderbook that is down at boot does not stop the filler starting. It prices from the local limit orders either way, a pass that throws is logged and its clock carries on, and a tick is skipped while the previous pass is still running.

`orderbook.renewMarginSecs` and `orderbook.reconcileIntervalSecs` were already accepted and validated but nothing read them. They now default to 120 and 300 seconds.

The last acceptance criterion on #1267 is also covered here. `src/tests/fixtures/orderbook-userops.json` is a copy of the orderbook's own golden vectors, 14 signed `fillOrder` UserOps whose descriptions each name the verdict they expect. `posted-userop.test.ts` reads an op the way the orderbook does and returns the refusal it would earn, all 14 vectors get the verdict they name, and the op `prepareLimitOrderUserOp` builds then goes through the same function and is accepted. Checking our op against a hand-written list of rules would only ever prove we agree with ourselves.

Two checks were added on top of that, and each found a bug the rest of the suite could not see, because nothing else here reads a query or talks to a server. `schema.test.ts` validates every document the client sends against the orderbook's published `schema.graphql`: `submitOrder` was asking for `code` on both `OrderRejected` and `OrderSubmissionFailed`, which return different enums, so a server would have refused the whole mutation and no order would ever have been posted. `orderbook.live.test.ts` takes one limit order through a real orderbook, and found that `CancelOrder` was signed without the `solver` field the server's struct carries, so every cancel came back `SOLVER_MISMATCH`: withdrawing left the entry live, and a resize would have put a second entry behind the same liability.

It also found that `backed: false` does not mean under-funded. It is false before any balance cycle has read the order too, and the order surfaces at its full quoted size until one does, so reconciliation was stamping `UNDER_FUNDED` on healthy postings seconds after they went up. `PostedOrder` now carries `validatedAt` and only a `backed: false` a cycle actually decided counts.

The live test is skipped unless `HYPERFX_ORDERBOOK_URL` or `HYPERFX_ORDERBOOK_BIN` says where to find a server, so it does not run in CI. It needs no chain: the config it writes runs the server the way its own `config.dev.toml` does, with the protocol fee a constant and the validation cycle reading nothing.

A review of the whole stack then turned up four defects, fixed in the last commit. An expired limit order was read only by the matcher, so it stopped matching while renewal kept its posting alive and reconciliation put it back, advertising depth the filler would always refuse; `expireStale()` now withdraws it and the row moves to a new `expired` status. The matcher ranked on the offer rather than on `min(offer, remaining - reserved)`, so an order quoting a good rate with nothing behind it beat one that could cover the swap, and a pre-filter on the offer made a same-chain partial fill impossible whenever the price rather than the size fell short. And the catch around `executeOrder` released a reservation directly even after the bid row had taken ownership of it, which gave the same hold back twice when a listener threw.

The last commit closes the rest of that review. Same-asset quoting comes back as a local limit order: no book trades a symbol against itself, so one that takes in and pays out the same symbol is stored and priced here and never sent anywhere, and it has to sit at or below par, which is the rule the curves expressed as ask-only and priced under 1. An operator whose `[[pairs]]` still carries curve keys is warned at boot that their prices are no longer read, since the config parses cleanly and quotes nothing. A cross-chain order nothing can value in dollars now waits the deepest confirmation its curve allows rather than the shallowest. `expiresAt` is validated on the way in, and the amount helpers handle a token with more than 18 decimals instead of raising `RangeError`.

A second read of the stack found one more, fixed in the last commit. `setPosting` and `setStatus` wrote whatever the caller asked, so a posting already in flight when the operator cancelled the order, or when the expiry sweep took it down, wrote the row back to `open` with a fresh commitment and a live entry: the cancel was undone and the order matched swaps again. Both writes now take the statuses the row must still hold and answer null when it has moved on, and `post` withdraws the entry the orderbook just accepted when its write does not apply.

One change outside the feature: a test in `log-store.test.ts` starts two `LogStore`s in the same tick and assumed the second would lose the `wx` race, but two concurrent `O_EXCL` opens finish in whichever order the pool serves them, so it now waits for the first launch file before starting the second. The stale `pairs.setCurve` example in the `src/index.ts` module doc, left behind by the curve removal, is replaced with a limit order.

Stacked on #1271. Review the last seven commits only, or wait for #1271 to merge and this rebases onto main.

The last commit answers the review here. The grace period that keeps reconciliation from posting over a repost in flight only covered `resizing` rows, but `create` posts after its insert and `renewExpiring` posts after its cancel, and both leave a live row with no entry to find. The guard is on `updatedAt` alone now, for any live row whose entry is missing, and `repost` marks the row for the duration so the renewal path says what it is doing. The decision note is rewritten around postings rather than resizes.

The latest round answers the second review. The bid is clamped to the ask, since `targetOutput` is `solverAmount` and the gateway hands everything above `totalRequired` to the beneficiary and the protocol while debiting the solver the lot; `payoutSurplusUsd` was booking that donation as the profit that justified the fill. `offer >= requestedOutput` is back as a participation condition, because proportional escrow release means every fill settles at `T / I` whatever fraction it covers, so an order below the ask would pay above the rate it signed. A fill settles its holds in one store transaction rather than three writes. The pinned schema is refreshed, token decimals now come from the orderbook's own registry instead of an RPC read per token, and a workflow diffs the pin against the server so it cannot drift silently again.

Two decisions in there are mine rather than the reviewer's, and both are worth disagreeing with if the reasoning is wrong. Orders are drawn on **tightest first**, not best price first: the rate is the swapper's whichever funds the slice, so the choice only decides what stays resting, and a generous order qualifies for every swap a tighter one does plus swaps it cannot serve, at the same cost per unit. And a swap draws on several orders through **one bid with slices**, not one bid per order: clamping to the ask means the total is the ask, so a single bid covers it, which removes the shared nonce key, the head-of-line blocking and the dependency on #1259 entirely.

`abb922e0` answers the newest comment: no path reposts an expired limit order. The check now sits in `repost`, which every posting goes through, rather than in the sweep, since reconciliation runs on its own clock and a late-settling fill never swept at all. An expired order reaching it is withdrawn and retired the way the sweep does it, through the same function. Reconciliation also counts a repost only when a posting actually landed, since a retired order was being reported as one.

Part of #1267.


sdk/packages/simplex/src/orderbook/limit-orders.ts
```diff
@@ -377,6 +449,174 @@ export class LimitOrderService {
...
+				report.underFunded++
+			}
+		}
+
+		for (const order of live) {
+			if (isLocal(order)) continue
+			if (order.commitment && found.has(order.commitment.toLowerCase())) continue
+			// A row that only just went to `resizing` has a repost in flight, and
+			// posting a second entry for one liability is worse than waiting a cycle.
+			if (order.status === "resizing" && sinceMs(order.updatedAt) < RESIZE_GRACE_MS) continue
```
Wizdave97: **The grace period covers `settleFill` but not the other two paths that repost.**

`2026-09-16-reconciliation-waits-out-a-resize-rather-than-locking-against-it.md` reasons this through correctly for resizes. The same window exists twice more, unguarded, because in both the row stays `open`:

1. **`create()`** — `store.create()` inserts with `status = 'open'` and `commitment = NULL`, *then* posts. A reconcile tick in that gap sees a row that should have an entry and no entry listed, so it reposts. Both postings land; `setPosting` records whichever finishes second, and the other is orphaned.
2. **`renewExpiring()` → `repost()`** — the old commitment is cancelled and the row keeps `status: "open"` across the cancel-then-post gap. Reconcile finds the old commitment absent from `entries`, the `resizing` check doesn't apply, and it posts a third entry.

Both produce exactly what the decision doc set out to prevent: two live entries behind one liability.

The fix is the one the doc already argues for. `updatedAt` is written on insert and on every `setStatus`/`setPosting`, so the evidence is there — widening the guard to *any* row whose commitment is missing and whose `updatedAt` is inside `RESIZE_GRACE_MS` covers all three paths with the same reasoning and no new column. Renewal is the awkward one, since the row's `updatedAt` isn't touched when `repost` starts; moving it to `resizing` for the duration (it is, after all, the same cancel-and-repost) would fold it into the existing rule.

Worth extending the decision doc to say the rule is about reposts in general rather than resizes specifically.

Wizdave97 (commented):
Reviewed against [#1267](https://github.com/polytope-labs/hyperbridge/issues/1267) §7 and `hyperfx-orderbook@main`.

The lifecycle itself is solid: three independent clocks, a `running` guard so a slow orderbook can't overlap passes, failures confined to their own clock, and boot not blocking on the first reconcile. Halving `heartbeatIntervalSecs` so one lost request isn't a suspension is the right margin. `expireStale` before `renewExpiring` is the right order and the comment says why.

Two checks I ran that came back clean and are worth recording:

- `myOrders` selects no `status`, and that's correct — `push_order_filter` with `status: None` emits no status clause, so the page covers `ACTIVE` and `SUSPENDED` alike (`crates/store/src/orders.rs::solver_order_pages`). Reconciliation sees suspended entries and won't repost duplicates for them.
- `underFunds()` gating `backed === false` on `validatedAt` being present is right: per the schema `backed` is also false before any cycle has read a balance, and an order surfaces at its quoted size meanwhile. Reporting that as under-funded would have been noise on every fresh posting.

One comment below. Note that the two other findings I had here — `repost()` discarding the cancel result, and the transport-error/`UNKNOWN_ORDER` conflation that makes it fire — belong to #1271 and #1270 respectively, where those lines were introduced. They both bear directly on the decision doc added in this PR.

sdk/packages/simplex/src/orderbook/matching.ts
```diff
@@ -69,13 +85,43 @@ export function matchLimitOrder(
...
-	}, null)
+	// The best candidate is always taken, then levels are added while the ask is
+	// still short. An order asking for nothing, which is how a quote reaches the
+	// engine, still draws on the one that prices it.
+	const taken: LimitOrderMatch[] = []
+	let covered = 0n
+	for (const candidate of candidates) {
+		taken.push(candidate)
+		covered += candidate.payout
+		if (covered >= incoming.requestedOutput) break
```
Wizdave97: **Summing `payout` across orders pays for the same input several times over.**

Every candidate's `offer` is computed against the *whole* `inputNet` — `offerFor({ inputAmount: incoming.inputNet, price: order.price })`, a few lines up. So `offer_i` is what order *i* would pay if it served the entire swap by itself. Summing those across orders bills one input to several orders at once, and `fx.ts` hands the sum straight to `targetOutput`.

Run against this file's own code, book USDC-cNGN, both tokens 6 decimals, one same-chain swap of 1000 USDC asking for 1,450,000 cNGN:

| limit order | price | available | offer (on the full 1000) | payout |
|---|---|---|---|---|
| L1 | 1500 | 600,000 | 1,500,000 | 600,000 |
| L2 | 1400 | 900,000 | 1,400,000 | 900,000 |

`combinedPayout` = 1,500,000 cNGN for 1000 USDC — a blended rate of 1500, which is the rate of the *worst* order in the set, on an input only one of them could have absorbed.

What the orderbook quoted that swapper is 1,400,000. From `crates/core/src/quote.rs::combined`, which is the function this needs to mirror:

1. Orders are walked **best price first** — `Side::cmp_best_first`, bids descending, asks ascending — not best payout.
2. Capacity accumulates until the first order where `capacity >= needs(order.price)`, with `needs(p) = output_for(side, amount_in, p)`: what the whole input demands at price `p`.
3. That order's price is the **clearing price**. The total output is `needs(clearing_price)` — computed once, at one price, on the whole input.
4. Only then are fills allocated as slices of that single total, each capped at its order's capacity.

Here: capacity after L1 = 600,000 < needs(1500) = 1,500,000, so not cleared; capacity after L2 = 1,500,000 >= needs(1400) = 1,400,000, so the clearing price is **1400** and the output is **1,400,000**, filled as L1 600,000 + L2 800,000. Simplex pays 1,500,000 — **100,000 cNGN over, 7.1%** — and the overfill ceiling at `fx.ts:391` is warn-only with the clamp disabled, so it goes out.

Two separate defects behind it:

- **No clearing price.** `covered += candidate.payout` is a sum of independent full-input offers, where the quantity wanted is `output_for(side, inputNet, clearingPrice)`. The `payout` cap belongs in the *allocation* step (step 4), not in the total.
- **Sorted by payout, not price.** `byPayoutThenAvailable` will put a worse-priced order ahead of a better-priced one purely because it has more depth, so even the clearing-price search would start from the wrong end. For the single-order case payout-ranking was right and I said so last time; once orders combine, price is the key.

`min(offer, available)` still does the work it was added for — it just belongs to allocation rather than to the total.

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -302,11 +302,18 @@ export class FXFiller implements FillerStrategy {
...
 				this.logger.info({ orderId: order.id }, "Skipping order: no limit order matches it")
 				return 0
 			}
+			// The best-priced match speaks for the set wherever one order has to be
+			// named: they all trade the same pair on the same chain, and the matcher
+			// ranked them, so this is the one the swap draws on first.
+			const match = matches[0]
+			// What the set pays together, which is what the orderbook quoted the
+			// swapper when it priced their swap across levels.
+			const combinedPayout = matches.reduce((total, candidate) => total + candidate.payout, 0n)
```
Wizdave97: This is where the over-payment reaches the wire — see my comment on `matching.ts`. `combinedPayout` is a sum of per-order full-input offers, so `targetOutput` at line 328 is the amount actually bid.

Worth noting what stops it being caught downstream: the overfill ceiling at line 391 is explicitly warn-only ("the clamp is DISABLED, so the filler pays the full matched amount"), and the surplus accounting at 529 books the excess as `payoutSurplusUsd` — which *credits* a partial fill's P&L with the overpay rather than flagging it. So an over-priced multi-order fill can read as more profitable than it is.

Once the matcher serves a clearing price, `targetOutput` should be `output_for(inputNet, clearingRate)` and the reservations become slices of that total — which also keeps `holdAll` reserving exactly what is paid, as it does today.

sdk/packages/simplex/src/orderbook/matching.ts
```diff
@@ -35,28 +35,44 @@ export interface LimitOrderMatch {
...
+ * first and stop as soon as the ask is covered, so a swap one order covers still
+ * draws on one.
+ *
+ * What each pays is `min(offer, remaining - reserved)`, not the offer alone: an
+ * order quoting a wonderful rate with almost nothing behind it would otherwise
+ * crowd out one that can actually cover the swap.
+ *
+ * An order whose offer falls short of what the swapper asked for is still a
+ * match. Whether a shortfall can be filled at all is the caller's rule, not the
+ * matcher's: a cross-chain order reverts on any under-fill, while a same-chain
```
Wizdave97: This is now false twice over, and it is load-bearing — it is the stated reason the matcher leaves the shortfall rule to the caller.

- The gateway supports cross-chain partial fills since [#980](https://github.com/polytope-labs/hyperbridge/pull/980) (`c92b56a1`), which `c56f8646` already acted on in `fx.ts` — `partialEligibleCheap` no longer tests `sourceChain === destChain`, and the comment there was updated. This one was missed.
- The orderbook dropped the distinction too. On current `main`, `RouteKind` reads "Whether a route's payment crosses chains, which decides which orders serve it. **Either kind is filled by one solver or by several together**", `CROSS_CHAIN` is now just "only orders declaring the source chain serve it", and `SwapQuote.rate` is unconditionally the clearing price. `quote.rs` says the declaration "is all the two kinds of route differ by".

So the only real asymmetry left is which orders *serve* a route, which `serves()` already handles correctly via `acceptedSources`. The shortfall rule is now the same on both paths.

Wizdave97 (commented):
Re-reviewed at `5abbf901` against `hyperfx-orderbook@main` (now `9121a06`). Every one of the 15 comments from the last pass is addressed, and several are fixed better than I suggested — `POST_GRACE_MS` generalising the reconciliation guard to all three posting paths rather than just the two I named, `holdAll` taking every hold or none, and `5d06ef80` catching a late posting resurrecting a cancelled row, which I had missed. `drawDown`/`release` now throw on a lost CAS, `orderAt` handles `ORDER_EXISTS` properly, transport failures are their own kind, and `serverInfo.chains` is validated. 279 tests pass locally.

The multi-order work has a **correctness bug in the pricing**, though, and the orderbook has moved under the pinned schema. Three comments below.

sdk/packages/simplex/src/tests/fixtures/orderbook-schema.graphql
```diff
@@ -0,0 +1,989 @@
+scalar Address
```
Wizdave97, line 1: **This fixture has drifted from the orderbook.** It matched `hyperfx-orderbook@main` exactly when I checked at `f944f38`; at `9121a06` it no longer does, so `schema.test.ts` is currently validating against a schema the server does not serve — which is the one failure mode the test exists to prevent.

What changed:

- **`SwapQuote.priceBucket` was removed.** A breaking change, though nothing here selects it.
- **`Query.chains: [ChainInfo!]!` added**, with `ChainInfo { id, name, tokens: [ChainToken!]! }` and `ChainToken { symbol, decimals }`. Distinct from `serverInfo.chains: [String!]!`, which still exists and which `validate()` correctly uses.
- **Quote semantics rewritten** for combined fills on both route kinds — see my comment on `matching.ts:66`.

The refresh command is already in this test's docstring. Two things worth doing beyond running it:

1. The new `chains` query carries per-chain token decimals, which is what `getTokenDecimals` currently costs an RPC round trip per token per fill chain to learn. `buildUserOp` needs both sides' decimals on every post and repost, so reading them from `limits()` — already cached for 5 minutes — would drop those reads entirely.
2. A pinned copy will drift again. Since the refresh is one `gh api` call, a CI job that re-fetches and diffs would turn silent drift into a failing build; short of that, a line in the changelog note saying the fixture is a pin and how to refresh it would at least make the staleness discoverable.

sdk/packages/simplex/src/orderbook/matching.ts
```diff
@@ -69,13 +85,43 @@ export function matchLimitOrder(
...
+	return matchLimitOrders(orders, incoming, resolve, now)[0] ?? null
+}
+
+/**
+ * Best payout first, ties to the one with more left, then by id.
+ *
+ * Deterministic to the last comparison, because the draw-down after a fill walks
+ * these orders in the same sequence and has to reach the same answer.
+ */
+function byPayoutThenAvailable(a: LimitOrderMatch, b: LimitOrderMatch): number {
```
Wizdave97: **Escrow release is strictly proportional, so every fill settles at the swapper's rate — which decides both the filter and the sort.**

`IntentsBase._cumulativeReleased`, used to size every `RedeemEscrow(Partial)` on the destination and every cancel refund on the source:

```solidity
Released(filled) = filled >= totalRequired ? escrowTotal : escrowTotal * filled / totalRequired
```

`IntrinsicIntents` does the same locally (`(order.inputs[i].amount * fillAmount) / totalRequired`, with the completing fill sweeping the residue).

So a fill of `f` out of `T` releases `I × f / T`, and its effective rate is `f / (I × f / T)` = **`T / I`** — constant, independent of `f`, and identical for every partial. A limit order drawing on this swap is paid the swapper's rate, never its own.

**That gives the filter exactly.** A BID limit order at price `p` (quote per base) is authorised to pay at most `p`. It can participate only if `T / I <= p`, which rearranges to `I × p >= T` — that is `offerFor(inputNet, price) >= requestedOutput`. The issue's §3 rule 5, which I wrongly waved through.

Without it the current code pays above the operator's signed rate, and not only in the combined case: a *single* order with `offer < reqOut` partial-fills at `payout`, and that partial still settles at `T / I > p`. So this is live on the single-order path today, not just the new one.

**The shape to build:**

1. **Filter** to `offer_i >= requestedOutput`. Everything surviving can be drawn on without breaching its signed rate; everything else is excluded, not merely ranked lower.
2. **Sort best price to worst** — `Side::cmp_best_first`, bids descending and asks ascending, as `quote.rs` orders them.
3. **Price the full input against each**: `payout_i = min(offerFor(inputNet, price_i), available_i)`, which is what `matchLimitOrders` already computes. Keeping the full input is right here precisely because each bid is its own fill.
4. **One bid per limit order**, in that sequence, each a partial fill the gateway accumulates in `_partialFills` and continues via `delete _filled[commitment]`.

Submitting separately is what removes the double-count I raised earlier: each bid is clamped independently to `remaining`, so the total paid is `min(Σ payout_i, T)` by construction rather than by arithmetic here. The last bid is the one that completes the order and sweeps the escrow residue.

**Worth settling before it is built:**

- **Sort direction.** Since every fill settles at `T / I` regardless, consumption order does not change what this swap costs — it decides which inventory is spent. Best-price-first spends the most generous orders first, which are also the ones most likely to match future swaps; worst-first (closest to `T / I`) preserves them. I read "best price to worst" as the orderbook's taker-facing convention, but if the intent is to preserve the generous end, say so and it should be the reverse.
- **Gas per bid.** N bids are N fills, so the profit gate at `fx.ts:575` has to run per bid against that bid's own payout, not once against the combined figure. A tail bid can easily be worth less than its gas.
- **Output calldata.** `PartialFillNotAllowed` means these orders take exactly one bid; `partialEligibleCheap` already tests it, and the multi-bid path has to respect it.
- **Bookkeeping.** `reservations` on the bid row becomes one hold on one limit order per bid, which is simpler than today's list — `holdAll` and the draw-down fan-out in `settleFilledLimitOrder` both collapse back to the one-to-one case the original docstring wanted.
- **Races.** Another solver can fill part of the order between two of our bids. Each bid must tolerate being clamped to a smaller `remaining` than it was sized for; the draw-down already reads the delivered amount from the emitted event, so that side is fine.

Wizdave97 (commented):
**Design direction for multi-order fills, plus a correction to something I got wrong in my first review.**

The target shape: when several limit orders serve one incoming order, sort them **best price to worst**, keeping only those that match the incoming order's rate; price the **full order input** against each one; and **submit a separate bid per limit order** rather than one bid for a combined amount.

Reading the gateway on current `main` makes the case for that stronger than I put it, and shows that a rate filter is not optional — see the comment below.

**The correction.** In my first review I said dropping the issue's `offer >= reqOut` match condition (§3 rule 5) was a safe, better-reasoned relaxation, on the grounds that ranking by `payout` never loses a fill the rule would have kept. That reasoning only holds for *selecting* a single order. It is wrong about whether an order should participate at all, because of how escrow is released. `offer >= reqOut` is a genuine participation condition and should come back.

sdk/packages/simplex/src/orderbook/matching.ts
Wizdave97: **Nonces, for step 4.** N bids on one incoming order share one commitment, so they share one nonce key — the sequence is what has to distinguish them, and it must increase.

An ERC-4337 nonce is a 192-bit key and a 64-bit sequence. `CryptoUtils.bidNonceKey(commitment, sessionKey)` is `keccak256(abi.encodePacked(commitment, sessionKey)) & (2^192 - 1)`, derived from the commitment alone, so every bid on the same incoming order lands on the same key. The i-th bid must carry sequence `base + i`.

**What breaks today.** One nonce is read per order and reused. `estimateGasFillPost` does

```ts
const nonce = await client.readContract({ ..., functionName: "getNonce",
  args: [this.solverAccountAddress, CryptoUtils.bidNonceKey(orderCommitment(order), order.session)] })
```

caches it on the gas estimate, and `prepareBidUserOp` signs with `cachedEstimate.nonce`. `getNonce` returns `key << 64 | currentSequence`, and the sequence only advances when an op actually executes. Build N bids before any of them has, and all N carry the same nonce: the first is usable and the rest fail EntryPoint validation with AA25. Read the base once and derive `base + i`, rather than N reads that all return the same thing.

The binding check is indifferent to this. The orderbook compares only the key — `op.nonce >> NONCE_SEQUENCE_BITS != bid_nonce_key(commitment, session)` → `BAD_NONCE_BINDING` in `verify.rs` — and the SDK does the same at `phantom-aggregation.ts:960`. The low 64 bits are free.

**The consequence to design around: strict sequencing means head-of-line blocking.** The EntryPoint's NonceManager requires `nonceSequenceNumber[sender][key]++ == seq`, so a key's sequences must be consumed in order with no gaps. Bids are executed by someone else, and any bid that never executes — not selected, reverted in validation, expired past its `validUntil` — permanently strands every later bid on that key, because the on-chain sequence never reaches theirs.

That collides with the execution model in #1259: *"A bid that fails is skipped and the walk continues; executed bids cannot be replayed, since the userOp nonce is consumed."* The walk assumes a skipped bid is free. On a shared key it is not. So either the signing order has to be the execution order with any dropped bid's successors re-signed at the freed sequence, or bids need distinct keys — which the commitment-derived key currently rules out. Worth settling on #1259, since its walk depends on the answer.

**A second assumption in the same place.** The bid store is built around one bid per commitment, and N bids share one:

- `claimReservation(commitment)` reads `... WHERE commitment = ? AND reservations IS NOT NULL ORDER BY id DESC LIMIT 1` — the newest row only. The other N-1 bids' holds are never claimed, so their reservations leak and that capacity is lost until someone notices.
- Its guarded write is `UPDATE bids SET reservations = NULL WHERE commitment = ? AND reservations = ?`. If two of the N rows carry identical `reservations` JSON — the same limit order and amount, entirely possible — the update clears both, `result.changes` is 2, and the function returns `[]`. The holds are wiped without ever being released.
- `releaseReservation(commitment)` and `settleFilledLimitOrder(commitment, ...)` are keyed on commitment alone and cannot tell the N bids apart, and `pendingRetractions` is a `Set` of commitments, so a retraction cannot name which bid it retracts.

None of this is hard, but it is more than the matcher: the bid row needs its own identity — the nonce sequence is the natural one, since it is already unique per bid on that key — and the reservation and retraction paths need to address a bid rather than a commitment.

dharjeezy:
Second round is in `104c939e`.

The overfill donation is fixed by clamping the bid to `min(offered, output.amount)`. Your reading of the gateway is right, and the part that stung is that `payoutSurplusUsd` was booking the donation as the profit justifying the fill. I checked #964 before touching it: it disabled the clamp for curve-priced legs, where the figure was a curve's answer for that leg, which does not carry over to a rate applied to the whole input.

`offer >= requestedOutput` is back as a participation condition, on your derivation. Worth noting it was live on the single-order path too, as you said, not only on the combined one.

A fill now settles its holds in one store transaction. I had argued ordering was enough because a transaction would span two stores; that was wrong, `SqliteDataStore` hands the same `bids.db` connection to both, so §6's requirement was always reachable. `settleFill` splits as a result: the draw-down goes inside the transaction, `resize` stays outside because it is a round trip.

Schema pin refreshed, and `.github/workflows/check-orderbook-schema.yml` diffs it against the server weekly and on any change here. Token decimals come from `serverInfo.chains` now rather than an RPC read per token per post. That change broke the live orderbook test against my locally built binary, which predated the field, which is a decent argument for the workflow existing.

Two decisions in there are mine rather than yours, and both are worth arguing with:

**Tightest first, not best price first.** Since every fill settles at `T / I` whatever funds the slice, consumption order does not change what this swap earns, only what stays resting. A more generous order qualifies for every swap a tighter one does and for swaps it cannot serve, at the same cost per unit, so the tight end is what to spend. This is the case you flagged as "if the intent is to preserve the generous end, say so and it should be the reverse".

**One bid with slices, not one bid per order.** With the bid clamped to the ask, the total is the ask, so a single bid covers it: each order funds a slice, is drawn down by that slice and receives that fraction of the input, so all of them settle at `T / I` inside their own terms. That removes the shared nonce key, the head-of-line blocking and the #1259 dependency rather than solving them. If you think the per-bid shape buys something this misses, I would rather hear it now than after #1259 lands.

sdk/packages/simplex/src/orderbook/limit-orders.ts
```diff
@@ -408,6 +490,8 @@ export class LimitOrderService {
 	 */
 	async repost(order: LimitOrder | null): Promise<LimitOrder | null> {
 		if (!order) return null
+		// A same-asset order was never on the book, so there is nothing to put back.
+		if (isLocal(order)) return this.store.setStatus(order.id, "open")
```
Wizdave97, line 552: **No repost path checks `hasExpired(order.expiresAt)`.** `repost()` tests `isLocal`, then withdraws and posts. Its three callers:

- **`reconcile()`** runs on its own clock (`reconcileIntervalSecs`, 300s by default) and never calls `expireStale`. Its guards are `isLocal`, whether the commitment was `found`, and `POST_GRACE_MS` — expiry is not among them. Any reconcile tick landing between an order's `expiresAt` and the next renew tick reposts it.
- **`settleFill()`** reposts whatever the draw-down left, with no expiry test. A fill that settles after the order expired puts it straight back up.
- **`renewExpiring()`** filters on `status: "open"` and `expiresWithin(bookExpiresAt, marginSecs)` — the *posting's* expiry, never the order's.

Only the renewal path is safe today, and only incidentally: `LimitOrderLifecycle.renew()` happens to call `expireStale()` before `renewExpiring()`. That ordering is the whole protection. It does not extend to the reconcile clock, it does not extend to `settleFill`, and any future caller of `repost` loses it silently.

**Why it matters, in this file's own words.** `expireStale`'s docstring already names both hazards:

> The matcher refuses an expired order, so leaving it alone would advertise depth that no swapper could ever draw on: renewal would keep its posting alive and reconciliation would put it back if it lapsed.

That is exactly right, and `serves()` confirms the first half — `if (order.expiresAt !== null && new Date(order.expiresAt) <= now) return false`. So a reposted expired order is depth simplex publishes and then refuses to honour. It is quoted to swappers through `bestRate`, `rates` and `quote`, it counts toward `availableLiquidity`, and every fill that comes back against it is skipped. That is worse than a missing order: it misrepresents the book.

**The fix is one guard in `repost()`**, not three. Expiry is a property of the order, so the check belongs where the posting is made rather than in each caller — and unlike `POST_GRACE_MS` there is no timing subtlety, since an expired order is expired on every clock. An expired order reaching `repost` should be withdrawn and moved to `expired`, which is what `expireStale` already does, rather than posted and left for the next sweep.

Worth a test per path: `reconcile()` and `settleFill()` each given a row whose `expiresAt` is in the past, asserting nothing is submitted. `lifecycle.test.ts` covers the ordering inside `renew()`, which is the one path that already works.

Wizdave97 (commented):
**Simplex should not repost a limit order whose operator expiry has passed** — and today all three repost paths can.

`expireStale` is the only code that reads `expiresAt`, and it is a separate sweep rather than a guard. `repost()` is the shared entry point for every posting and never checks it, so whether an expired order gets put back on the book depends on which clock fires first.

sdk/packages/simplex/src/tests/orderbook/orderbook.live.test.ts
```diff
@@ -0,0 +1,215 @@
...
+[min_order_size]
+cNGN = 100
+USDC = 100
+
+[chains."EVM-8453"]
+rpc_url = "${indexer}/rpc/EVM-8453"
+gateway = "${GATEWAY}"
+solver_accounts = ["0x7cb55539d1144F62422099c3FA3405092022c88C"]
+
+[chains."EVM-8453".tokens]
```
Wizdave97: orderbook only needs you to specify rpc endpoints no need for tokens

sdk/packages/simplex/src/strategies/fx.ts
Wizdave97: **Direction on this: do not combine payouts from different limit orders into one bid.** `combinedPayout` should go, not be re-derived as a clearing price.

Send **one bid per matching limit order**, each carrying that order's own offer. Three matching limit orders means three bids — not one bid for their sum, and not one bid for a blended rate.

Each bid's output is that order's own figure, unchanged from what the matcher already computes:

```
payout_i = min(offerFor(fullInput, price_i), available_i)
```

The full input is priced against each order independently, which is correct here precisely *because* each bid is its own fill. Nothing is summed in simplex.

**The gateway does the combining, and it is already built for it.** Each fill clamps itself to what is outstanding — `fillAmount = solverAmount > remaining ? remaining : solverAmount` — accumulates progress in `_partialFills[commitment][outputToken]`, and `delete _filled[commitment]` on an under-fill lets the next bid continue. So the total delivered is `min(Σ payout_i, T)` by construction, enforced on chain, rather than by arithmetic here that can get it wrong. That is what removes the over-payment: there is no sum to double-count.

This is also the execution model #1259 specifies — *"It sorts the bids on an order by rate, best first, and executes them in that order until the order is full, one `select` and one fill per bid. The last bid executed is usually larger than what is outstanding; the contract caps the fill and scales the take with it, so no solver re-signs a smaller slice."* Building it now is forward-compatible; a combined bid is the "splits an order into child orders" model that issue retires.

**What a tail bid does.** Once the order is complete `_filled[commitment]` stays set, and `IntentGatewayV2.sol:524` is `if (_filled[commitment] != address(0)) revert Filled();`. So a bid that arrives after the order is full **reverts** — it does not silently fill zero. Clean, but it means simplex must treat a reverted tail bid as a normal outcome and give its hold back, not as an error.

**What has to change with it:**

- `matchLimitOrders` keeps returning the list; `fx.ts` stops reducing it. `targetOutput` becomes per-bid, and the `finalOutputAmount` / `fillerOutputs` / `fillOptions.outputs` chain runs once per bid.
- `holdAll`'s all-or-nothing becomes per-bid: each bid reserves only against its own limit order, and a hold that cannot be taken drops that bid rather than the whole set. A bid that reverts with `Filled()` releases its own hold.
- The profitability gate at `fx.ts:575` runs per bid against that bid's own payout. A tail bid can easily be worth less than its gas, and a combined figure hides that.
- The filter and sort from my earlier comment on `matching.ts` apply here: only orders whose offer meets the incoming order's rate may bid at all, best price first.
- Nonces and bid identity are the two structural pieces — both covered in the thread on `matching.ts:121`, and both are prerequisites rather than follow-ups.

Wizdave97 (approved):

Issue #1239, opened 2026-09-09
[simplex] Desktop code signing and notarization


seunlanlege:
Part of #1235

## Why

Unsigned desktop apps are actively blocked. macOS Gatekeeper refuses to open them without a right-click override, and Windows SmartScreen shows a red warning. For an app that holds solver private keys and moves money, shipping something the OS calls untrusted is not an option.

**Start procurement now.** Apple enrolment and Windows certificate issuance take days to weeks and are on the critical path for the first public release.

## macOS

- [ ] Apple Developer Program enrolment (~$99/yr). Organization enrolment needs a D-U-N-S number and takes longer than individual.
- [ ] Developer ID Application certificate.
- [ ] Hardened runtime enabled, with an entitlements plist. Electron typically needs `allow-jit` and `allow-unsigned-executable-memory`; grant nothing beyond what fails without it.
- [ ] **Sign the bundled Node binary** (#1237) with the same Developer ID, under the hardened runtime with the same JIT entitlements, listed in `electron-builder`'s `binaries` config. A nested executable that is unsigned or signed by someone else fails to launch under the app's hardened runtime — the symptom is the solver simply never starting.
- [ ] Notarization via `notarytool`, then staple. Notarization covers nested executables, so the Node binary must be staged and signed before this runs.
- [ ] Verify with `spctl -a -vvv` and by opening on a machine that has never seen the app.

## Windows

- [ ] Choose between Azure Trusted Signing (cheap, monthly, cloud-held key) and a traditional EV certificate (expensive, annual, hardware token). Trusted Signing is the better fit for CI — an EV token in a GitHub runner is painful.
- [ ] Sign the NSIS installer and the bundled executables, the bundled Node binary included.
- [ ] Note that SmartScreen reputation accrues over downloads. Even correctly signed, early users may see a warning. Budget for it rather than being surprised.

## Linux

No signing gate. Optionally GPG-sign the deb and publish the public key; consider Flathub later, though sandboxing versus a local API server needs its own thought.

## CI

- [ ] Store credentials as GitHub Actions secrets: Apple cert p12 + password, App Store Connect API key, Windows signing credentials.
- [ ] Confirm they are unavailable to workflows triggered from forks.
- [ ] Document rotation. A certificate expiring silently breaks releases and nobody notices until a release fails.

## Acceptance criteria

- [ ] macOS dmg opens on a clean machine with no Gatekeeper prompt, and `spctl` passes.
[...]

Issue #1238, opened 2026-09-09
[simplex] Multiplatform desktop builds and release CI


seunlanlege:
Part of #1235

## Why

Three OSes, two architectures each. macOS and Windows builds must run on their own runners — signing tools are platform-native and cross-building is not an option once #1239 lands.

## Targets

| Platform | Arch | Format | Runner |
|---|---|---|---|
| macOS | arm64 + x64 | dmg + zip | `macos-latest` |
| Windows | x64 | NSIS | `windows-latest` |
| Linux | x64 | AppImage + deb | `ubuntu-latest` |
| Linux | arm64 | AppImage + deb | `ubuntu-24.04-arm` |

macOS should ship a universal binary or both slices. The zip alongside the dmg is what the updater consumes. Windows arm64 is optional for v1 — decide based on whether any operator asks.

## Tasks

- [ ] Add `electron-builder` config. Bundle id, product name, category, icons per platform.
- [ ] Ship the solver outside `app.asar`, as plain files. The bundled Node binary cannot read asar archives, so none of the solver can live inside one. Put `simplex/dist` under `resources/simplex`, with a `node_modules` next to it holding the six packages the bundle leaves external: `ssh2`, `pino`, `pino-pretty`, `thread-stream`, `@solana/web3.js` and `@solana/spl-token`. Put the Node binary under `resources/runtime`. These are the paths `desktop-paths.ts` already expects. Then confirm the packaged solver starts from there with no warnings. (Moved from #1237. `ssh2`'s optional native add-ons already fail silently and fall back to JavaScript, and logs are formatted on the main thread, not through a pino worker transport.)
- [ ] New workflow `publish-simplex-desktop.yml`, tag-driven, following the shape of the existing `.github/workflows/publish-simplex.yml`. Proposed tag: `simplex-desktop-v*`, kept separate from `simplex-v*` so a desktop release does not force an npm release.
- [ ] Build matrix per the table. `submodules: recursive` and the protoc install step both carry over from the existing workflow.
- [ ] Attach artifacts to a GitHub release, plus the updater metadata files from #1240.
- [ ] Reproducibility: pin the Electron version and the builder version explicitly.
- [ ] Smoke test in CI — launch the packaged app headless, hit `/health`, exit non-zero on failure. WebdriverIO or Playwright both support Electron.

## Sizing

[...]

PR #1277, opened 2026-09-16
[simplex-desktop]: add safe desktop auto-update lifecycle

### LINKED ISSUE #1240: [simplex] Desktop auto-update
Part of #1235

## Why

Operators run this unattended for long stretches. A solver on a stale build can be mispricing, or missing a fix that costs money. Manual updates will not happen reliably.

## The constraint

The solver runs detached (#1237) and holds its own binary open. **On Windows you cannot overwrite a running executable**, so the installer would fail outright. The solver has to stop for its own update — that is not avoidable, and it is a tie with the in-process model.

The trick is ordering: **stop the solver as part of the update, not despite it.**

## The sequence

1. App checks for updates and **downloads in the background.** The solver keeps filling throughout.
2. Update staged. App waits for a safe moment: solver idle, nothing in flight.
3. App calls `/api/stop` — the existing endpoint, routing to `stopAll()`, which closes the store cleanly. Same teardown as Ctrl-C.
4. Solver exits, releases its file handle, removes its socket.
5. App calls `autoUpdater.quitAndInstall()`. Nothing is holding the files open now.
6. electron-updater relaunches the app, which spawns the new solver. `restorePhantomBids` (`core/boot.ts:657`) retracts anything left live on boot.

Downtime is step 3 to step 6 — roughly 30-90 seconds, mostly the installer and the solver's chain-client startup.

## Do not attempt a zero-downtime handoff

Running a new solver alongside the draining old one means **two processes signing with the same key, racing on nonces**. That risks double-bidding. A minute offline is the better failure.

## Where the binary lives

Keep the solver **inside the app bundle**. One artifact, one signature, and the update is just the app update.

Rejected: extracting to `userData/runtime/<version>/` to avoid overwriting a running file. It buys rollback and sidesteps Windows locking, but complicates macOS notarization for a helper outside the bundle. Not worth it for a 60-second window.

## Three rules

- **Never force the stop.** If the solver will not go idle, defer and retry at the next window. Killing it mid-fill leaves a signed bid outstanding with no process to complete it.
- **Do not try to survive in-flight work across the restart.** The idle gate exists so there is nothing to survive.
- **Add a staleness nag.** An operator who is never idle would otherwise drift months behind without knowing.

## Tasks

- [ ] `electron-updater`, backed by GitHub Releases.
- [ ] Verify update signatures. On macOS this rides on the code signature; on Windows confirm the updater validates the Authenticode signature rather than just the hash.
- [ ] Implement the ordered sequence above. The stop-before-install ordering is the whole design — do not let `quitAndInstall()` run with the solver alive.
- [ ] Idle gate before step 3. Reuse or extend the existing pause/halt state rather than inventing a second notion of "busy".
- [ ] Defer-and-retry when never idle, plus a staleness nag after some threshold.
- [ ] **Post-update health check.** After relaunch, confirm the new solver answers on its socket within a timeout. Surface a failure loudly rather than sitting there not filling.
- [ ] **Version skew banner.** `/api/status` already returns the solver's version (`dto.ts:50`, from `bin/simplex.ts:211`). If it disagrees with the app's — a failed stop, or a manually started solver — offer a restart rather than letting a new UI drive an old API.
- [ ] Channels: stable by default, optional beta. Reuse the tag scheme from #1238.
- [ ] Rollback: keep previous releases downloadable and document the downgrade.

## The headless gap

A solver autostarted alone, with the app never opened, would never be updated. Resolved in #1243 by **autostarting the app rather than the solver** — the app is then always present to run this sequence. If that changes, this issue needs a solver-side update check, which is the point where an OS service starts looking better than a supervised child.

## Acceptance criteria

- [ ] An installed older build detects, downloads and applies a newer release on all three platforms, and the solver comes back filling.
- [ ] The update never runs while a fill is in flight.
- [ ] A tampered artifact is rejected.
- [ ] A new solver that fails to boot is reported, not silently ignored.

ddboy19912:
## Summary

closes #1240 by adding automatic updates to the Simplex desktop application through GitHub Releases and `electron-updater`.

Updates download in the background while the solver continues operating. Installation waits for a safe idle window, gracefully stops the solver, verifies that both its socket and process have exited, and only then launches the installer.

## What changed

- Added `electron-updater` with GitHub Releases configuration.
- Added stable and optional beta update channels.
- Disabled automatic installation during ordinary Electron shutdown.
- Added persistent, atomic update receipts under Electron user data.
- Added six-hour update checks and resumable staged updates.
- Added solver work reporting for:
  - Queued and active evaluations.
  - Queued and active fills.
  - Pending and active bid retractions.
- Added the safe update sequence:
  - Download without interrupting the solver.
  - Wait until the solver is idle.
  - Call the existing graceful `/api/stop` endpoint.
  - Wait for socket removal and operating-system PID exit.
  - Call `quitAndInstall()` and relaunch Simplex.
- Added deferred retries when the solver remains busy or cannot exit.
- Added a 24-hour staleness notification for updates waiting on an idle window.
- Added post-update application and solver version verification.
- Added a blocking version-skew screen when the desktop and solver versions differ.
- Added a native **Restart with bundled solver** action.
- Added SHA-512 update metadata verification configuration.
- Kept macOS application-signature and Windows Authenticode publisher verification enabled.
- Documented update behavior, channels, rollback, and release requirements.

## Safety properties

- The updater never force-kills the solver.
- `quitAndInstall()` is never called while the detached solver is still running.
- Active fills, evaluations, and retractions must drain before shutdown.
- A paused solver may discard only evaluations that have not started, matching its existing graceful-stop behavior.
- A failed installer invocation remains staged and retryable.
- Ordinary Electron quit does not install a downloaded update.
- Channel changes never permit automatic downgrades.
- A new UI cannot operate a solver with a different version.

## Verification completed

- [x] Desktop unit and integration tests: 82 passed, 1 platform-only skip.
- [x] Affected Simplex CLI, filler, socket-server, and setup tests: 117 passed.
- [x] Real Electron E2E tests: 10 passed.
- [x] Version-skew banner and native restart action verified through `simplex://local`.
- [x] Desktop typecheck and lint passed.
- [x] Desktop and Simplex production builds passed.
- [x] UI typecheck and production build passed.
- [x] Update ordering, idle gating, retry behavior, staleness notification, channel persistence, and post-update verification covered by tests.
- [x] No force-stop path introduced.
- [x] Documentation and rollback instructions added.

## Remaining release validation

The following acceptance checks require the installer and signing work from #1238 and #1239. They cannot be completed using the current source-only development build:

- [ ] Install an older signed Simplex desktop release and verify that it detects, downloads, installs, and relaunches a newer release on macOS, Windows, and Linux.
- [ ] Confirm that the relaunched packaged application starts the updated bundled solver and resumes filling.
- [ ] Tamper with a generated release artifact and confirm rejection by SHA-512 and platform-signature verification.
- [ ] Confirm macOS code-signature validation using a notarized update.
- [ ] Confirm Windows Authenticode publisher validation using the production signing certificate.

The application-side update lifecycle and signature-verification configuration are implemented by this PR. Final release validation should be performed once #1238 produces installers and #1239 signs them.


ddboy19912:
@seunlanlege This is ready

sdk/packages/simplex-desktop/electron-builder.yml
```diff
@@ -0,0 +1,11 @@
+publish:
+  provider: github
+  owner: polytope-labs
+  repo: hyperbridge
+  tagNamePrefix: simplex-desktop-v
```
seunlanlege, line 7: `tagNamePrefix` is only used by electron-builder when it publishes. electron-updater's GitHub provider never reads it, so installed builds won't find these releases.

- **Stable** (`allowPrerelease = false`) asks `https://github.com/polytope-labs/hyperbridge/releases/latest`. Today that returns `tesseract-v2.4.12`. The updater then requests `latest.yml` from the Tesseract release and gets a 404.
- **Beta** scans `releases.atom` and skips any tag that isn't valid semver. `simplex-desktop-v1.0.0-beta.1` isn't valid semver, so nothing matches.
- The Atom feed also only lists the 10 newest releases across indexer, sdk, simplex and tesseract.

See `GitHubProvider.getLatestVersion()` and `getLatestTagName()` in electron-updater 6.8.9.

Two options would work. One is a `generic` provider pointed at a fixed URL that always holds the current `latest.yml` and `beta.yml`. The other is a custom provider that filters releases by the `simplex-desktop-v` prefix.

sdk/packages/simplex-desktop/src/update-coordinator.ts
```diff
@@ -0,0 +1,426 @@
...
+		if (this.preferences.receipt?.installAttemptedAt) {
+			await this.verifyPreviousInstall()
+			if (this.current.state === "error") return
+		}
+		if (this.preferences.receipt) {
+			this.setStatus({
+				state: "waiting-for-idle",
+				targetVersion: this.preferences.receipt.targetVersion,
+			})
+			await this.tryInstall()
```
seunlanlege: After a relaunch, this installs from the saved receipt alone. But a fresh process has no downloaded file in electron-updater, because `downloadedUpdateHelper` is only created by `executeDownload()`.

- **Windows and Linux:** `BaseUpdater.install()` sees `installerPath == null`. It emits `error` ("No update filepath provided, can't quit and install") and returns without quitting.
- **macOS:** `MacUpdater.quitAndInstall()` calls `nativeUpdater.checkForUpdates()` with no feed URL set, because `setFeedURL()` only runs during a download.

By then `runInstallAttempt()` has already stopped the solver.

Suggest calling `checkForUpdates()` on launch and only installing after `update-downloaded` fires in this process. electron-updater re-validates the cached file and emits the event again without downloading it again.

The test "resumes a staged download after the desktop app restarts" asserts that `checkForUpdates` is not called, so it currently locks this behaviour in.

sdk/packages/simplex-desktop/src/update-coordinator.ts
```diff
@@ -0,0 +1,426 @@
...
+		this.installing = true
+		try {
+			const attempted: UpdateReceipt = {
+				...receipt,
+				installAttemptedAt: this.options.now?.() ?? Date.now(),
+			}
+			this.preferences = { ...this.preferences, receipt: attempted }
+			this.options.store.write(this.preferences)
+			this.setStatus({ state: "installing", targetVersion: receipt.targetVersion })
+			this.options.updater.quitAndInstall(false, true)
```
seunlanlege, line 398: `quitAndInstall()` never throws in electron-updater, so this catch block doesn't run for real failures. `BaseUpdater.install()` catches internally, calls `dispatchError()` and returns `false`. `MacUpdater` reports failures through the native updater's `error` event.

When an install fails:
- The `error` listener sets the state to `deferred`.
- `installing` stays `true`, so `scheduleIdleRetry()` and `runInstallAttempt()` return early from then on.
- `quitting` in `main.ts` stays `true`.
- Nothing restarts the solver, which was stopped just before this call.

On macOS this includes a rejected update. Squirrel checks the code signature after `quitAndInstall()`, when the solver is already down. Windows checks the signature and SHA-512 during download, before the stop.

Suggest treating an `error` event while `installing` as a failed install. Reset `installing`, clear `installAttemptedAt`, and restart the solver.

The test "keeps a staged update retryable when the installer call fails" mocks a throw that the library never does. `FakeUpdater` should emit `error` instead.

sdk/packages/simplex-desktop/src/update-coordinator.ts
```diff
@@ -0,0 +1,426 @@
...
+		for (const [event, listener] of Object.entries(this.listeners))
+			this.options.updater.removeListener(event, listener)
+		this.started = false
+	}
+
+	async checkNow(): Promise<void> {
+		if (!this.options.packaged || this.installing) return
+		let checkChannel: UpdateChannel | undefined
+		try {
+			if (this.preferences.receipt?.installAttemptedAt) {
```
seunlanlege, line 199: Once `installAttemptedAt` is set, this branch never checks for updates again.

Say the installer fails after the app has quit, for example UAC is denied or NSIS errors. The next launch reports an error here. Every later `checkNow()`, from the menu or the 6-hour timer, repeats that error and sends another notification. `setChannel()` keeps the attempted receipt too. The only way out is deleting `desktop-updates.json`.

Suggest clearing `installAttemptedAt` after reporting the failure once, then falling back to a normal check.

sdk/packages/simplex-desktop/src/solver-supervisor.ts
```diff
@@ -53,29 +54,89 @@ function socketRequest(
...
+	)
+}
+
+export function solverVersion(status: SolverStatus): string | undefined {
+	return status.state === "setup" || status.state === "running" || status.state === "paused"
+		? status.version
+		: undefined
+}
+
+export function solverHasVersionSkew(status: SolverStatus, appVersion: string): boolean {
```
seunlanlege: This compares the desktop's `app.getVersion()` (from `simplex-desktop/package.json`) with the solver's `packageJson.version` (from `simplex/package.json`). `verifyPreviousInstall()` and the UI's `resolveBootstrapState()` make the same comparison.

Both files say 0.16.2 today, but nothing keeps them equal, and simplex is released on its own schedule. The first simplex version bump without a matching desktop bump causes three problems:
- The desktop reports a mismatch with its own bundled solver. It now also blocks the setup wizard for new users.
- "Restart with bundled solver" restarts into the same mismatch.
- Every update is reported as failed.

Suggest comparing against the bundled solver's own version, read from the bundled simplex `package.json` at startup. Alternatively, fail the desktop build when the two versions differ.

sdk/packages/simplex-desktop/src/update-coordinator.ts
```diff
@@ -0,0 +1,426 @@
...
+		if (!receipt) return
+		const solver = await this.options.probeSolver().catch((error) => {
+			this.defer(error)
+			return undefined
+		})
+		if (!solver) return this.scheduleIdleRetry()
+		// A channel change can discard the receipt while the idle probe is pending.
+		if (this.preferences.receipt !== receipt) return
+
+		if (solver.state === "stopped") {
```
seunlanlege, line 354: When the solver is stopped, this installs right away. The relaunched app then starts the solver in `prepareDesktop()` through `startOrAttachSolver(true)`.

If an operator stopped filling on purpose, for example during an incident, an unattended update turns filling back on. Suggest either not auto-installing while the solver is stopped, or recording that it was stopped so the relaunch doesn't start it.

sdk/packages/simplex-desktop/src/solver-supervisor.ts
```diff
@@ -105,10 +166,7 @@ export function stopRequestAccepted(status: SolverStatus): boolean {
 }
 
 function sameStatus(left: SolverStatus, right: SolverStatus): boolean {
-	return (
-		left.state === right.state &&
-		("detail" in left ? left.detail : undefined) === ("detail" in right ? right.detail : undefined)
-	)
+	return JSON.stringify(left) === JSON.stringify(right)
```
seunlanlege: The status now includes the work counts, so this comparison fails whenever any count changes. `onChange` then rebuilds the tray menu and the application menu on the 3-second poll, instead of only when the state changes.

Comparing only the fields the menus show (state, detail and version) would avoid those extra rebuilds.

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -391,6 +400,31 @@ export class IntentFiller {
...
+		for (const queue of this.chainQueues.values()) {
+			queuedFills += queue.size
+			activeFills += queue.pending
+		}
+		return {
+			queuedEvaluations: this.globalQueue.size,
+			evaluating: this.globalQueue.pending,
+			queuedFills,
+			activeFills,
+			retractions: this.pendingRetractions.size + this.retractionQueue.size + this.retractionQueue.pending,
```
seunlanlege: `pendingRetractions` holds fills that arrived before their bid was stored. It isn't work in progress, and `stop()` doesn't wait for it.

If an entry never gets a stored bid, this count stays above zero. The desktop then never sees an idle window, so the update never installs. Suggest leaving `pendingRetractions` out of this count.

sdk/packages/simplex/src/core/filler.ts
```diff
@@ -19,12 +19,18 @@ import { parseChainKey } from "@/config/interpolated-curve"
 import { INTENT_GATEWAY_V2_ABI } from "@/config/abis/IntentGatewayV2"
 import type { Address } from "viem"
 import pQueue from "p-queue"
-import { type ChainClientManager, type ContractInteractionService, DelegationService, type RebalancingService } from "@/services"
+import {
```
seunlanlege: This import change is Prettier reformatting, and so are most of the other hunks in this file and in `UiServer.ts`. None of it relates to auto-update. Could you drop it? It hides the real changes and will conflict with other work on these files.

The removed `biome-ignore` comment on `hyperbridgeApi` is unrelated too.

seunlanlege (approved):

PR #1278, opened 2026-09-16
[evm]: refund protocol fees on cancelled escrow

### LINKED ISSUE #1260: [evm]: refund the protocol fee on cancellation
## Motivation

The protocol fee leaves the user at placement. `placeOrder` deducts it from each input, commits the order over the post-fee amounts, escrows those, and emits `DustCollected` for the fee right away. From then on the fee is untracked, so a cancel refunds only the post-fee escrow and the user pays the protocol for an order that was never filled. The relayer fee does not behave this way: `_withdraw` forwards `TRANSACTION_FEES` to whoever the refund goes to, so it comes back on cancel. The protocol fee should too.

## Design

Hold the fee, settle it at the end.

1. **Placement.** Deduct the fee exactly as now, so the commitment, the escrow, bids and every off-chain decoder stay untouched. Instead of declaring it dust, record it per order and input token under the commitment, the way `TRANSACTION_FEES` is recorded. One appended mapping, proxy-safe.
2. **A fill completes the order.** The recorded fee becomes protocol dust. `DustCollected` moves here.
3. **A cancel refunds escrow.** Refund the fee in the same proportion as the escrow being refunded: `fee × refunded / committed`, where `committed` is the post-fee input amount. The remainder becomes dust. One rule covers both cases: an unfilled order gets the whole fee back, a partially filled one gets the unfilled fraction back and the protocol keeps the fee on what was filled.

### Where it lives

All of it fits in `_withdraw`, which already sees every terminal settlement for same-chain and cross-chain orders: the finalize branch settles the fee to dust, the refund branch pays the proportional part back before releasing the escrow. Cross-chain cancels reach it through `onGetResponse` and `RefundEscrow`, so #980's proven-unredeemed refunds get the fee treatment for free, since the refunded amount already is the unfilled fraction.

### Why record it rather than recompute it

The fee can be approximated from the committed amount and the fee rate, but rounding is off by a wei and the rate is per destination and governance-set, so it may have changed between placement and cancel. Storing it costs one slot per input token per order, cleared at settlement.

## Considerations

- **Dust is untracked and governance sweeps by amount.** Fees held pending settlement must not be sweepable, or a sweep could empty a refund. Either track settled dust per token, or make `_sweepDust` subtract the pending fee balance. Decide which.
- **Indexer.** Fee volume is read from `DustCollected` at placement, and #1104 wants the placement fee token tracked. Both move to settlement. The refunded fee should appear in `EscrowRefunded`, or in its own event, so the indexer can reconcile.
- **SDK.** The gross-up for the protocol fee is unchanged, since filled orders still pay it.
- **Product question.** Same-chain cancels are immediate, so a full refund makes placing and cancelling free apart from gas. If solver effort on abandoned orders is a concern, refund fully only for orders that expired unfilled and keep a small fee on voluntary early cancels. Same rule with one extra branch. Needs a decision before implementation.

## Tasks

- [ ] Record the protocol fee per order and input token at placement; stop emitting `DustCollected` there
- [ ] `_withdraw`: settle the fee to dust on finalize, refund it pro rata on refund, clear the record
- [ ] Protect pending fees from `_sweepDust`
- [ ] Event for the refunded fee, or include it in `EscrowRefunded`
- [ ] Tests: unfilled cancel refunds the whole fee, partial fill then cancel refunds the unfilled fraction, full fill settles it to dust, cross-chain cancel through `onGetResponse` and `RefundEscrow`, native and ERC-20 inputs, fee-on-transfer inputs
- [ ] Indexer: move fee volume to settlement, index the refund
- [ ] Docs: intent gateway overview, fees section

royvardhan:
Refunds the protocol fee attributable to cancelled input escrow, including voluntary cancellation after a partial fill. A 1,000-token deposit with a 1% fee that releases half its net escrow returns 495 principal plus 5 fee tokens when the remainder is cancelled; the protocol earns the other 5 fee tokens.

Store the exact placement fee and original net input per order, and settle once using authenticated refundable principal. This preserves delayed cross-chain solver redemptions and existing order commitments. `DustCollected` reports earned fees at settlement; `ProtocolFeeRefunded` reports fee refunds separately from principal refunds. The indexer exposes refunds through `protocolFeeRefunds` without changing order status. SDK and Simplex ABIs also include the contract's `solver` field in `EscrowReleased`.

Governance sizes dust sweeps from earned dust less amounts swept or already in flight. The gateway does not enforce an escrow reservation. Orders placed before the upgrade retain their original nonrefundable fee treatment. Per-order storage is appended and existing proof slots remain unchanged. The deployment helper uses `migrate()` for version 2, empty initialization for version 3, and rejects unsupported versions.

Validation:

- 201 Foundry tests pass, covering refunds, rounding, native/taxed tokens, delayed redemptions, earned-dust sweeps, legacy upgrades, storage layout, and deployment sizes.
- 23 SDK tests, 1 Simplex ABI decoding test, and 12 targeted indexer tests pass.
- Indexer code generation/build and documentation MDX/TypeScript checks pass.
- Retained fee and settlement ABI entries match the freshly compiled contract; the removed aggregate getter is absent from all three client ABIs.

Closes #1260.


evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -184,6 +185,18 @@ abstract contract IntentsBase is EIP712 {
...
+    struct ProtocolFee {
+        uint256 amount;
+        uint256 committed;
+    }
+
+    /// @dev Appended accounting shared by the implementation and both delegatecall modules.
+    mapping(bytes32 => mapping(address => ProtocolFee)) public _protocolFees;
+
+    /// @dev Held protocol fees excluded from governance dust sweeps, keyed by input token.
+    mapping(address => uint256) public _pendingProtocolFees;
```
seunlanlege: The only reader is the check in `_sweepDust`. Sweeps are sized off-chain from `DustCollected`, which no longer includes pending fees, so this can go.

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -721,6 +760,9 @@ abstract contract IntentsBase is EIP712 {
             TokenInfo memory info = req.outputs[i];
             address token = address(uint160(uint256(info.token)));
             uint256 amount = info.amount;
+            uint256 balance = token == address(0) ? address(this).balance : IERC20(token).balanceOf(address(this));
+            uint256 reserved = _pendingProtocolFees[token];
+            if (balance < reserved || amount > balance - reserved) revert InvalidInput();
```
seunlanlege: This keeps a sweep off pending fees but not off escrowed principal, so it doesn't make a mis-sized sweep safe. Let's remove it with the mapping.

seunlanlege (changes requested):
Let's drop `_pendingProtocolFees`. The per-order `_protocolFees` record should stay: `RefundEscrow` doesn't carry the post-fee input, so the source chain has to remember it.

The aggregate only feeds the check in `_sweepDust`, and that check covers half the problem. It keeps a sweep off pending fees, but the same sweep can still take escrowed principal. The changelog note already says so.

Governance picks sweep amounts in the coprocessor's `sweep_dust` call. `DustCollected` now fires only for earned fees, so collected minus swept in the indexer (`ProtocolDustCollected`, `ProtocolDustSwept`) already leaves pending fees out. A sweep sized from those numbers never reaches them.

Keeping it isn't free. It adds a storage write to every placement and every settlement, and a `balanceOf` call for each swept token. A token whose balance ever falls below the reserved amount also becomes unsweepable.

To remove:

- The mapping, the `+=` in `placeOrder` and the `-=` in `_settleProtocolFee`.
- The reserved-amount check in `_sweepDust`.
- The getter in `IIntentGatewayV2` and its entries in the indexer, SDK and Simplex ABIs.
- The `_pendingProtocolFees` entry in the storage-layout test, and the `vm.store` of its slot in the upgrade test.
- The `_pendingProtocolFees` assertions. The fee tests can check `_protocolFees(commitment, token)` instead.
- The sweep tests that depend on the reservation: `testSweepCannotConsumeReservedERC20Fees`, `testSweepCannotConsumeReservedNativeFees`, `testSweepDuplicateTokensChecksCombinedTransfersAndRollsBack` and `testSweepBelowReservationRevertsWithoutUnderflow`.
- The reservation text in the changelog note and the PR description.

seunlanlege (approved):
Looks good, thanks for dropping the aggregate reservation.

PR #1279, opened 2026-09-17
[evm, sdk]: per-leg escrow for multi-leg intent orders

seunlanlege:
## Summary

Orders are a list of **legs** again: leg `i` sells `order.inputs[i]` for `order.output.assets[i]`, and legs may repeat tokens, e.g. one pair offered at several prices so different solvers can fill different price levels. To make that safe, the gateway keys escrow, fill progress and held protocol fees by **(commitment, leg index)** instead of by token, in the same storage slots.

| Storage (slot) | Before | After |
|---|---|---|
| `_orders` (9) | `[commitment][token]` | `[commitment][leg]`, fee pot still at `[commitment][TRANSACTION_FEES]` |
| `_partialFills` (11) | `[commitment][outputToken]` | `[commitment][leg]` |
| `_protocolFees` (14) | `[commitment][token]` | `[commitment][leg]` |

- `placeOrder`: requires `inputs` and `output.assets` non-empty, equal length, every output amount non-zero (replaces the single-leg rule from #980).
- Same-chain fill: a completing leg releases what remains of **its own** escrow. Same-chain cancel refunds each leg's remainder.
- Cross-chain: destination progress is per leg, `cancelFromSource` proves `_partialFills[commitment][i]` for each leg (distinct proof keys even when legs share an output token), `cancelFromDest` refunds per leg.
- Cancel proofs: `onGetResponse` indexes the proven values by storage key in transient storage (cleared before it settles), so each leg's value is one lookup instead of a rescan. Delivering a cancel proof hashes about 2N keys rather than N(N+1)/2: 642k gas instead of 1.14M at 32 legs, 2.5M instead of 12.2M at 128 legs.
- `_withdraw` treats `body.tokens[i]` as leg `i`. Every caller, and the peer gateway's `RedeemEscrow`/`RedeemEscrowPartial`/`RefundEscrow` bodies, already list one entry per leg in order, so the message format is unchanged.
- Predispatch: an order with predispatch calldata may not repeat an input token (`InvalidInput`); its escrow code is otherwise unchanged. The sweep is built and measured per leg, and `CallDispatcher` only checks that a call did not revert. With two legs of one token, the second sweep of a `false`-returning or fee-on-transfer token would pass, both legs would count the same arrived balance, and escrow could be credited beyond what arrived. Same-pair orders without predispatch are unaffected.
- Token values: a `bytes32` token must hold an address with its upper 12 bytes zero. `placeOrder` rejects such an input token and both fill paths such an output token (`InvalidInput`). The repeated-token checks compare the full `bytes32` while transfers read only the address, so `T` and `T | 1 << 255` would otherwise pass as two tokens; with a `false`-returning token, two aliased predispatch legs would each credit one sweep and a cancel would pay the copy out of other orders' escrow (reported in review, reproduced, and pinned by `testPlaceOrder_PredispatchRejectsAliasedInputToken`). Output tokens are checked at fill, on the chain they belong to.
- `_execute`: sweeps each output token from the dispatcher once. A repeated output token would otherwise build two sweeps of the same balance and revert every calldata fill.
- Off-chain readers pass the leg index through new SDK helpers `readLegEscrow` and `readLegPartialFill`, used by SDK `isOrderRefunded` and Simplex's source-escrow check and `partialFillsFor`; the SDK, Simplex and indexer ABIs are updated. The helpers call the per-leg getter and, only if that call reverts, the token-keyed `_orders(bytes32,address)` / `_partialFills(bytes32,bytes32)` of a gateway not yet upgraded, remembering per client which one each gateway answers. Token-keyed implementations (live mainnet #1207 and testnet) reject repeated input and output tokens, so a token key there is one leg. Without this, Simplex's live testnet test timed out: the testnet gateways revert `_orders(bytes32,uint256)`, so the filler skipped every order.

## SRLabs audit

| Finding | How this change relates |
|---|---|
| **S3-2** same-chain partial fills over-release repeated input escrow | Implements the auditors' preferred fix, escrow stored and released per entry. `testSameChain_CompletingOneLegReleasesOnlyItsOwnEscrow` |
| **S2-15** partial fills prematurely finalize output legs | Implements the recommended "track fill progress by (commitment, index)". `testSameChain_RepeatedNativeOutputTracksLegsSeparately` covers distinct inputs with a repeated native output |
| **S2-12** fee-on-transfer over-credit | Unchanged. The direct path measures each leg's own transfer, and predispatch orders cannot repeat an input token, so no two legs count the same balance. `testPlaceOrder_PredispatchRejectsRepeatedInputToken` |
| **S2-13** predispatch harvests pre-existing balances (risk accepted) | Unchanged. The report notes duplicate input tokens interact with the predispatch path; predispatch orders still cannot repeat an input token. |
| **S2-14** excess native stuck | `msg.value` accounting and refunds unchanged |
| **S0-10** protocol fee rounds to zero (risk accepted) | Fee is now computed per leg, so splitting into many tiny legs rounds each one down, the same as splitting into separate orders |

The duplicate-token rejections (the "at minimum" mitigation) are not reintroduced, except for input tokens on predispatch orders: with per-leg accounting a repeated token is a valid order shape.

## Rollout

Same slots, different key meaning, so this is only safe with **no in-flight orders**:

- Upgrade only once every commitment placed under the old implementation, on every chain including Tron deployments, has `_orders` (inputs and `TRANSACTION_FEES`) and `_protocolFees` at zero, and no redeem, refund or cancel-proof response is still pending. Expired orders can be closed permissionlessly (same-chain cancel and `cancelFromDest` after the deadline).
- Keep order placement stopped on **all** chains until every chain runs this implementation. The source chain computes the `_partialFills` proof key and the destination stores under it, so mixed versions would prove an empty slot and refund escrow that solvers already earned.
- #1280's `pause()` stops placement, but it also stops `fillOrder`, `onGetResponse` and non-governance `onAccept` deliveries (redeems, refunds, cancel proofs), which draining in-flight orders needs, and deliveries refused while paused would land on this implementation after the upgrade. So `pause()` can bracket the upgrade itself only once the drain is complete; stopping placement on-chain while fills and deliveries keep draining still needs a placement-only switch, which is not part of this PR.
- Getter selectors change to `_orders(bytes32,uint256)`, `_partialFills(bytes32,uint256)` and `_protocolFees(bytes32,uint256)`. Third-party solvers reading them must update with the upgrade.
- No `migrate` and no `VERSION` bump: nothing is copied between slots.
- The Tron copy (`evm/tron/contracts/apps/IntentGatewayV2.sol`) is not updated. It has not tracked #980, #1265 or #1278 either.

## Testing

Mainnet-fork suites, run one at a time:

| Suite | Result |
|---|---|
| `IntentGatewayV2MultiLegTest` (new) | 15 passed |
| `IntentGatewayV2Test` | 120 passed |
| `IntentGatewayV2SameChainTest` | 53 passed |
| `IntentGatewayProtocolFeesTest` | 19 passed |
| `IntentGatewayModulesTest` | 18 passed |
| `IntrinsicIntentsReentrancyTest` | 5 passed (restored the two multi-leg reentrancy tests removed in #980, now with legs sharing an input token) |
| `SolverAccountTest` | 21 passed |

- Existing single-leg tests read leg `0`. `testPlaceOrder_RequiresExactlyOneInputAndOneOutput` became `testPlaceOrder_LegShape`.
- Sensitivity check: disabling the repeated-token sweep skip, or proving leg 0's slot for every leg, each fails its test (`Dai/insufficient-balance`, proof key mismatch).
- SDK `escrowReads.test.ts`: 5 passed (per-leg read, fallback and memo, switch back after an upgrade, no fallback on transport errors, partial fills). A live check against the BSC Chapel and Ethereum gateways (both token-keyed) returned through the fallback. Simplex `pairs` and `fx.curve-payout` suites: 58 passed. Biome lint clean; Simplex `tsc` reports nothing in the changed files.
- Sensitivity check for the token check: without it, a 1,000,000 deposit across two aliased predispatch legs of a `false`-returning token withdraws 2,000,000 on cancel and empties the other order's escrow; removing the input, same-chain output or cross-chain output check each fails its test.
- Rebased onto `main` after #1280: `placeOrder` keeps `whenNotPaused`, and the multi-leg tests initialize through `InitParams`.
- Transient proof index: `testCrossChainCancelFromSource_ManyLegsResolveEachValueByKey` (32 legs, values in scrambled order) and `testCrossChainCancelFromSource_ProofIndexIsClearedBetweenResponses` (a second response in the same transaction cannot resolve its key against the first's values). Skipping the clear, or indexing every key to the first value, fails them.
- Runtime sizes: `IntentGatewayV2` 16,484, `IntrinsicModule` 7,675, `ExtrinsicModule` 17,960 bytes.
- `tsc --noEmit` on `sdk` and `simplex` reports no errors in the changed files.


Wizdave97:
### Hardening: the predispatch duplicate-input-token guard can be bypassed by `bytes32` aliasing

The new guard in `placeOrder` compares the raw `bytes32`:

```solidity
// evm/src/apps/IntentGatewayV2.sol:238
if (order.inputs[j].token == order.inputs[i].token) revert InvalidInput();
```

but every downstream use truncates to 160 bits (`address(uint160(uint256(...)))`), and the high 96 bits of `TokenInfo.token` are never validated anywhere — I grepped `evm/src` and `sdk/packages/core/contracts` for masking and found none. So `T` and `T | (1 << 255)` are two distinct values to the guard and the same ERC-20 to everything else. `_isRepeatedToken` (`IntentsBase.sol:679`) has the identical hole on the output side.

**Not currently exploitable, and I want to be clear about that up front** — see the impact section. Raising it as cheap hardening rather than a blocker.

#### Exploit pathway

The predispatch sweep snapshots **all** of `balancesBefore[i]` in the build loop (`IntentGatewayV2.sol:291`) before any sweep call executes, and `CallDispatcher` checks only the call's success flag, ignoring the ERC-20 return value (`CallDispatcher.sol:59`). So with two aliased input legs:

```
predispatch.assets = [{ T, A }]
predispatch.call   = abi.encode(new Call[](0))          // no-op, length 64 > 0
inputs             = [{ bytes32(uint160(T)),              A },
                      { bytes32(uint160(T)) | (1 << 255), A }]
```

| Step | Dispatcher `T` | Gateway `T` | Escrow credited |
|---|---|---|---|
| Predispatch funding | `A` | `X` | — |
| Build sweeps — both legs read `balanceOf(dispatcher) = A`, both queue `transfer(gateway, A)`, both snapshot `balancesBefore[i] = X` | `A` | `X` | — |
| Sweep call 0 → transfers `A`, returns `true` | `0` | `X + A` | — |
| Sweep call 1 → dispatcher holds `0`, **returns `false` without reverting** | `0` | `X + A` | — |
| Measure: `received[0] = received[1] = (X+A) − X = A` | | | — |
| Credit (`:374`) | | `X + A` | **`2A`** |

Both legs measure the same single arrival as their own. A same-chain cancel (which `order.user` may call at any time) then withdraws `A + A`, the surplus coming out of other orders' pooled escrow in `T`. It generalises: `bytes32` has 96 spare high bits, so `N` aliased legs credit `N × A` against one arrival.

I confirmed this against the branch with a Foundry PoC — deposit 1,000,000, both legs credited 1,000,000, withdraw 2,000,000, gateway's other escrow in that token drained to zero. A negative control also passes: without predispatch the direct path is safe, because it pulls each leg separately with `SafeERC20` and re-reads `balBefore` inside the loop. Happy to push both as regression tests.

#### Impact — bounded, and zero today

It needs a token whose `transfer` returns `false` on insufficient balance instead of reverting (ZRX-class), **and** pooled gateway escrow in that same token. I probed every asset in `chain.ts` against live mainnet by calling `transfer` from an address holding zero:

| Token | Behaviour |
|---|---|
| USDC, USDT, DAI, WETH, cNGN, ZARP, EURC, XSGD, TRYB, USDR | all **revert** — safe |
| ZRX (control) | returns `false` — would be vulnerable |

So **none of the shipped assets can trigger this**, and with a reverting token the second sweep reverts and `placeOrder` simply fails. The exposure is forward-looking: `placeOrder` has no token allowlist and the asset registry is operator-extensible, so this becomes live the day a false-returning token accumulates escrow. Note the guard also silently does nothing for the reverting tokens it nominally protects — they were already failing closed on their own.

#### Why it's cheap hardening

One line at the entry boundary, which also closes the output-side hole in `_isRepeatedToken` and removes the whole malformed-`bytes32` class rather than patching two sites:

```solidity
if (uint256(order.inputs[i].token) >> 160 != 0) revert InvalidInput();
```

It fits in the loop that already walks `inputs` to check for zero output amounts, so it costs one `SHR` and one branch per leg and no extra pass. Comparing truncated addresses in both guards works too, but rejecting malformed input at the boundary is cheaper and means neither guard has to stay correct independently.

Worth doing in this PR: multi-leg orders are what makes the aliasing reachable at all — under the previous `inputs.length == 1` rule no order could have two legs.

Wizdave97 (approved):

royvardhan (approved):

PR #1281, opened 2026-09-17
[simplex-desktop]: add multiplatform release packaging

### LINKED ISSUE #1238: [simplex] Multiplatform desktop builds and release CI

ddboy19912:
## Summary

Closes #1238 by adding reproducible, multi-platform release packaging for Simplex Desktop.

This PR also implements the signing and notarization automation required by #1239. The release pipeline builds the existing Electron application, bundles the verified Node runtime and Simplex solver artifacts, validates the packaged layout, and produces platform-native releases from trusted `simplex-desktop-v*` tags.

Unsigned pull-request builds remain available for CI validation. Credentialed release builds fail closed and run only through a protected GitHub environment.

### Multi-platform packaging

- Added Electron Builder packaging for:
  - macOS arm64 and x64
  - Windows x64
  - Linux x64 and arm64
- Added native artifacts:
  - macOS DMG and ZIP
  - Windows NSIS installer
  - Linux DEB and AppImage
- Bundled the verified Node runtime, Simplex solver, UI, and production dependency closure outside ASAR.
- Added isolated, locked runtime and packaging-tool workspaces.
- Added stable and beta updater metadata.
- Added SHA-512 validation and exact release-asset enforcement.
- Added installed-size enforcement with a 520 MiB limit.
- Centralized solver external dependencies so build and packaging configuration cannot drift.

### Release security

- Restricted signed manual releases to `main`.
- Required release tags to resolve to commits contained in `origin/main`.
- Separated unsigned CI from credentialed release jobs using:
  - `simplex-desktop-ci`
  - `simplex-desktop-release`
- Documented required environment protection rules, approvals, tag rules, secrets, and variables.
- Recreate existing draft releases before publishing so stale assets cannot survive.
- Refuse to modify an already-published release.
- Enforce the exact expected artifact and updater-metadata set.
- Hardened updater trust:
  - Update assets must resolve through pinned HTTPS GitHub release URLs.
  - Windows updates require the configured publisher identity.
  - Linux automatic updates remain disabled until independently authenticated metadata is available.

### macOS signing and notarization

- Added hardened-runtime Developer ID signing for:
  - The Simplex application
  - Electron helper applications
  - The bundled Node runtime
  - The final DMG
- Added explicit minimal entitlements for Electron and Node runtime requirements.
- Disabled automatic entitlement inference.
- Added application notarization and stapling.
- Added separate notarization and stapling for the outer DMG.
- Added CI verification for:
  - Signing identity and Team ID
  - Exact entitlements
  - Nested executables and helpers
  - Application and DMG notarization tickets
  - Gatekeeper assessment of the application and DMG

### Windows signing

- Added Azure Trusted Signing support.
- Restricted the signing endpoint to pinned HTTPS Azure Code Signing hosts.
- Configured SHA-256 file and timestamp digests.
- Signed:
  - The Electron application
  - Every packaged executable
  - The bundled Node runtime
  - Native dependency executables
  - The NSIS installer
- Added recursive Authenticode verification for the installed application and installer.
- Verify the expected publisher identity and a valid trusted timestamp.

### Packaged application validation

- Added native artifact smoke tests covering:
  - Mounting or extracting artifacts
  - Installation
  - Application launch
  - Solver startup
  - Structured logging
  - First-run onboarding
  - Writing `filler-config.toml`
  - Unix config mode `0600`
  - Configured startup failing closed against unavailable endpoints
  - Graceful cleanup
- Added a tag-driven GitHub release workflow for `simplex-desktop-v*`.
- Documented package-local release commands, signing requirements, and artifact layout.

## Verification completed

- [x] Desktop unit suite: 125 passed, 1 platform-specific test skipped
- [x] Socket integration suite: 7 passed
- [x] Release and signing test subset: 26 passed
- [x] Real Electron E2E suite: 10 scenarios passed
- [x] Desktop lint passed
- [x] Desktop typecheck passed
- [x] SDK and Simplex production builds passed
- [x] Release workflow YAML validated
- [x] Solver and Node runtime staging completed successfully
- [x] Local macOS unpacked application smoke test passed
- [x] Local DMG smoke test passed
- [x] Local ZIP smoke test passed
- [x] Installed application size measured at 495.5 MiB, within the 520 MiB limit
- [x] Exact release-asset contract tested
- [x] Pull-request builds confirmed to remain unsigned and secretless
- [x] `git diff --check` passed
- [x] Standards and specification re-review completed with no remaining findings

## Remaining credentialed release validation

The implementation is complete, but the following acceptance checks require the Apple Developer and Microsoft Azure Trusted Signing accounts currently being prepared by operations.

Before running a signed release:

- [ ] Configure the protected `simplex-desktop-release` GitHub environment.
- [ ] Restrict that environment to `main` and `simplex-desktop-v*` tags.
- [ ] Require release-maintainer approval.
- [ ] Add the Apple signing/notarization secrets and Azure Trusted Signing variables to that environment.
- [ ] Protect creation, modification, and deletion of the `simplex-desktop-v*` tag namespace.

After credentials are available:

- [ ] Run a signed manual build from `main`:

      gh workflow run publish-simplex-desktop.yml \
        --ref main \
        -f sign_artifacts=true

- [ ] Confirm all five native jobs pass:
  - macOS arm64
  - macOS x64
  - Windows x64
  - Linux x64
  - Linux arm64
- [ ] Install the signed DMG on a clean Mac and confirm Gatekeeper accepts it without bypass instructions.
- [ ] Install the signed NSIS artifact on a clean Windows machine and confirm Authenticode validation succeeds without a SmartScreen block.
- [ ] Complete onboarding with dedicated credentials and verify a real fill.
- [ ] Create `simplex-desktop-v0.16.2` from the merged `main` commit.
- [ ] Confirm the exact installers and updater metadata are attached to the GitHub release.
- [ ] Confirm the desktop tag does not trigger the Simplex npm or Docker release workflows.

A real fill remains manual because CI does not receive funded signing credentials.

## Issue status

This PR closes #1238.

It implements the engineering work for #1239, but #1239 should remain open until the credentialed signed-build, clean-machine Gatekeeper/SmartScreen, and live-fill validations above are complete.

Closes #1238.


sdk/packages/simplex-desktop/scripts/assert-release-assets.mjs
```diff
@@ -0,0 +1,80 @@
...
+export async function assertReleaseAssets(directory, version) {
+	const channelName = version.includes("-beta.") ? "beta" : "latest"
+	const required = [
+		`Simplex-${version}-mac-arm64.dmg`,
+		`Simplex-${version}-mac-arm64.zip`,
+		`Simplex-${version}-mac-x64.dmg`,
+		`Simplex-${version}-mac-x64.zip`,
+		`Simplex-${version}-win-x64.exe`,
+		`Simplex-${version}-linux-x64.AppImage`,
+		`Simplex-${version}-linux-x64.deb`,
```
seunlanlege: The Linux x64 artifacts won't have these names, so this check fails on the first stable tag.

electron-builder spells the architecture by file type, even with a custom `artifactName`. In 26.15.3, `getArtifactArchName` (builder-util `out/arch.js`) maps:

- x64 AppImage → `x86_64`
- x64 deb → `amd64`
- arm64 AppImage and deb → `arm64`

So the real files are `Simplex-<v>-linux-x86_64.AppImage` and `Simplex-<v>-linux-amd64.deb`. The Windows, macOS and arm64 names do match.

The same names need fixing on line 41 and in the fixtures in `package-release.test.ts`. The test writes its own fake files, which is why it passes.

.github/workflows/publish-simplex-desktop.yml
```diff
@@ -0,0 +1,183 @@
...
+          pnpm --filter="@hyperbridge/simplex" build
+          pnpm --dir packages/simplex-desktop build
+        env:
+          CI: "true"
+
+      - name: Verify and stage target Node runtime
+        run: pnpm --dir packages/simplex-desktop stage:node -- --target ${{ matrix.target }}
+
+      - name: Build native installers and updater metadata
+        run: pnpm --dir packages/simplex-desktop package -- ${{ matrix.builder_args }} --publish never
```
seunlanlege: Beta tags won't produce `beta*.yml` files, so a beta release can't publish.

For the `github` provider, electron-builder ignores the prerelease part of the version. In 26.15.3, `getResolvedPublishConfig` returns the GitHub options without a channel. `computeChannelNames` then falls back to `latest` and says so in a comment: "for GitHub should be pre-release way be used". A `0.17.0-beta.1` build still writes `latest.yml`, `latest-mac.yml` and so on.

That breaks two things:

- `assert-release-assets.mjs` looks for `beta.yml` and fails.
- If it passed, `SimplexReleaseProvider` would request `beta-mac.yml` etc. from the prerelease, and those files wouldn't exist.

One fix is to set the channel explicitly for beta tags:

```yaml
run: pnpm --dir packages/simplex-desktop package -- ${{ matrix.builder_args }} ${{ contains(github.ref_name, '-beta.') && '-c.publish.channel=beta' || '' }} --publish never
```

.github/workflows/publish-simplex-desktop.yml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Refuse a tag that disagrees with the manifests
+        env:
+          RELEASE_TAG: ${{ github.ref_name }}
+        run: pnpm --dir packages/simplex-desktop check:versions && node packages/simplex-desktop/scripts/verify-release-tag.mjs "$RELEASE_TAG"
+
+      - name: Build SDK, solver, UI, and Electron main process
+        run: |
+          pnpm --filter="@hyperbridge/sdk" build
+          pnpm --filter="@hyperbridge/simplex" build
+          pnpm --dir packages/simplex-desktop build
```
seunlanlege: I expect the Windows job to fail here. I haven't run it, so this is from reading the scripts.

`@hyperbridge/simplex` builds through `build.sh` and then `generate-proto.sh`. That script passes `node_modules/.bin/protoc-gen-ts_proto` to protoc as the plugin. On Windows that file is a shell script, and native protoc can't run it. It needs the `.cmd` version, which ts-proto's docs call out.

`test-simplex-desktop.yml` already avoids this. It builds once on Linux, uploads the dist folders, and each OS job only stages Node and runs Electron. Its comment says the output is plain JavaScript and the generated proto files are portable.

I'd use the same shape here:

- one Linux job for install and build
- native jobs that download the dist, then stage Node and package

That also removes the full-workspace `pnpm install` on line 65. The E2E workflow uses `--filter="@hyperbridge/simplex-desktop..."` instead.

.github/workflows/publish-simplex-desktop.yml
```diff
@@ -0,0 +1,183 @@
+name: "Publish Simplex Desktop"
+
+on:
+  push:
+    tags:
+      - "simplex-desktop-v*"
```
seunlanlege, line 17: Right now the only trigger is a tag, so the first real run of this matrix is the `simplex-desktop-v0.16.2` push. If a job fails, recovering means deleting and re-pushing the tag.

Could we add a `pull_request` trigger filtered to the packaging paths? It would run the build and smoke matrix and skip the `release` job, so we'd see all five targets pass before merging.

.github/workflows/publish-simplex-desktop.yml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Install Linux smoke-test display
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y xvfb libfuse2t64
+          sudo chown root:root packages/simplex-desktop/release/*-unpacked/chrome-sandbox
+          sudo chmod 4755 packages/simplex-desktop/release/*-unpacked/chrome-sandbox
+          if sysctl kernel.apparmor_restrict_unprivileged_userns >/dev/null 2>&1; then
+            sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+          fi
```
seunlanlege: This turns off a restriction that stock Ubuntu 24.04+ has on by default. It's also the setting that usually breaks Electron AppImages there: they abort with "The SUID sandbox helper binary was found, but is not configured correctly".

I don't think the other two Linux checks need it:

- The unpacked app already gets a setuid `chrome-sandbox` on lines 94–95.
- electron-builder 26's deb installs an AppArmor profile.

So this mostly lets the AppImage pass in CI when it may not launch for real users. Can we remove it and see how the AppImage does? If it fails, we can either ship only the deb or document `--no-sandbox`.

sdk/packages/simplex-desktop/scripts/package-release.test.ts
```diff
@@ -0,0 +1,212 @@
...
+				version,
+				files: metadata[`${channel}-linux.yml`].map((url) => ({ url, sha512: sha512(url) })),
+			}),
+		)
+		await unlink(join(directory, `Simplex-${version}-linux-arm64.deb`))
+		await expect(assertReleaseAssets(directory, version)).rejects.toThrow()
+	})
+
+	it("requires the release tag to match the package version and namespace", async () => {
+		await expect(verifyReleaseTag("simplex-desktop-v0.16.2")).resolves.toBe("0.16.2")
```
seunlanlege: `verifyReleaseTag` reads the real `package.json`, so this test fails on the next version bump. Can we read the version from the manifest in the test, the way the script does?

Also, `verify-versions.mjs` now makes `runtime/package.json` match the desktop version. That package is private and never published, so the check adds another file to edit on every bump without protecting anything.

sdk/packages/simplex-desktop/scripts/e2e/packaged-smoke.mjs
```diff
@@ -0,0 +1,177 @@
...
+async function waitFor(check, description, timeoutMs = 120_000) {
+	const deadline = Date.now() + timeoutMs
+	let lastError
+	while (Date.now() < deadline) {
+		try {
+			const result = await check()
+			if (result) return result
+		} catch (error) {
+			lastError = error
+		}
```
seunlanlege: Errors thrown by `check` get retried. When the app or solver crashes, `check` throws "exited N" (or `spawnError`), and the loop keeps polling until the 120 s timeout.

A failed launch should fail right away. For example, have `check` return a `fatal` marker, or check `child.exitCode` and `spawnError` outside the try/catch.

sdk/packages/simplex-desktop/scripts/e2e/packaged-smoke.mjs
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+async function assertCleanSolverStartup(userData) {
+	const logsDirectory = join(userData, "logs")
+	const records = await waitFor(async () => {
+		const names = (await readdir(logsDirectory)).filter((name) => /^simplex-[\dT-]+\.log$/.test(name)).sort()
+		if (names.length !== 1) return false
+		const lines = (await readFile(join(logsDirectory, names[0]), "utf8")).trim().split("\n").filter(Boolean)
+		return lines.length > 0 ? lines.map((line) => JSON.parse(line)) : false
```
seunlanlege: This returns once the log has its first line. Warnings logged after that are never checked, so "startup warnings fail the build" only covers the first lines of the log.

Could we wait for a known startup-complete record, or re-read the log after `/health` has come back ok?

.github/workflows/publish-simplex-desktop.yml
```diff
@@ -0,0 +1,183 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          VERSION="${GITHUB_REF_NAME#simplex-desktop-v}"
+          PRERELEASE=""
+          case "$VERSION" in *-beta.*) PRERELEASE="--prerelease" ;; esac
+          if gh release view "$GITHUB_REF_NAME" >/dev/null 2>&1; then
+            gh release upload "$GITHUB_REF_NAME" release-assets/* --clobber
+          else
+            gh release create "$GITHUB_REF_NAME" release-assets/* --verify-tag --generate-notes $PRERELEASE --title "Simplex Desktop v${VERSION}"
```
seunlanlege: Question, not a blocker. Until #1239 adds signing, this publishes unsigned builds as normal stable releases.

On macOS that has two effects:

- Downloaded builds hit Gatekeeper warnings.
- Installed copies can't auto-update to the later signed builds, because the updater requires the running app to be signed.

Early Mac users would need to reinstall by hand. Should these releases be prerelease or draft until #1239 lands?

seunlanlege (commented):
I checked the packaging against electron-builder 26.15.3's source. I think the first tag push would fail for two confirmed reasons:

- The Linux x64 artifact names don't match what `assert-release-assets.mjs` expects.
- Beta tags never produce `beta*.yml` files.

The Windows job will also very likely fail while building simplex. Details are inline.

These things look fine:

- The packaged resource layout matches `resourcePaths` and the tray-icon paths.
- Moving `@hyperbridge/simplex` to devDependencies is safe.
- Removing `tagNamePrefix` is covered by `SIMPLEX_UPDATE_FEED`.
- The merged `latest-mac.yml` and the SHA-512 checks are correct.

Wizdave97:
Reviewed the packaging pipeline at `02f10c60`, then re-checked against `b2ace0e9` after the signing commit landed. Two things I found are already resolved by that commit; one residual item and one process note remain.

### Resolved by `b2ace0e9` (noting for the record)

At `02f10c60` the release job's `pattern: simplex-desktop-*` also matched the prepare job's `simplex-desktop-release-inputs` artifact, so `assemble-release.mjs` would have ingested `sdk/dist/**`, `simplex/dist/**` and `simplex-desktop/dist/**` alongside the installers and died on `Duplicate release artifact: index.js` (`dist/browser/index.js` vs `dist/node/index.js` flatten to the same basename). Renaming the artifact to `simplex-portable-release-inputs` fixes it, and the new "Unexpected release assets" check in `assert-release-assets.mjs` is a good second line of defence. Since the `release` job is tag-gated it is skipped on every PR run, so this would only have surfaced on the first real `simplex-desktop-v*` push.

### Residual: the runtime macOS gate does not pin the team ID

`updateAuthenticityForInstallation` enables the updater when `codesign --verify --deep --strict` succeeds. That establishes the bundle has *a* valid seal, not that it is Polytope's. `verify-macos-signatures.mjs` gets this right at release time by checking `TeamIdentifier` against `APPLE_TEAM_ID`; the runtime check does not apply the same pin.

With `mac.identity: null` on the unsigned path this is much narrower than it was — electron-builder now skips macOS signing entirely rather than falling back to an ad-hoc identity, so builds from this repo fail closed correctly. The remaining case is a bundle validly signed by someone else (a fork's Developer ID build, or a re-signed app), where updates would enable and then fail at install because Squirrel.Mac requires the update signature to match the running app. Applying the same requirement the release verifier uses would keep the two checks consistent:

```
codesign --verify --strict -R '=anchor apple generic and certificate leaf[subject.OU] = <TEAM_ID>' <bundle>
```

Non-blocking.

### Process note: the signed path is unexercised while the release is now public

`gh release edit --draft=false` replaces the previous draft-only behaviour, so a tag push now goes straight to the live updater feed. Everything that path depends on — `release-signing.cjs`'s signed branch, `notarize-macos-dmg.mjs`, `verify-macos-signatures.mjs`, `verify-windows-signatures.ps1`, and `signPackagedWindowsRuntime` in the afterPack hook — only runs when `sign_artifacts` is true, which no PR run can reach. The first tag push would be both the first execution of that code and a live push to every installed client.

The failure modes mostly fail safe (a failed verify step fails the matrix job, and `release` needs `build`), and the workflow already has the right escape hatch. Worth running `workflow_dispatch` on `main` with `sign_artifacts: true` once and confirming the macOS and Windows verify steps pass before cutting `simplex-desktop-v0.16.2`.

### Checked and holding up

- `resolveTrustedReleaseFiles` is correct including the case the tests miss: `resolveFiles` applies the same path transformer to `updateInfo.packages[arch].path`, so the nsis-web `packageInfo` escape hatch is constrained too.
- `artifactName` has no spaces, so `RELEASE_ARTIFACT_NAME` cannot reject a legitimate installer; the Linux `x86_64.AppImage` / `amd64.deb` naming matches electron-builder's target-specific arch names.
- The narrowed upload globs (`*.zip.blockmap` / `*.exe.blockmap`) line up exactly with `releaseAssetContract`'s required set, so the new unexpected-assets check should not false-positive on `.dmg.blockmap`.
- `simplex-v[0-9]+...` cannot match `simplex-desktop-v*`, so desktop tags genuinely do not trigger the npm or Docker workflows.
- Merging both mac archs into one `latest-mac.yml` matches electron-builder's own behaviour, and `MacUpdater` filters `files` by arch, so the retained top-level `path` is never used for resolution.
- The tsup `createRequire` banner works as intended — esbuild's `__require` helper probes `typeof require !== "undefined"` and picks up the banner binding, with no name collision.

sdk/package.json
```diff
@@ -53,20 +53,6 @@
...
-			"utf-8-validate",
-			"esbuild"
-		],
-		"overrides": {
-			"viem": "2.47.6",
-			"axios": "1.13.6",
-			"vite": "6.4.2",
-			"pino": "~10.3.1"
-		}
-	},
```
seunlanlege, line 69: These pins are now gone from the repo, and I don't think that was the intent.

Following the commits in order:

- `52250765` moved them to `pnpm-workspace.yaml`, which is where pnpm 11 reads them, and the lockfile picked up an `overrides:` section.
- `abfe8579` deleted them from both files to get frozen installs passing.

So nothing pins them any more. The lockfile currently resolves axios to 14 different versions, and holds both vite 6.4.2 and 8.0.16.

The middle commit shows the migration works. It only failed because applying overrides changes resolution, which needs a deliberate lockfile regeneration.

Could we keep the pins in `pnpm-workspace.yaml` and regenerate the lockfile in its own PR, so that churn gets reviewed on its own? Dropping `onlyBuiltDependencies` is fine, since `allowBuilds` replaces it.

sdk/packages/simplex-desktop/scripts/workspace-policy.test.ts
```diff
@@ -0,0 +1,58 @@
...
+		expect(workspace.minimumReleaseAgeExclude).toEqual(
+			expect.arrayContaining([
+				"@base-ui/react@1.8.0",
+				"@base-ui/utils@0.4.0",
+				"electron@44.3.0",
+				"playwright-core@1.63.0",
+			]),
+		)
+		expect(workspace).not.toHaveProperty("overrides")
+		expect(manifest).not.toHaveProperty("pnpm")
```
seunlanlege: These two assertions lock in the removal of the overrides. Anyone who re-pins viem, axios, vite or pino in the supported pnpm 11 location will fail a desktop test, with nothing explaining why.

Can we drop both lines? The rest of this test, checking that lockfile specifiers match each manifest, is worth keeping.

sdk/pnpm-workspace.yaml
```diff
@@ -14,9 +14,16 @@ allowBuilds:
...
+  # ws treats this as an optional performance dependency. Node 24 arm64 has no
+  # compatible prebuild, and compiling it is not required by the desktop app.
+  utf-8-validate: false
 blockExoticSubdeps: true
 minimumReleaseAge: 43200
+minimumReleaseAgeExclude:
+  - '@base-ui/react@1.8.0'
+  - '@base-ui/utils@0.4.0'
+  - 'electron@44.3.0'
+  - 'playwright-core@1.63.0'
```
seunlanlege: Only `electron@44.3.0` belongs to this PR. The other three exempt unrelated packages from the 30-day quarantine that `minimumReleaseAge` exists to enforce.

If they're only blocking because they're recent, waiting out the window costs nothing. If something needs them now, a comment saying why would help the next person decide when to remove them.

Worth noting too that `workspace-policy.test.ts` hardcodes this exact list, so removing an entry later fails a test in another package.

sdk/packages/simplex-desktop/src/update-authenticity.ts
```diff
@@ -0,0 +1,90 @@
...
+	if (options.platform === "darwin") {
+		const bundlePath = macApplicationBundlePath(options.executablePath)
+		const requirement = options.expectedMacTeamId ? macCodeSigningRequirement(options.expectedMacTeamId) : undefined
+		if (!requirement) {
+			return { enabled: false, reason: "The installed macOS application has no trusted Apple Team ID" }
+		}
+		const verify =
+			options.verifyMacSignature ??
+			((path: string, signingRequirement: string) =>
+				spawnSync("/usr/bin/codesign", ["--verify", "--deep", "--strict", "-R", signingRequirement, path], {
```
seunlanlege: This runs on every launch, on the main thread, through `spawnSync`. Deep-verifying a ~500 MB Electron bundle takes seconds, and the tray and menus are frozen for all of it.

The `-R` requirement is what actually pins the Team ID, and it doesn't need `--deep`. Gatekeeper already verified the nested code at first launch.

Two options: drop `--deep`, or run the check asynchronously and enable updates when it returns.

sdk/packages/simplex-desktop/scripts/release-signing.cjs
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if (platform === "win32") {
+		config.forceCodeSigning = true
+		config.win = {
+			...config.win,
+			signExecutable: true,
+			signExts: [".exe"],
+			azureSignOptions: {
+				publisherName: signing.publisherName,
```
seunlanlege, line 129: The Windows update trust root ends up depending on this line in a way that isn't visible from the runtime check.

`publisherName` reaches `app-update.yml` only through `getAppUpdatePublishConfiguration`, which fills it from `azureSignOptions.publisherName`, and only because `verifyUpdateCodeSignature` isn't false. I traced it in electron-builder 26.15.3 and it works today. But a later config change would silently disable Windows updates for every new install. It fails closed, so it isn't dangerous, just invisible.

Could the release job assert that the generated `app-update.yml` has a non-empty `publisherName`? That's the value the whole Windows gate rests on.

Separately, `signExts: [".exe"]` matches electron-builder's default behavior, so it changes nothing. Electron's DLLs stay unsigned either way, which the verify script also doesn't check. Fine if that's intended.

sdk/packages/simplex-desktop/package.json
```diff
@@ -2,27 +2,38 @@
...
+		"test:unit": "vitest --watch=false src/tests/blackhole-server.test.ts src/tests/desktop-paths.test.ts src/tests/daemon.test.ts src/tests/desktop-security.test.ts src/tests/electron-launch.test.ts src/tests/login-item.test.ts src/tests/release-provider.test.ts src/tests/solver-supervisor.test.ts src/tests/tray-icon.test.ts src/tests/tray-menu.test.ts src/tests/update-authenticity.test.ts src/tests/update-coordinator.test.ts src/tests/update-store.test.ts scripts/package-release.test.ts scripts/release-signing.test.ts scripts/stage-node.test.ts scripts/workspace-policy.test.ts",
 		"test:integration": "vitest --watch=false src/tests/protocol.test.ts",
 		"lint": "biome lint src scripts",
 		"format": "prettier --write \"{src,scripts}/**/*.{ts,mjs}\"",
 		"typecheck": "tsc --noEmit"
 	},
 	"dependencies": {
-		"@hyperbridge/simplex": "workspace:*",
 		"electron-updater": "6.8.9"
```
seunlanlege: Now that tsup bundles electron-updater into `dist/main.js`, keeping it in `dependencies` means electron-builder also packs a full copy into `app.asar`. Moving it to `devDependencies`, the way `@hyperbridge/simplex` was moved, drops the unused copy.

seunlanlege (commented):
Reviewed the signing and updater-trust commits. The signing work looks careful, and I checked the parts I doubted against electron-builder 26.15.3 rather than assuming:

- `mac.binaries` entries resolve relative to the `.app`, so `Contents/Resources/runtime/node` is right.
- `signIf` exists on the Windows packager, and `node.exe` matches the sign filter.
- electron-builder notarizes and staples the `.app` before the DMG is built, so `stapler validate` on the unpacked app will pass.
- `resolveTrustedReleaseFiles` correctly pins scheme, origin and path prefix, and rejects credentials, queries and fragments.

Draft-then-publish, and refusing to overwrite a published release, are both good changes.

My comments are mostly about shared workspace config that changed along the way, particularly the dependency overrides. One note on the release process: the signed path only runs on a tag or a maintainer dispatch from `main`, so PR CI never exercises it. I'd treat the signed dry run the README documents as required before the first tag, not optional.

.github/workflows/publish-simplex-desktop.yml
```diff
@@ -0,0 +1,352 @@
...
+          path: |
+            sdk/packages/simplex-desktop/release/*.dmg
+            sdk/packages/simplex-desktop/release/*.zip
+            sdk/packages/simplex-desktop/release/*.exe
+            sdk/packages/simplex-desktop/release/*.AppImage
+            sdk/packages/simplex-desktop/release/*.deb
+            sdk/packages/simplex-desktop/release/*.zip.blockmap
+            sdk/packages/simplex-desktop/release/*.exe.blockmap
+            sdk/packages/simplex-desktop/release/latest*.yml
+            sdk/packages/simplex-desktop/release/beta*.yml
```
seunlanlege, line 298: Small note, no change needed. Linux auto-updates are disabled at runtime, so the `latest-linux*.yml` files published here are never read by an installed app.

Keeping them seems right, since signed Linux metadata would need them later. Just flagging that the asset check enforces a file nothing currently consumes.

seunlanlege (approved):
Approving. All five native packaging targets and the desktop suites pass on 1ba5795c, and the findings from both reviews are addressed in the code.

Two things to carry forward after merge:

- The signed path has never run. Do a signed `workflow_dispatch` from main and confirm it is green before tagging simplex-desktop-v0.16.2.
- The axios pin still needs to come back at a patched version, tracked separately.

PR #1282, opened 2026-09-17
feat(intents): require solver rate quotes for every fill

royvardhan:
Every fill declares a quote per order leg: `FillOptions.inputs[i]` is the most input the solver will
take and `FillOptions.outputs[i]` the most output it will pay. Their ratio is the solver's rate and
must not be below the order's own rate. A leg is skipped by quoting zero on both sides; an `inputs`
array of the wrong length is rejected.

Settlement lives in one path in `IntentsBase` for same-chain and cross-chain fills. Credited output
is the take priced at the order's rate, capped to what the leg still needs. Released input is the
difference of cumulative floors, so a completing fill releases exactly the remaining escrow. Payment
is the credited output or the released input priced at the solver's rate, whichever is higher; the
difference is surplus, split by `surplusShareBps` or kept entirely by the protocol for output-call
orders. Unused ERC-20 budget stays with the solver and unused native value is refunded. Events and
cross-chain proofs carry credited output only.

The gateway, both modules and `SolverAccount` are release 3. `fillOrder` moves to selector
`0x68ddf058`; the account also refuses the two historical selectors on the plain-signature path.
`migrate(owner)` takes a version-2 proxy to 3, shifting the relayer slot and setting the owner;
nothing has shipped at 3 yet, so the module split, the owner and solver quotes land together. The
indexer ABI is regenerated and its fill decoder pins the two deployed `FillOptions` shapes so
historical fills still decode.

All seven mainnet gateways are at version 2, so the live upgrade path is 2 to 3. Before upgrading,
finish or cancel outstanding orders and drain escrow, fees and pending messages on every chain, then
install matching gateways, modules and account delegations before enabling bids. Partially filled
orders from the current per-slice rounding are not migrated; completing one under cumulative
accounting can leave dust that the completed-order guard blocks from cancellation.

Stack: this PR, then #1287 (client quotes and ranking), then #1288 (submission recovery).

Addresses #1259.


seunlanlege:
We should modify the existing fillOrder

sdk/packages/simplex/src/strategies/fx.ts
```diff
@@ -860,13 +861,25 @@ export class FXFiller implements FillerStrategy {
...
 				// of the same token share one wallet balance.
 				const walletRemaining = balance - walletContribution
 				balanceCache.set(tokenAddress, walletRemaining > 0n ? walletRemaining : 0n)
 
+				if (rateFills) {
+					// Cap exposure in input units, then scale the take if funding shortened
+					// the output quote. Ceiling avoids silently improving our signed price.
+					const budgetTake = BigInt(new Decimal(input.amount.toString()).mul(capFraction).floor().toFixed(0))
+					policyInputByLeg.set(i, budgetTake)
+					const take = (budgetTake * finalOutputAmount + policyMaxOutput - 1n) / policyMaxOutput
```
Wizdave97: **`take` is rounded in the direction that trips `RateBelowOrder`.**

The comment says ceiling "avoids silently improving our signed price", but `take` is the *input* the solver claims — a larger `take` means more of the user's escrow for the same `offered`, i.e. a better price for us. `RateFillMath.quote` enforces `ceil(take * required / escrow) <= offered`, so rounding `take` up is precisely what breaches it.

Worked counterexample against the real formulas (`escrow=1000, required=99, policyMaxOutput=99, finalOutputAmount=37, capFraction=1`):

- ceiling: `take = 374` -> `ceil(374*99/1000) = 38 > 37` -> **`RateBelowOrder`**
- floor: `take = 373` -> accepted (`credit=36, release=363, delivered=37`)

Because `previewRateFill` throws rather than returning, the exception unwinds to the generic `catch` at the end of `calculateProfitability` and the order is logged as "Error calculating profitability" and scored `0` — so this shows up as a silently skipped order with a misleading log line, not as a visible failure.

Suggest floor, and updating the comment. Note there is currently no test covering a funding-shortfall rate quote (`finalOutputAmount < policyMaxOutput`), which is the case that triggers this.

sdk/packages/sdk/src/protocols/intents/Bid.ts
```diff
@@ -207,18 +238,45 @@ export class BidImpl implements Bid {
...
+				`Bid submission attempt could not be recorded durably: ${err instanceof Error ? err.message : String(err)}`,
+			)
+		}
+		try {
+			await this.crypto.sendBundler<HexString>(BundlerMethod.ETH_SEND_USER_OPERATION, [
+				CryptoUtils.prepareBundlerCall(signedUserOp),
+				entryPointAddress,
+			])
+		} catch (err) {
+			throw new BidExecutionPendingError(
```
Wizdave97: **Deterministic bundler rejections are classified as "uncertain", which aborts the whole order.**

`CryptoUtils.sendBundler` throws a plain `Error` for any JSON-RPC error body, so this `catch` wraps *every* `eth_sendUserOperation` failure in `BidExecutionPendingError` — including deterministic, safe-to-skip rejections like `AA25 invalid account nonce`, `AA24 signature error` and `AA21 didn't pay prefund`.

That classification propagates: `BidManager.selectAndExecuteBest` rethrows instead of trying the next bid, and `OrderExecutor` yields `FAILED` and **returns**. One malformed bid from one solver now terminates the entire auction, where previously `autoSelect` swallowed it and kept polling to the deadline.

This is made likely rather than theoretical by the new `isUserOperationNonceConsumed` pre-check — it is a TOCTOU read, so a bid whose nonce is consumed between the poll and the submit produces exactly `AA25` and therefore a permanent `FAILED`.

Worth distinguishing "the bundler rejected this deterministically" (skip the bid, continue) from "we don't know whether it landed" (stop safely), rather than treating every send error as the latter.

sdk/packages/sdk/src/protocols/intents/Bid.ts
```diff
@@ -228,45 +286,80 @@ export class BidImpl implements Bid {
...
+			throw new Error(`UserOperation failed in confirmed transaction ${txnHash}`)
+		}
+
+		try {
+			const events = parseEventLogs({
+				abi: IntentGatewayV2ABI,
+				logs: chainReceipt.logs,
+				eventName: ["OrderFilled", "PartialFill"],
+			})
+			const matched = events.find((e) => {
```
Wizdave97: **The fill event isn't bound to our filler.**

This matches `OrderFilled`/`PartialFill` on gateway address + commitment only. `e.args.filler` is a non-indexed param and is decodable here, but it is never checked against `this.solverAddress`, and the log index is not correlated with our `UserOperationEvent`.

Bundlers batch multiple userOps into one transaction. If another solver's op for the same commitment lands in the same bundle, `find` returns *their* `PartialFill` and `filledAssets` is populated with another solver's credited amounts — then attributed to our `selectedSolver` in the emitted `PARTIAL_FILL` update.

The on-chain `readCreditedProgress` reconciliation on the next loop iteration does repair the totals (it takes the max of cumulative values, so there's no double-count), but the update we emit to consumers is wrong in the meantime.

Adding `e.args.filler.toLowerCase() === this.solverAddress.toLowerCase()` to the predicate closes it.

Wizdave97 (commented):
Three findings from an adversarial review pass (AI-assisted; each verified against the code, and the surrounding settlement math in `RateFillMath`/`_fillLegs` held up under fuzzing, so these are all off-chain).

Flagging only these three here;

Wizdave97:
**CI is red on this branch, and the cause is in this PR rather than flakiness.**

The `@hyperbridge/sdk` workflow [fails](https://github.com/polytope-labs/hyperbridge/actions/runs/35220779067/job/105200108008) at the "Run simplex test" step: **7 of 99 tests fail** in `sdk/packages/simplex/src/tests/pairs.test.ts`, all with `AssertionError: expected 0 to be greater than 0` from `await filler.calculateProfitability(o)` — lines 838, 897, 993, 1011, 1089, 1191 and 1237. That file is not modified by this PR, and the same workflow is green on every recent `main` commit.

**Root cause** (reproduced locally by instrumenting the `catch` at the end of `calculateProfitability`):

```
TypeError: this.contractService.rateFillsSupported is not a function
    at FXFiller.calculateProfitability (src/strategies/fx.ts:541:79)
```

`fx.ts:541` now calls `await this.contractService.rateFillsSupported(destChain)`. The `gateFiller` contract-service stub used by the "FXFiller profit gates" suite (`pairs.test.ts:759-779`) does not define that method. `fx.curve-payout.test.ts:58` got `rateFillsSupported: async () => false`, but this stub did not.

**Verified fix:** adding `rateFillsSupported: async () => false` to that one stub takes the file from 7 failed / 48 passed to **55 passed**.

All 7 failures are single-input orders — the `order.inputs.length === 1 &&` short-circuit at `fx.ts:541` is why the multi-leg tests in the same suite are unaffected.

**Worth addressing beyond the stub:** `calculateProfitability`'s blanket `catch { ... return 0 }` converts any programming error into "unprofitable, skip". A `TypeError` on a missing method produced seven silent zeros rather than a visible failure, in the main strategy suite. Rethrowing `TypeError` (or otherwise narrowing that catch) would make the next gap of this shape fail loudly.

Minor note: the PR description reports "**47 Simplex tests** ... passed", which doesn't match the 99 collected in this job — likely a narrower local run.

_(AI-assisted review; root cause reproduced locally rather than inferred.)_

seunlanlege:
Also merge conflicts

evm/src/apps/intentsv2/ExtrinsicIntents.sol
```diff
@@ -157,109 +157,33 @@ abstract contract ExtrinsicIntents is IntentsBase, HyperApp {
      * @param options Fill options including output amounts, relayer fee, and native dispatch fee.
      * @param commitment The keccak256 hash of the ABI-encoded order.
      */
-    function _fillCrossChain(Order calldata order, FillOptions calldata options, bytes32 commitment) internal {
-        uint256 outputsLen = order.output.assets.length;
-
+    function _fillCrossChain(
+        Order calldata order,
+        FillOptions calldata options,
```
seunlanlege: lets add inputs to `FillOptions`

evm/src/apps/IntentGatewayV2.sol
```diff
@@ -480,6 +480,25 @@ contract IntentGatewayV2 is
...
      */
     function fillOrder(Order calldata order, FillOptions calldata options) public payable whenNotPaused nonReentrant {
+        _fillOrder(order, options, new TokenInfo[](0));
+    }
+
+    /// @notice Whether this implementation accepts solver-priced fills.
+    function supportsRateFills() external pure returns (bool) {
+        return true;
+    }
+
```
seunlanlege: unneeded, we'll use the verstion number

evm/src/apps/intentsv2/RateFillMath.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.24;
+
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+/// @dev No fixed-point rates: the solver signs input/output raw amounts.
+library RateFillMath {
```
seunlanlege: This can probably live in `IntentsBase` right?

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -461,6 +466,114 @@ abstract contract IntentsBase is EIP712 {
...
 
+    struct FillResult {
+        TokenInfo[] inputs;
+        TokenInfo[] outputs;
+        bool complete;
+        uint256 nativeRemaining;
+    }
+
+    /// @dev Shared destination-side loop. Credit, not gross payment, drives settlement/proofs.
+    function _fillLegs(
```
seunlanlege: is there any chance of making this more human-readable?

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -465,6 +470,176 @@ abstract contract IntentsBase is EIP712 {
...
+        uint256 nativeRemaining;
+    }
+
+    /// @dev Records each leg's credited output and delivers its payment. The caller settles
+    /// released inputs locally or sends them to the source chain in a redemption request.
+    function _fillLegs(Order calldata order, FillOptions calldata options, bytes32 commitment, bool sameChain)
+        internal
+        returns (FillResult memory result)
+    {
+        bool useInputQuotes = options.inputs.length != 0;
```
seunlanlege: Solvers must always declare their rates. Make `FillOptions.inputs` required and delete this branch.

A quote already expresses every ordinary fill. Taking the whole escrow is a full fill, and a quote larger than what remains is capped to the remainder at the same rate. Keeping the empty case gives us two settlement rules for one behaviour: `_validateRateLegs`, the second half of `_fillAmounts`, the `sameChain` plumbing and the legacy-rounding check all exist only to serve it.

There is no compatibility argument either. This PR changes the selector, so every bid is signed afresh against the new ABI regardless.

evm/src/apps/IntentGatewayV2.sol
```diff
@@ -536,6 +538,7 @@ contract IntentGatewayV2 is
         uint256 outputsLen = order.output.assets.length;
         if (options.outputs.length != outputsLen) revert InvalidInput();
         if (order.inputs.length != outputsLen) revert InvalidInput();
+        if (options.inputs.length != 0 && options.inputs.length != outputsLen) revert InvalidInput();
```
seunlanlege: With rates always declared this is just `if (options.inputs.length != outputsLen) revert InvalidInput();`.

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -465,6 +470,176 @@ abstract contract IntentsBase is EIP712 {
...
+        uint256 previousCredit,
+        bool sameChain
+    ) private view returns (uint256 creditedOutput, uint256 releasedInput, uint256 deliveredOutput) {
+        uint256 escrowInput = order.inputs[legIndex].amount;
+        uint256 requiredOutput = order.output.assets[legIndex].amount;
+        uint256 offeredOutput = options.outputs[legIndex].amount;
+        uint256 quotedInput = options.inputs.length == 0 ? 0 : options.inputs[legIndex].amount;
+        uint256 previouslyReleased = _cumulativeReleased(escrowInput, previousCredit, requiredOutput);
+        uint256 remainingEscrow = sameChain ? _orders[commitment][legIndex] : 0;
+        bool hasLegacyRounding = sameChain && remainingEscrow != escrowInput - previouslyReleased;
```
seunlanlege: This detection goes away with mandatory rates, along with `LegacyRateAccounting` and the `_orders` read above it.

It serves only same-chain orders left partially filled by main's per-slice rounding. Those need main deployed somewhere, and the live gateways are token-keyed and have to be drained before this upgrade anyway. Even if such an order existed, per-slice floors always sum to at most the cumulative floor, so the cumulative formula can never over-release: worst case a few wei stay in escrow, and cancellation still refunds them.

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -465,6 +470,176 @@ abstract contract IntentsBase is EIP712 {
...
+
+        if (quotedInput > 0) {
+            if (hasLegacyRounding) revert LegacyRateAccounting();
+            return _quoteRateFill(escrowInput, requiredOutput, previousCredit, quotedInput, offeredOutput);
+        }
+
+        creditedOutput = Math.min(offeredOutput, requiredOutput - previousCredit);
+        // Ordinary fills accept surplus only on the first fill of a leg.
```
seunlanlege: This rule exists only because one number carries both the size and the price of an ordinary fill, so the gateway cannot tell a generous price from a stale full-size bid. When the solver declares a rate, take and payment are separate numbers and surplus applies on every fill. One more reason to require `inputs`.

seunlanlege (commented):
First pass, and it is all one theme: solvers must always declare their rates.

`FillOptions.inputs` should be required, not optional. The quote path already covers every ordinary fill, so the empty-array path is a second settlement rule for behaviour we already have, and it drags the legacy-rounding special case along with it. Comments inline on what making rates mandatory removes.

sdk/packages/core/contracts/apps/IntentGatewayV2.sol
```diff
@@ -179,8 +179,11 @@ struct FillOptions {
     /// same clock (`_blockNumber()`, which is the L2 block number where that differs).
     uint256 validUntil;
     /// @dev The output tokens with amounts the solver is willing to give
-    /// @dev Must be strictly >= the amounts requested in order.output.assets
     TokenInfo[] outputs;
+    /// @dev Positional maximum input takes, matched to order.inputs and outputs by index.
+    /// Integer output credit determines actual release, which can be below these maxima.
+    /// An empty array uses the order rate, including settlement of legacy rounding debt.
```
seunlanlege: And this line goes with it: there is no empty-array case. Every fill declares a rate per leg.

sdk/packages/sdk/src/protocols/intents/OrderExecutor.ts
```diff
@@ -307,13 +553,48 @@ export class OrderExecutor {
 				if (done) return
 
 				const fed = yield value
-				if (value.status === "BIDS_RECEIVED") input = fed
+				if (value.status === "BIDS_RECEIVED") {
+					if (automatic) {
+						const executableBids =
+							order.inputs.length === 1 && order.output.assets.length === 1
+								? value.bids
+								: value.bids.filter((bid) => bid.inputs.length === 0)
```
seunlanlege: With rates mandatory this filter has nothing left to select, since every multi-leg bid carries takes. Rank a bid that takes the full escrow on each leg the way we rank today and the single-leg restriction disappears.

seunlanlege:
@royvardhan can you split this one up?

Keep this branch to the contract change: `evm/src`, `evm/tests`, the `IntentGatewayV2.sol` interface in `sdk/packages/core`, and the regenerated ABI artifacts, since those are compiled output of the same change. Move `sdk/packages/sdk`, `sdk/packages/simplex` and the indexer decode onto a second branch stacked on this one.

Right now it is roughly 1,300 lines of contracts and tests against roughly 5,000 lines of SDK, Simplex and indexer, and the two halves carry different risk. One is an upgrade we have to get right on chain behind a governance call. The other ships on npm and can iterate.

While you are at it: the durable submission journal, the restart reconciliation and the EntryPoint receipt attribution are not about solver quotes at all. They fix a genuine double-fill risk and deserve their own review rather than riding along with a pricing change. Their own branch would be best; separate commits at minimum.

One thing that affects the split: see my review comments, I want `FillOptions.inputs` to be required rather than optional. That drops the empty-inputs handling from the SDK and Simplex too, so it is worth settling the contract shape here before the SDK branch gets much further.

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -465,6 +470,176 @@ abstract contract IntentsBase is EIP712 {
...
+        result.releasedInputs = new TokenInfo[](legCount);
+        result.creditedOutputs = new TokenInfo[](legCount);
+        result.fullyFilled = true;
+        result.nativeRemaining = msg.value;
+        bool madeProgress;
+
+        for (uint256 i; i < legCount; ++i) {
+            bytes32 outputToken = order.output.assets[i].token;
+            if (uint256(outputToken) >> 160 != 0) revert InvalidInput();
+            if (options.outputs[i].token != outputToken) revert InvalidInput();
```
seunlanlege: Move these into `_validateRateLegs` and rename it `_validateLegs`, since it is no longer only about rates.

With `inputs` required it runs on every fill anyway, so every per-leg shape check belongs in that one pass and this loop is left doing nothing but accounting. It already repeats the output high-bit check on line 561, so that duplication goes with it; the output token match is the only part it does not have yet.

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -465,6 +469,130 @@ abstract contract IntentsBase is EIP712 {
...
+
+        // Round the minimum payment up so the quote cannot fall below the order's price.
+        uint256 minimumOutput = Math.mulDiv(quotedInput, requiredOutput, escrowInput, Math.Rounding.Ceil);
+        if (minimumOutput > offeredOutput) revert RateBelowOrder();
+
+        uint256 quotedCredit = Math.mulDiv(quotedInput, requiredOutput, escrowInput);
+        uint256 remainingOutput = requiredOutput - previousCredit;
+        creditedOutput = Math.min(quotedCredit, remainingOutput);
+
+        // Subtract cumulative floors so the completing fill releases all remaining escrow.
```
seunlanlege: `_fillAmounts` still treats `options.inputs` as optional. The ternary should go: read `options.inputs[legIndex].amount` directly.

The `quotedInput > 0` branch below then becomes the only path, so the ordinary-settlement tail of this function goes with it and `_fillAmounts` collapses into a call to `_quoteRateFill`. A zero take is already a skipped leg, rejected earlier by the validation pass unless the output is zero too.

evm/src/apps/intentsv2/IntentsBase.sol
```diff
@@ -465,6 +469,130 @@ abstract contract IntentsBase is EIP712 {
...
+        uint256 previouslyReleased = Math.mulDiv(escrowInput, previousCredit, requiredOutput);
+        uint256 cumulativeRelease = Math.mulDiv(escrowInput, previousCredit + creditedOutput, requiredOutput);
+        releasedInput = cumulativeRelease - previouslyReleased;
```
seunlanlege: Delete the legacy-rounding handling outright: this read and flag, `LegacyRateAccounting`, and the branch they guard in `_fillAmounts`.

We are taking a breaking migration anyway. Escrow, fees and pending messages get drained before the upgrade, so there are no part-filled orders carrying old rounding debt to rescue.

Even if one did slip through, the debt is at most about one unit of the input token per earlier fill. On a 6-decimal stablecoin that is a millionth of a dollar. It does not justify a special case in the fill path, a second settlement rule, or an error solvers have to handle.

seunlanlege (commented):
Follow-ups on making rates mandatory: one validation pass for every per-leg check, `_fillAmounts` reading `inputs` unconditionally, and the legacy rounding handling deleted rather than special-cased.

evm/src/apps/IntentGatewayV2.sol
```diff
@@ -88,9 +88,8 @@ contract IntentGatewayV2 is
     address public immutable extrinsicModule;
 
     /// @dev The `Initializable` version this implementation lands a proxy on, through `initialize`
-    /// or `migrate`. 3 is the module split and the owner; bumped by every implementation that ships
-    /// a `migrate`.
-    uint64 private constant VERSION = 3;
+    /// or `migrate`. Version 4 adds solver input quotes to FillOptions.
+    uint64 private constant VERSION = 4;
```
seunlanlege: Keep this at 3. We have not upgraded anything to version 3 yet, so the module split, the owner and solver quotes all land on proxies together as one release.

`version() == 3` then means quotes are supported, which is the signal the SDK and the account need, and no proxy in the wild can answer 3 without them.

evm/src/apps/IntentGatewayV2.sol
```diff
@@ -188,24 +187,30 @@ contract IntentGatewayV2 is
...
+        if (previousVersion == 3 && owner() == address(0)) revert InvalidInitialization();
+        _migrate(previousVersion, owner_);
+    }
+
+    function _migrate(uint64 previousVersion, address owner_) private reinitializer(VERSION) {
+        if (previousVersion == 2) {
+            assembly ("memory-safe") {
+                sstore(_relayer.slot, shr(8, sload(_relayer.slot)))
+            }
+            __Ownable_init(owner_);
```
seunlanlege: With the release staying at 3, this collapses back to the single path: require the previous version to be 2, shift the relayer slot, initialize ownership, `reinitializer(VERSION)`.

The version-3 branch, the ownerless guard and the `previousVersion` plumbing all go. There is no version-3 proxy to migrate from, with an owner or without one, and an implementation that cannot be installed on one is clearer than a guard explaining which flavour of 3 we accept.

evm/src/apps/intentsv2/SolverAccount.sol
```diff
@@ -75,6 +82,11 @@ contract SolverAccount is Account, ERC7821, IERC1271 {
         INTENT_GATEWAY_V2 = intentGatewayV2;
     }
 
+    /// @notice The gateway release this account is built for.
+    function version() external pure returns (uint64) {
```
seunlanlege, line 86: 3 here too, to match the gateway release this is built for.

Separately, this account is not upgradeable, so the number can only ever be wrong by hand. Worth considering whether the SDK should identify a supported account by its address, the one it already has configured and reads out of the delegation, instead of asking the code to describe itself.

seunlanlege (commented):
One more: the release number. Nothing at version 3 has been deployed, so this release is still 3 and the whole migration should stay a single version-2 path. Knock-on in #1287: `SUPPORTED_INTENTS_VERSION` becomes `3n`, and the mapping that reads 3 as the v2 FillOptions shape has to go with it.

seunlanlege (approved):
.

seunlanlege (approved):
Image

Wizdave97 (approved):

Wizdave97:
@royvardhan  I'll merge this, your other PRs should be retargeted to main where appropriate after a rebase.

PR #1286, opened 2026-09-18
[indexer]: read the store, not a per-worker set, to filter solver transfers

Wizdave97:
## What

`applyTokenTransfer` and `applyVaultShareTransfer` decided whether a `Transfer` involved a tracked solver by consulting `trackedCache`, a module-level `Set` loaded once per process and mutated only by `seedPendingSolvers`. That set is gone. The filter keeps the same shape — one set of tracked solvers per chain — but it now lives in SubQuery's cross-worker `cache`, with the per-solver store read behind it as the authority.

## Why a module-level set could not be correct

SubQuery runs the mapping handlers in a worker thread per `--workers`, each thread a separate module registry with its own copy of every module-level binding, and `worker-block-dispatcher` rotates batches across all of them via a least-loaded `getNextWorkerIndex()`. Every `SUBQL_WORKERS` default in the package is above one — 16 in `scripts/templates/partials/docker-command.hbs`, 6 in `docker/docker-compose.local.yml` and `docker-compose.nexus-ci.yml`, 2 in `docker-compose.solver-ci.yml`.

So the worker that ran the handler making a solver `TRACKED` was the only one whose set learned about it. Every other worker held a set loaded before that solver existed and discarded its supported-token `Transfer`s at the first filter — no store read, no RPC, no log line.

**The harm over-reported.** A solver's outgoing `Transfer`s are what a fill is made of, so the dropped events were spends: the published `SolverInventory.balance` stayed at its pre-spend value and the orderbook quoted depth that was no longer there. Nothing repaired it short of the 24-hour `RECONCILE_INTERVAL_SECS` sweep or a process restart, and it hit freshly discovered solvers hardest — with 16 workers, roughly 15/16 of a new solver's transfers were dropped for the life of the process.

## Why `cache` and not a keyed store read

`cache` is SubQuery's cross-worker cache. The object lives on the main thread (`InMemoryCacheService`); each worker's `WorkerInMemoryCacheService` proxies `get`/`set` to it over the same host channel `store` already uses, and `WorkerCoreModule` installs that override, so it is live under `@subql/node-ethereum`'s `WorkerFetchModule`. A `cache.set` from the worker that seeds a solver is visible to all of them — the property module state lacked.

A store read is correct but not cheap enough to be the filter. `SolverInventory` and `SolverVaultShares` rows are written only by `applyReading`, so they exist for exactly the solvers this node tracks — but `cacheModel.get` populates its LFU only `if (record)`:

```js
record = (await this.model.findOne({ where: { id } }))?.toJSON();
if (record) {
    // getCache only keep records from db
    this.getCache.set(id, record);
}
```

A miss is never cached, and a miss is what every non-solver address is, so filtering on the store means a fresh Postgres `findOne` per address per `Transfer` on tokens whose transfers are overwhelmingly not solvers'. From a worker both `cache.get` and `store.get` are the same single host round-trip; the difference is entirely what the main thread does with it.

## Staleness, and what bounds it

- **`seedPendingSolvers` publishes each newly seeded solver to the cache**, right after the `TrackedSolver` row is saved and before any worker can see that solver's next `Transfer`. This covers ordinary discovery.
- **`advanceHead` rebuilds the set from the store**, past the `HEAD_INTERVAL_SECS` throttle it already applies. This covers the one race left: two workers seeding at once, where the loser's store read predates the winner's write. The rebuild unions the store's rows with whatever the cache already held, so a lost update repairs itself rather than persisting.

The rebuild rides the head throttle rather than carrying a bound of its own. The throttle is in seconds of block time, so the staleness bound is the same on every chain — a block count would have meant ~2 minutes on Arbitrum and ~100 on Ethereum from one constant, and the slow end sits well outside the 120s bound the orderbook works to. It is also gated on the shared `SolverInventoryHead` row, so a rebuild costs one `getByFields` per chain per interval, not one per worker.

The transfer handlers carry no clock of their own: their `timestamp` is deliberately a lazy RPC, resolved only once a tracked solver is involved, so reading it per `Transfer` to age an entry would cost more than the filter saves.

`InMemoryCacheService` has no rollback hook, unlike the store cache, so a reorg that un-discovers a solver leaves a stale member. That direction is harmless — the store read behind the hit finds no row. The dangerous direction, a member missing, is what the publish and the refresh bound cover. The cache is empty after a restart, which is correct rather than stale: first use per chain rebuilds from the store.

## Interface

`resetSolverInventoryCache` is removed; it was exported for tests and had no production caller. `src/types/global.d.ts` gains `cache` — this package declared its own globals and had left it out.

`etag` / `lastSkipReport` in `solverWatchlist.service.ts` and `deployed` in `utils/multicall.ts` are per-worker in the same way and are deliberately left alone: they cost a redundant fetch or `getCode` probe per worker, and neither can drop an event.

## Validation

`pnpm exec jest --runInBand src/services/__tests__/solverInventory.service.test.ts` — 23/23 pass. `tsc --noEmit` clean (the only output is a pre-existing `TS5101` deprecation on `baseUrl` in `tsconfig.json`). Prettier clean.

Three tests carry the change, each verified to fail against the code it guards:

- **`a solver another worker discovered still has its transfers applied`** is the regression guard. `jest.isolateModulesAsync` gives a second module registry over the same store and cache — a second worker — which handles a batch before the solver exists, then sees its outgoing transfer after the first worker discovered it. Fails against the original module-level `Set` with `wallet: 1000n` instead of `600n`, the 400 spend silently dropped. Also fails, identically, if the `cacheTrackedSolver` publish is removed from `seedPendingSolvers`, so that path is load-bearing rather than incidental.
- **`a cached set that lost a concurrent seed repairs itself on the next head advance`** pins the bound: a block inside the head throttle leaves the stale set in place and the spend is still dropped; the block that advances the head rebuilds the set and the solver is seen again. Fails if the `refreshTrackedSolvers` call is removed from `advanceHead`.
- **`transfers between untracked addresses cost no store read and no RPC`** keeps the cost assertion the original suite had, now with the cache as the thing answering: 2,000 transfers and 2,000 vault-share transfers between untracked addresses do `cache.get` only — no `store.get`, no paged `getByFields`, no RPC, no write.

The case for the second commit rests on that `findOne` being real traffic. I have not profiled production DB load; the argument is from the code path.


sdk/packages/indexer/src/services/solverInventory.service.ts
```diff
@@ -531,9 +512,7 @@ export async function applyTokenTransfer(input: TransferInput & { token: string;
 	const from = input.from.toLowerCase()
 	const to = input.to.toLowerCase()
 	if (from === to || input.value === 0n) return
-	const tracked = await trackedSolvers(input.chain)
-	const parties = [from, to].filter((address) => tracked.has(address))
-	if (parties.length === 0) return
+	const parties = [from, to]
```
royvardhan: The fix itself looks right to me. Three things:

**Cost of a miss.** A miss here isn't a cheap cached read. node-core's `cacheModel.get` only puts found records into `getCache` (`if (record) this.getCache.set(id, record)`), so a missing row always falls through to `model.findOne`. Nearly every supported-token Transfer misses on both sides, which makes it two Postgres round trips plus two worker-to-main hops per transfer, awaited in sequence. At the head that's probably fine. Where I'd expect it to show is a fresh deploy or reindex on Base, BSC or Polygon USDC, where it caps backfill speed. Have you checked blocks/s on Base before and after? If it's noticeable, the versioned cache you mention may be worth doing now instead of later: bump a version on a row that always exists from `seedPendingSolvers`, and have each worker reload its set when the version changes. A hit on an existing row is served from `getCache`, so the common path never touches Postgres. Not blocking the fix on it, since the correctness matters more.

**Zero address.** `applyVaultShareTransfer` filters `ZERO_ADDRESS` out of `parties` but this one doesn't any more (the set used to do it implicitly), so every mint and burn pays a `findOne` for `0x0`.

**Stale docs.** A few places still describe the in-memory set and now contradict the code:
- `docs/ai/flows/solver-inventory-discovery-genesis-transfers.md` step 5 ("joins the process's in-memory tracked set") and step 6 ("A side not in the in-memory tracked set is dropped with no store read and no RPC. The set is loaded by one paged field query...")
- `docs/ai/decisions/2026-09-14-solver-inventory-is-event-sourced-per-chain-with-a-discovery-queue.md` line 40 ("Chosen: transfers are filtered by an in-memory set of tracked solvers") and line 66
- the docstring in `handlers/events/solverInventory/tokenTransfer.event.handler.ts` ("those are dropped in memory")
- the comment in `scripts/templates/evm-chain.yaml.hbs` above the supported-token datasources ("dropped in memory, without a store read or an RPC")

The flow doc in particular should be updated in this PR since the behaviour it documents changed.

royvardhan (approved):

PR #1287, opened 2026-09-18
feat(sdk): declare and rank solver quotes for every fill leg

royvardhan:
Every bid carries one input take and one output budget per leg. The SDK validates the quotes before
signing and previews settlement with the same rule as the contract, so a signed quote cannot be
rejected as below the order's rate on chain. The SDK speaks one `fillOrder` shape:
`encodeFillOrder(order, options)` and `decodeFillOrder` handle the release-3 selector only, and
`assertGatewayRelease` refuses any gateway whose `version()` is not 3. Before signing, the gateway,
the configured `SolverAccount` implementation and the live delegation must all report release 3.

Automatic selection ranks bids by price normalized to the order's full input, so an oversized quote
cannot improve its rank while its signed amounts still bound funding and execution. A bid without a
take for every leg is not a candidate. Zero-output legs are excluded from valuation. Multi-leg
orders execute again; the earlier single-leg restriction is gone. Phantom aggregation gates every
bid on the release check and prices each leg at its normalized output.

Simplex sends a minimal quote so it keeps bidding once gateways are on release 3: the take is the
leg's full input, scaled by the pair's `maxOrderSize` cap fraction when the cap binds, and zero for
a skipped leg. Funding-limited takes and the rate gate for balance-limited legs wait for the
limit-order stack.

Removed: `FILL_ORDER_V1_ABI`, `FILL_ORDER_V2_ABI`, `HistoricalFillOptions`, `FillOptionsVersion`,
`getFillOptionsVersion`, `resetFillOptionsVersionCache`, `LEGACY_FILL_OPTIONS_IMPLEMENTATIONS`,
`CHAINS_WITHOUT_VALID_UNTIL`. Earlier SDK releases stay published for gateways below release 3.

Based on #1282. Merge before #1288.

Addresses #1259.


seunlanlege:
@royvardhan one more split, sorry. This touches Simplex, and @dharjeezy has a stack open that rewrites the same files: #1268 removes phantom bids, which deletes the `ContractInteractionService` method this PR extends and takes 288 lines out of `fx.ts`; #1271 rewrites `fx.ts` (+275 −901) and touches `CacheService`; #1275 touches both again. Whichever lands first, the other side redoes the work. And against a posted limit order the quote takes a different shape anyway: the order already is a take and an output per leg, so it does not come out of the fx curve and a cap fraction.

So:

- Make this PR SDK-only: `sdk/packages/sdk` plus its docs. That is what #1288 and the Simplex work both need.
- Move the Simplex quote work to its own PR stacked on #1275, written against the limit-order strategy once that is stable.
- We will not upgrade any gateway until that lands. Simplex on `main` speaks v2, and a release-3 gateway would reject every fill from it.

One thing to carry into the SDK-only version: `FillOptions.inputs` is required now, and Simplex's two `FillOptions` literals in `ContractInteractionService` do not set it. `inputs: []` on those two lines keeps Simplex type-clean, and it is correct today since the v2 encoder takes an empty array and no v3 gateway exists yet.

Review comments on the SDK half will follow once it is rebased.

seunlanlege:
And drop the legacy support while you are in here. The SDK release that carries quotes only has to talk to release-3 gateways; older SDK versions stay published for anything older. Same breaking-migration call we made for the contracts.

Concretely:

- One `fillOrder` shape. `encodeFillOrder(order, options)` with no version parameter, and `decodeFillOrder` decoding the v3 selector only. `FILL_ORDER_V1_ABI`, `FILL_ORDER_V2_ABI`, `HistoricalFillOptions` and the `intents-helpers` re-exports go. The indexer keeps its own pinned copies in `fill.helpers.ts` for reading history, so nothing there breaks.
- `getFillOptionsVersion` becomes a release check: `version()` must report 3, anything else throws. `LEGACY_FILL_OPTIONS_IMPLEMENTATIONS`, `CHAINS_WITHOUT_VALID_UNTIL`, `resolveImplementation` and the deprecated `resetFillOptionsVersionCache` go with it.
- `BidManager`: no implied take for a bid without `inputs`, and no historical branch in `rankingOutputs`. A bid that is not a v3 quote is not a candidate.
- Phantom aggregation: `inputTake` no longer defaults to the full order input, and the capability gate applies to every bid rather than to `version === 3`.
- The decision doc reduces to a few lines: `version()` must report 3.

With the SDK-only split this should take a good amount out of the diff.

Wizdave97:
@royvardhan  rebase on main, phantom bidding has been deleted

royvardhan:
@seunlanlege rebased on main and cut down to `sdk/packages/sdk` plus its docs. The simplex quote work is out of this PR; the one `FillOptions` literal left in simplex sets `inputs: []`.

Two questions before the legacy drop:

1. Simplex's `buildApprovalAndFillCalldata` still calls `getFillOptionsVersion` and the three-argument `encodeFillOrder`. Once those go, simplex stops compiling (it depends on the SDK via `workspace:*`). I would have simplex encode its v2 `fillOrder` through its own ABI copy in `config/abis/IntentGatewayV2.ts` via viem directly, about ten lines in that one file, and drop the v1 warning. That keeps simplex speaking v2. Is that touch acceptable here, or should it wait for the stacked simplex PR?
2. Simplex's `estimateFillOrder` goes through the SDK gas estimator, which will encode v3 and require release 3. Against a v2 gateway the estimate fails, so simplex on main cannot bid until the gateways are upgraded, and the filler tests in CI may go red if they hit live testnet gateways. Confirming that is the accepted cost of the breaking migration.

---

seunlanlege:
@royvardhan 

Taking the two together: a v2 shim does not get Simplex anywhere, because the estimator refuses a v2 gateway regardless. Simplex on `main` has to speak v3 to bid at all after this lands.

So for 1: make the shim v3, not v2. Have Simplex send a minimal quote: `inputs` equal to the order's inputs, scaled by the same cap fraction it already applies to the outputs, full input when uncapped. That is the contract's full fill at Simplex's price, about the same ten lines, and it belongs in this PR. The rich quote logic (funding-limited takes, the rate gate, previews) still waits for the limit-order stack; basic compatibility does not.

For 2: yes, accepted, and with that shim the window is short. Once this merges Simplex only bids on release-3 gateways. We upgrade testnet right after the merge so the filler tests go green again, and mainnet with the release. If the gap runs past a day, skip the live-testnet filler tests in between rather than leave `main` red.

This revises what I said above: with a minimal v3 quote in Simplex, the gateway upgrade no longer waits for the limit-order stack. Only the full Simplex quote PR does.

royvardhan:
Pushed both parts.

**Legacy drop** (`deb7df8`): one `fillOrder` shape. `encodeFillOrder(order, options)` has no version parameter, `decodeFillOrder` reads the release-3 selector only, and `assertGatewayRelease` replaces `getFillOptionsVersion` and throws unless `version()` reports 3. The v1/v2 ABIs, `HistoricalFillOptions`, the legacy implementation and chain lists, the cache reset and the `intents-helpers` re-exports are gone. `BidManager` no longer implies a take and ranks only quoted bids; every submission checks the release on the gateway, the live delegation and the configured account. Phantom aggregation requires a take per leg and gates every bid; it reads the delegate from the delegation check it already made, so no extra `eth_getCode` per bid. The decision note is now a few lines.

**Simplex minimal quote** (`f40c140`): `FXFiller` caches a take next to each output: the leg's full input, scaled by the pair's `maxOrderSize` cap fraction when the cap binds, zero for a skipped leg. `prepareBidUserOp` signs them into `inputs`; `buildApprovalAndFillCalldata` encodes the current shape only.

One thing to be aware of until the rate gate lands: a balance-limited leg that pays less than the pro-rata ask now signs the full take, so it will revert with `RateBelowOrder` on a release-3 gateway where it used to settle as a partial. The price gate still refuses legs whose curve is below the order's rate, so this only affects short fills caused by wallet balance.

Validation: SDK typecheck and 313 concurrent tests green (the five failures in `stateCalls.test.ts` are live-RPC reads that cannot reach the network from this sandbox); simplex typecheck and the fx/pairs suites green, with the curve-payout tests pinning the take for the uncapped, capped and balance-limited cases.

---

seunlanlege (approved):
Approving.

Reviewed the rebased SDK-only diff at `f40c140c9` including the legacy drop and the Simplex shim, and ran the touched suites from a fresh install: SDK 134 passed across the seven changed test files, Simplex `fx.curve-payout` 3 passed, and `tsc` reports no new errors against `main` in either package.

One non-blocking follow-up: the shim's take scales with the cap fraction but not with a wallet-limited output, so a balance-limited partial now fails the rate floor at simulation. `take = floor(budgetTake × finalOutputAmount / policyMaxOutput)` keeps those fills possible. Fine to land with the Simplex work on the limit-order stack.

royvardhan:
CI on `f40c140`: the `test` job is red only in the live-testnet filler step. Both cases in `src/tests/strategies/fx.testnet.test.ts` (BSC Chapel → Polygon Amoy) hit their 600 s timeout waiting for a bid; `pairs`, `fx.curve-payout` and `quorum-public-client` pass, and the SDK steps after it were skipped by the failure, not red.

This is the accepted consequence of the migration, not a defect in the diff. `version()` on the testnet deployments reverts with no data on both chains, for the gateway `0x6CF4…7cFA` and the SolverAccount `0x110C…6a17`, so they are pre-release-3 and the SDK's estimate refuses them. Simplex cannot bid, and the test waits for a fill that never comes.

No fix exists in this PR: it clears once testnet is upgraded to release 3, which is the plan agreed above. I have not skipped or quarantined the test. If the gap after merge runs past a day, the skip you mentioned is the call for `main`, not this branch.

---

PR #1288, opened 2026-09-18
fix(sdk): follow destination progress and attribute fills to our own operation

royvardhan:
The executor now takes an order's progress from its destination chain instead of only counting its own fills.

- Before starting and before each round of bids it reads `_filled` and the credited output per leg from `_partialFills`. Fills by other solvers count, and a resumed order continues from what is already credited.
- Once `_filled` is set the stream ends with `FILLED`, or with the new `CANCELLED` status when the order was finalized short of its output.
- `Bid.execute` reads fill events only from its own operation's logs in the bundle, bounded by the EntryPoint's `UserOperationEvent`, and only from the gateway. Another solver's fill of the same order in the same bundle is no longer reported as ours, and a reverted operation throws.

The submission journal, restart reconciliation and pending/rejected classification are dropped. With #1294 a stale bid fails selection on the gateway and two bids in one bundle settle against the order's real state.


seunlanlege:
#1287 is squashed into `main` as `a13765add` and GitHub has retargeted this PR to `main`, so its diff now shows #1287's original commits again. Rebase with `--onto` so only this PR's commits replay:

```bash
git fetch origin
git rebase --onto origin/main f40c140c9 roy/intents-submission-recovery
git push --force-with-lease
```

`f40c140c9` is #1287's final head. A plain `git rebase origin/main` would try to replay those commits over the squash and conflict.

seunlanlege:
@royvardhan heads up: #1294 changes the contract in a way that bears on this PR.

Two selection fixes, both on the gateway:

- `_select` now reverts `Filled` on a finalized order. A stale bid fails at `validateUserOp`, so the bundler's simulation drops it before it costs the solver gas or a nonce.
- The selection slot is keyed `keccak256(commitment, solver)` instead of by commitment alone. Two bids for one order in the same bundle no longer clobber each other; the first fills, the second takes what is left or reverts `Filled`.

Between them, the case the journal was built for is handled on chain: a second bid sent after a lost response either bounces at validation or settles against the order's real state. So I would cut this PR down to the parts that stand on their own: progress read from `_partialFills` and `_filled`, fill attribution to our own `UserOperationEvent` in the bundle, and the `CANCELLED` status. The journal, the restart reconciliation and the pending/rejected classification can go, along with their tests. That also removes most of the fixture work the rebase would otherwise need.

royvardhan:
Agreed, with #1294 the double-bid case settles on chain, so I've cut this down as you suggested. It's now one commit on main: progress read from `_filled` and `_partialFills`, fill attribution bounded by our own `UserOperationEvent`, and the `CANCELLED` status. The journal, restart reconciliation and pending/rejected classification are gone with their tests.

One thing I left out that wasn't on your list: the abort plumbing. On main the deadline watcher keeps polling until the order's deadline if a consumer stops early. Happy to send that as its own small PR if you want it.

The overview doc mentions stale bids failing selection on the gateway, so this should land after #1294.

seunlanlege (approved):
Approving.

Reviewed the reduced diff at `b2e13311c` and ran the five touched suites from a fresh install: 121 passed, no new `tsc` errors against `main`. The `_filled`-then-`_partialFills` read order is right, and the bundle-range walk matches the EntryPoint's event ordering.

Land it after #1294. Send the abort plumbing as its own PR, it is worth having.

PR #1292, opened 2026-09-18
[evm]: IntentGatewayV2 nits

seunlanlege:
`IntentGatewayV2.fillOrder` now runs the steps that same-chain and cross-chain fills share. It
validates the fill, claims the order in `_filled`, and delegatecalls the route's module. From the
`FillResult` the module returns, it runs the output calldata or reopens the order, emits
`OrderFilled` or `PartialFill`, and refunds unspent native value.

Each module keeps only its own route in `_fillOrder`. `IntrinsicModule` pays the legs and releases
the escrow here. `ExtrinsicModule` pays the legs and posts `RedeemEscrow` or `RedeemEscrowPartial`
to the source chain, taking the native dispatch fee out of the value it returns. Both take
`(Order, FillOptions, bytes32)` and return `FillResult`, which `_delegate` now hands back. They
replace `fillSameChain` and `fillCrossChain`.

## Behaviour

A completing cross-chain fill dispatches its redeem request before running the output calldata, so
the host's `PostRequestEvent` now precedes the output sweep's `DustCollected`. Same-chain fills keep
their order of effects and events.

`placeOrder` rejects an output token whose upper 12 bytes are set, as it already did for inputs.
Fills keep checking both tokens: the destination of a cross-chain order never sees `placeOrder`, and
the output sweep's `_isRepeatedToken` needs every token in one form.
`testFill_RejectsOutputTokenWithUpperBytesSet` fills such an order without placing it.

`SolverAccount` drops `version()`, which nothing reads on chain or off. It still refuses the current
and both historical `fillOrder` selectors on its plain ECDSA path.

The gateway's own ABI, events and errors are unchanged, and `fillOrder` keeps the selector
`0x68ddf058`. The modules' bytecode changes, so an upgrade deploys new modules with the new
implementation, as any module change does.

## Also

The leg loop body moved into `_fillLeg`. Returning the `FillResult` leaves the extrinsic module
without the stack for an inlined loop, because RLPReader's assembly stops the compiler spilling to
memory.

`_withdraw` takes an internal `Withdrawal` struct that names its `isRefund` and `finalize` flags,
instead of two positional booleans; `WithdrawalRequest` stays the wire format decoded in `onAccept`.
`onAccept` and `onGetResponse` take the relayer gate as an `onlyRelayer` modifier. Method docs
across the gateway, its modules and `IntentsBase` are short multiline blocks now, down from 382
lines to 112.

## Validation

357 tests pass across the gateway, multi-leg, same-chain, protocol-fee, rate-fill, modules,
reentrancy, oracle and `SolverAccount` suites. `forge fmt --check` is clean.

| Contract | Before | After | Headroom |
|---|---|---|---|
| `IntentGatewayV2` | 16,523 | 18,232 | 6,344 |
| `ExtrinsicModule` | 18,549 | 16,597 | 7,979 |
| `IntrinsicModule` | 8,091 | 6,113 | 18,463 |

Returning the `FillResult` costs about 2.1k gas on a single-leg same-chain fill and about 2.6k on a
cross-chain one, measured with `forge snapshot` before and after.


ddboy19912 (approved):

Wizdave97:
### Regression: token upper-byte sanitation moved to the wrong chain

`_validateLegs` loses the check that ran on **every** fill:

```solidity
// 4a6e1115^ evm/src/apps/intentsv2/IntentsBase.sol:497
// Every use of a token reads the address in its low 20 bytes. Anything above would let one
// token pass `_isRepeatedToken` as two.
if (uint256(inputToken) >> 160 != 0 || uint256(outputToken) >> 160 != 0) revert InvalidInput();
```

replaced by two checks in `placeOrder` (`IntentGatewayV2.sol:209-210`), justified as:

> `placeOrder` rejects an output token whose upper 12 bytes are set, as it already did for inputs. The fill path no longer checks either token: escrow is only reachable through `placeOrder`.

That holds for escrow, but the check had a second job. **`placeOrder` runs on the source chain; `fillOrder` runs on the destination.** The destination has no record of the order — the solver supplies the whole `Order` as calldata and the commitment is just `keccak256(abi.encode(order))`. So the struct is now unsanitized exactly where it is most caller-controlled.

The consumer is `_execute` (`IntentsBase.sol:715-745`): `_isRepeatedToken` compares raw `bytes32` while the sweep truncates with `address(uint160(...))`. Legs carrying `T` and `T | (1 << 255)` are not seen as repeats, so two sweep calls are built against the same snapshotted balance and `DustCollected(T, B)` fires twice for one `B`.

**Reachability** — three independent sources of unsanitized orders on an upgraded destination:

* `evm/tron/contracts/apps/IntentGatewayV2.sol` has no `>> 160` check anywhere.
* Pre-upgrade EVM gateways checked inputs only, never outputs — so the rollout window and already-escrowed orders qualify.
* A destination-side fill of a fabricated order with no source escrow costs only gas: self-`select` defeats solver selection and `output.beneficiary` can be the filler.

**Impact** is bounded — I could not turn it into theft. For a revert-on-failure ERC20, or for native, the second sweep fails and `CallDispatcher` reverts the whole fill (self-DoS). The residual case is a `false`-returning ERC20: `CallDispatcher.dispatch` checks only `success`, never the returned bool (`evm/src/utils/CallDispatcher.sol:59-60`), so the duplicate sweep silently no-ops while both events stand. `_sweepDust` pays governance-named amounts out of a pooled balance that also holds escrow, with no on-chain solvency check, so a sweep sized off the inflated event ledger is the path to real loss. That last hop is governance-dependent, which is why I'd call this low rather than medium.

**Test coverage was removed rather than relocated.** `IntentGatewayV2MultiLegTest.sol` drops `testFill_RejectsOutputTokenWithUpperBytesSet`, whose docstring named this exact invariant — *"on the same chain and on the destination of a cross-chain order, so `_isRepeatedToken` never sees one token in two forms"* — and whose cross-chain half deliberately skipped `_place()`:

```solidity
Order memory crossChain = _ladder(bytes("SOURCE_CHAIN"), host.host());
crossChain.output.assets[1].token = aliasedDai;
vm.expectRevert(IntentsBase.InvalidInput.selector);
_fill(solverA, crossChain, 1200 * 1e18, 990 * 1e18);
```

The replacement `testPlaceOrder_RejectsOutputTokenWithUpperBytesSet` only exercises `placeOrder`, which a destination-chain fill never touches.

**Suggested fix**, either one line:

* keep `if (uint256(order.output.assets[i].token) >> 160 != 0) revert InvalidInput();` in `_validateLegs`, or
* make `_isRepeatedToken` compare `address(uint160(uint256(token)))` instead of the raw `bytes32`.

and restore the cross-chain half of the deleted test.

seunlanlege:
@Wizdave97 re [your comment](https://github.com/polytope-labs/hyperbridge/pull/1292#issuecomment-5734933608): you're right. The destination never sees `placeOrder`, and `CallDispatcher` lets the duplicate sweep through for an ERC20 whose `transfer` returns `false`, so both `DustCollected` events stand.

Fixed in 17c4f9232:

- `_validateLegs` rejects an input or output token with any of its upper 12 bytes set again, your first option. The `placeOrder` check stays, so a bad order is refused at placement and again at fill.
- `testFill_RejectsOutputTokenWithUpperBytesSet` is back with its cross-chain half: a fill of an order that was never placed, carrying an aliased output token, reverts `InvalidInput`. With only the output check removed, the test fails because the fill goes through, so it guards this case. The same-chain half can't come back as written, since placing the aliased order now reverts first; `testPlaceOrder_RejectsOutputTokenWithUpperBytesSet` covers that.

I kept the check rather than comparing addresses in `_isRepeatedToken`, so an aliased token also never reaches the fill events or anything added later that reads a token.

Wizdave97 (approved):

Issue #982, opened 2026-06-19
[sdk/simplex/indexer]: off-chain support for cross-chain partial fills


seunlanlege:
## Summary

[#980](https://github.com/polytope-labs/hyperbridge/pull/980) adds **cross-chain partial fills** to `IntentGatewayV2` at the contract level (multiple solvers each fill a slice of a cross-chain order; `RedeemEscrowPartial` releases a proportional escrow slice without finalizing; cancels refund only the proven-unredeemed fraction). That PR is contract-only — its own "Follow-ups" section defers the off-chain work. This issue tracks everything else required to ship the feature end-to-end.

## Prerequisite: complete #981 first

**This work should land after [#981](https://github.com/polytope-labs/hyperbridge/issues/981) (make `IntentGatewayV2` a UUPS proxy upgradeable via cross-chain governance).** Reasons:

- Cross-chain partial fills are a substantial change to a **fund-custody contract deployed at one canonical address across 9 chains**. Without the proxy, rolling it out means redeploying and re-registering the gateway on every chain — which changes the canonical address and breaks cross-chain auth (`_authenticate` keys off the gateway address) and any in-flight orders. With the proxy, it ships as a single coordinated **governance upgrade** via the `intents-coprocessor`, preserving the address.
- The PR itself flags "cross-chain fund movement — independent/cloud review recommended before mainnet." The proxy gives a governance-driven **hotfix/rollback path** if an issue is found post-deployment, instead of another multi-chain redeploy.
- #980 adds storage (`_partialFills` use for cross-chain, plus a slot guard). Landing the proxy first fixes the canonical storage layout so these additions are appended cleanly under the append-only rule.

### SDK (`sdk/packages/sdk/src/protocols/intents/`)
- **`Bid.ts` / `BidManager.ts`** — construct partial cross-chain fills (fill a slice of the outputs), decode/allow partial-amount fill-options for cross-chain bids. Today `Bid.ts` parses `OrderFilled`/`PartialFill` only to classify status; it must also be able to *build* a partial cross-chain fill.
- **`OrderExecutor.ts`** — extend the cross-chain lifecycle (`AWAITING_BIDS → BIDS_RECEIVED → BID_SELECTED → FILLED`) to support repeated `PARTIAL_FILL` transitions like the same-chain path already does, until the order is fully filled, cancelled, or expired.
[...]

PR #1295, opened 2026-09-21
fix(sdk): prove per-leg fill progress when cancelling cross-chain orders

royvardhan:
A source-initiated cross-chain cancel has the gateway dispatch a GET for one `_partialFills[commitment][index]` key per order leg, with the context `abi.encode(commitment, user, inputs, totalRequired)`. The SDK was still quoting and proving the single `_filled` slot, so its proof did not cover the keys the gateway asks for.

- `OrderCanceller` quotes the GET with the per-leg keys and context, and `fetchDestinationProof` proves every leg's slot. The slots come from `partialFillSlot(commitment, index)`, which mirrors the gateway's `_calculatePartialFillSlotHash`, so neither path calls `calculateCommitmentSlotHash`.
- `OrderStatusChecker.isOrderFilled` reads the `_filled` getter. It is true once a completing fill or a destination-side cancel has finalized the order.
- `getFillProgress(order)`, also on `IntentGateway`, returns the credited output per leg from `_partialFills`.
- `RequestKind` gains `Execute = 5` and `RedeemEscrowPartial = 6`, matching the gateway and the coprocessor pallet.

This covers the three SDK callers that #1294 lists as needing a follow-up.

Part of #982.

**Note:** the generated ABI files in sdk, simplex and indexer still need regenerating once #1294 merges. This targets release 3 gateways only, so it should ship with or after that upgrade.


royvardhan:
This PR covers the SDK half of #982 only. The Simplex items are still open and I'm deliberately leaving them out of here.

What's left on the Simplex side:

1. **Fee-aware profitability for cross-chain partials.** #1271 already lifts the same-chain restriction in `fx.ts`, so cross-chain orders can be sliced once it lands. What it doesn't do is price them. Every slice pays its own relayer fee to carry `RedeemEscrowPartial` back to the source, and only the completing solver collects `order.fees`. The partial profit figure nets neither, so right now it is a code comment asking the operator's margin to absorb it. That needs to be an actual check.
2. **In-flight redeem tolerance.** A cross-chain slice isn't final until its redeem lands on the source. A cancel refunds only the unredeemed fraction, so an in-flight `RedeemEscrowPartial` stays covered, but settlement tracking shouldn't book the slice as settled before then.
3. **Follow-on fills.** The event monitor already watches `PartialFill`. I still need to confirm that a cross-chain partial by another solver gets the order re-evaluated for the remainder.

Why a separate PR, after the limit-order stack lands:

- All three live in the pricing and sizing path of `fx.ts`, which #1271 and #1275 are rewriting. Doing it now means either conflicting with that stack or building on code that is about to change.
- Item 1 only makes sense on top of #1271, since that is where cross-chain partials become reachable at all.
- This PR is a fix for a proof mismatch in the SDK and is small enough to review on its own. Mixing solver economics into it would slow down both.

I'll pick the Simplex work up once #1271 merges and open it against #982.

seunlanlege (approved):
Approving.

Checked the slot derivation, keys and context against the gateway byte for byte, and the enum values against the pallet. Ran the related suites from a fresh install at `935383dd3`: 29 passed, no new `tsc` errors against `main`.

ABI regeneration and one `sdk` version bump covering #1287, #1288 and this can follow together.

PR #1297, opened 2026-09-21
[evm, pallet-intents-coprocessor, sdk, indexer]: give every bid its own nonce key, let a filler hold several bids on one order, remove phantom orders

Wizdave97:
A solver quoting from several resting prices bids once per price on the same order, and those bids have to execute independently. Two things prevented that.

**One nonce key per order (`SolverAccount`).** A selected bid's nonce key was `keccak256(commitment ‖ sessionKey)`, so a solver's bids on one order were sequences of a single key. The EntryPoint runs a key's sequences strictly in order, so one bid that was never selected blocked every bid signed after it. The key is now

```
uint192(keccak256(commitment ‖ sessionKey ‖ keccak256(op.callData)))
```

Each bid is sequence 0 of its own key and executes on its own, in any order. The key is still derived entirely from what the op carries, and the same calldata still executes once. Hashing the calldata costs about 1.1–1.7k gas per validation on a realistic bid (1.9–3.5 KB of calldata). Decoding the `fillOrder` inside it would cost 4.4k–15k.

**One bid per filler (`pallet-intents-coprocessor`).** Bids were keyed by `(commitment, filler)`, and `place_bid` replaced the previous one. The RPC's pool cache and `RpcBidInfo`'s ordering did the same. Bids now live in a new `OrderBids` map keyed by `(commitment, filler, bid)`:
- `place_bid(commitment, bid, user_op)` and `retract_bid(commitment, bid)`. `bid` is an `H256` the filler chooses; by convention it is `keccak256(callData)` (`CryptoUtils.bidId`), the same hash the nonce key takes. Placing again under an identifier the filler already holds replaces that bid alone, and each bid holds its own deposit.
- A filler holds at most `MaxBidsPerFiller` bids on one order (5 on both runtimes). A new identifier past that is refused with `TooManyBids`, so one account can't crowd other fillers' bids out of what the RPC serves.
- `BidPlaced` and `BidRetracted` carry `bid`. The offchain key is `intents::bid:: ++ commitment ++ filler ++ bid`, and `intents_getBidsForOrder` returns every bid with its identifier.
- `OrderBids` is a new storage item rather than a migrated `Bids`, so the new layout lives under its own prefix and nothing is migrated. A bid standing in the old `Bids` map at the upgrade is neither served nor retractable through the new calls.

**Phantom orders removed from the pallet.** Simplex no longer bids on phantom orders. The pallet stops generating them and indexing them offchain: the `on_initialize`/`on_finalize` hook, its storage (`CurrentPhantomOrder`, `LastPhantomGeneration`, `PhantomBidWindow`, `PhantomOrderConfig`, `PhantomOrderInterval`, `PhantomChains`) and the governance calls that configured it are gone. So are its events and errors, the phantom bid rules in `place_bid`, the `PhantomOrderBidWindowBlocks` constant, the phantom order types, the v1 → v2 config migration, and their benchmarks, weights and tests. The removed storage is left in place, not cleared. The SDK's `fetchPhantomOrder` and `pollPhantomOrders` read what the pallet no longer writes; nothing in the indexer or simplex calls them.

**Runtime versions.** nexus goes to `spec_version` 8_600 and gargantua to 8_500.

**SDK and indexer**
- `CryptoUtils.bidNonceKey(commitment, sessionKey, callData)` derives the new key. `BidManager` and the phantom bid verifier check it.
- `IntentsCoprocessor.submitBid(commitment, userOp, bid)` and `retractBid(commitment, bid)` take the identifier. `FillerBid`, `BidStorageEntry` and the helpers' `RpcBidInfo` report it. The phantom helpers keep one bid per filler, under the zero identifier.
- The indexer's `FillerBid` gains `bid`, and bid data is matched on commitment and identifier. `BidPlaced` events from before the upgrade (three fields, with the deposit third) are skipped.
- `RpcBidInfo` is declared once, in `types`. In the RPC, its equality and ordering now both compare `(commitment, filler, bid)`.

**Simplex on main** reads each bid's nonce from its own key when it signs the bid, files the bid under `keccak256(callData)`, and records that identifier on the bid's row. Retraction names the recorded identifiers instead of reading the pallet's storage. A bid whose `place_bid` is still in the pool isn't in storage yet, so reading storage would have marked it retracted and abandoned its deposit once it landed.

The limit-order stack (#1270 → #1271 → #1275 → #1291) builds on this to send one bid per matching limit order, best offer first. Merge this first.

**Needs updating outside this repo:** the HyperFX orderbook's `BAD_NONCE_BINDING` check in `verify.rs` must derive the key the same way, since limit-order postings now bind their calldata too.

**Validation**
- `forge test --match-contract SolverAccount`: 29 pass. That includes two bids on one order validating in either order, a key derived from other calldata being refused, and the cross-language key vector. The full foundry run passes everything except the fork suites that need `MAINNET_FORK_URL`.
- `cargo test -p pallet-intents-coprocessor -p pallet-intents-rpc`: 41 + 8 pass, and 56 with `runtime-benchmarks` (the benchmark tests included). The `place_bid` benchmark now measures a filler one bid short of the cap; the runtime weight files still need regenerating. `cargo check` passes with `runtime-benchmarks`, for both runtimes, and for the simtests. The RPC simtest wasn't run, since it needs a live node.
- SDK: 112 tests pass across bid selection, the coprocessor, phantom aggregation and the key vector. Indexer extrinsic helpers: 9 pass. Simplex offline suites (341 tests) and lint pass.


seunlanlege:
Review of the branch at `2d780a6`, covering both the original sequenced-bids commits and the nonce-key redesign.

The redesign is the right shape. Giving each bid its own nonce key fixes the ordering problem at the root instead of teaching bid selection about sequences, and it closed three other things along the way: the `Twox64Concat` hasher on a caller-chosen key, the u64-through-JSON precision hop, and the possibility of the registered identifier disagreeing with the op's nonce. The cross-language vector test pinning `bidNonceKey` to the Solidity derivation is a good addition.

What follows is what I think is still open.

## Solvers on the previous SolverAccount cannot bid during the rollout

`SolverAccount.validateUserOp` now requires `keccak256(commitment ‖ sessionKey ‖ keccak256(op.callData))`, and `CryptoUtils.bidNonceKey` computes only that derivation.

`delegatedSolverAccount` in `sdk/packages/sdk/src/protocols/intents/phantom-aggregation.ts:885` takes a *list* of accepted SolverAccounts, with the comment "so a SolverAccount redeployment does not unseat solvers still delegated to the previous one". The system is built to run several implementations at once. This change has no matching branch.

So during the window where both are delegated: a solver still on the old contract has its op built with the new key by `BidManager.prepareSubmitBid`, its `validateUserOp` compares against the old `keccak256(commitment ‖ sessionKey)`, and returns `SIG_VALIDATION_FAILED`. Every bid it places is unexecutable. In the other direction `isVerifiedSolverBid` rejects a legitimate old-derivation bid as "nonce does not bind order and session key".

`SolverAccount` has no `version()`, so nothing can detect which derivation an account expects. As written this is a flag day across every solver and chain. Worth deciding explicitly whether that is acceptable, or whether `bidNonceKey` should take the delegate address and pick.

## A bid still in the pool gets marked retracted and its deposit abandoned

`sdk/packages/simplex/src/core/filler.ts:1099`

`retractOurBids` reads the identifiers back from chain and returns `BidNotFound` when it finds none. `enqueueRetraction` treats `BidNotFound` as terminal and calls `markRetracted(commitment, null)`.

But `BidStore.expiredUnretracted` deliberately sweeps bids whose `place_bid` is still pending, because — quoting its own doc — "a pooled extrinsic that later lands reserves a deposit exactly like a confirmed one". For such a bid the storage read at the current best head finds nothing, so it is permanently marked retracted. When the pooled `place_bid` lands it reserves the deposit, and `unretractedReclaimable()` and `expiredUnretracted()` never surface that row again.

The old code submitted the retraction on a later nonce from the same account, so the chain ordered it after the pooled `place_bid` and it always reclaimed.

The clean fix is the one below.

## Store the bid identifier instead of reading it back

`sdk/packages/simplex/src/core/filler.ts:1094`

`prepareBidUserOp` already returns `bid`, and `fx.ts:1100` drops it after `submitBid`. It is `keccak256(callData)` — a pure function of data simplex holds at placement time, so there is nothing to look up.

`BidStore`'s doc calls it "the money-critical store: a successful bid locks a deposit that is only reclaimable by retracting it, and the retraction sweep finds what to reclaim by querying here". Adding `bid` to `BidInsert`/`StoredBid` beside `commitment` keeps that true, makes retraction a local lookup, and removes the race above.

It would also drop the per-retraction chain read and let several identifiers go out in one `utility.batch` rather than one awaited extrinsic each.

## One filler can crowd every other bid out of the RPC

`modules/pallets/intents-coprocessor/rpc/src/lib.rs:259`

`MAX_ON_CHAIN_BIDS = 30` used to bound fillers per order. It now bounds `(filler, bid)` pairs.

The identifier is an arbitrary `H256` the filler picks, enforced nowhere, and `place_bid` only requires `!user_op.is_empty()`. There is no per-`(commitment, filler)` cap in the pallet. One funded account can send 30 `place_bid(commitment, <random H256>, vec![0u8])` calls, or 30 copies of one real op under 30 identifiers.

`storage_keys` returns keys in trie order, so `keys.take(30)` fills from that filler and every honest bid is dropped from the response `BidManager` and `OrderExecutor` read. The cost is 30 transaction fees — every deposit comes back through `retract_bid`.

A per-filler bid cap in the pallet is the fix that actually holds; raising the RPC constant only moves the number.

## Historical BidPlaced events decode the deposit as the bid identifier

`sdk/packages/indexer/src/handlers/events/substrateChains/handleBidPlaced.handler.ts:32`

Changing `FillerBid` forces SubQuery to re-index from the configured start block, which predates the runtime upgrade. Events in those blocks decode against their own block's metadata, where `BidPlaced` is `{ filler, commitment, deposit }` — three fields. So `bidIdData` is the deposit, and `.toHex()` on that `u128` is stored as the identifier for every historical bid.

Those bids lose their payload too. `extractUserOpFromExtrinsic` destructures the old `placeBid(commitment, user_op)` args as `[commitmentArg, bidArg, userOpArg]`, so `bidArg.toHex()` is the user op and `userOpArg` is undefined. The RPC fallback then cannot match either.

A guard on `data.length` would let the handler skip or special-case pre-upgrade events.

## Smaller things

**`Ord` and `Eq` disagree on `RpcBidInfo`** — `modules/pallets/intents-coprocessor/rpc/src/lib.rs:61`. The derived `Eq` covers all four fields; the hand-written `Ord` compares `(filler, bid)`. Two entries with the same filler and identifier but different `user_op` are `Ordering::Equal` yet `!=`, which breaks the contract `BTreeSet` relies on. Which `user_op` survives the merge in `get_bids_for_order` currently depends on `BTreeSet::insert` keeping the existing element — correct behaviour, but undocumented and the opposite of what `Eq` says.

**The migration comment argues from the hazard** — `modules/pallets/intents-coprocessor/src/migrations.rs:237`. "The old keys are a prefix of the new ones under the same storage prefix, so every old entry is taken before any new one is written." The prefix relationship is what makes a lazy drain unsafe, not what makes it safe: the new key sorts immediately after the old one in the same range. The migration is correct because `.collect()` fully drains first. Someone refactoring to `for (c, f, d) in Bids::<T>::drain() { ... }` — which the comment reads as endorsed — would re-drain each key it just wrote. Worth saying that `.collect()` is load-bearing. The `Vec` is also unbounded.

**`RpcBidInfo` is declared twice** — `sdk/packages/sdk/src/protocols/intents/phantom-aggregation.ts:577` and `sdk/packages/sdk/src/chains/intentsCoprocessor.ts:280`. The same field was hand-edited in both copies twice in this PR. One exported declaration in `types/index.ts`, beside `FillerBid` and `BidStorageEntry`, avoids the drift.

**`getBidStorageEntries` is unguarded** — `sdk/packages/sdk/src/chains/intentsCoprocessor.ts:1083`. `storageKey.args[2].toHex()` throws on a pre-upgrade two-key `Bids` map, out of a public SDK method. The sibling `getBidsViaStorage` wraps each entry in `try`/`catch`.

**Migrated bids leave orphaned offchain blobs.** The row moves to `(commitment, filler, MIGRATED_BID)`, so `retract_bid` clears `intents::bid:: ++ commitment ++ filler ++ 0x00…`, a key never written. The original payload stays in every node's offchain database with no key derivable from post-migration state.

**No version bump.** `submitBid`, `retractBid`, `bidNonceKey`, `BidNonceKeyFn`, `FillerBid` and `BidStorageEntry` all change shape, and `sdk/packages/sdk/package.json` is still at 2.8.13. Simplex consumes all of them and is not bumped either.

**One wording nit.** `evm/docs/ai/changelog/2026-09-21-every-bid-has-its-own-nonce-key.md` says "The same calldata still executes once: the EntryPoint consumes the key's first sequence." `_filled[commitment]` is what provides that, not the EntryPoint — and on a partial fill `_filled` is deleted and the order stays open, so the same calldata re-running is intended.

seunlanlege (approved):

PR #1298, opened 2026-09-21
[simplex-desktop]: serve bundled UI from desktop shell

ddboy19912:
## Summary

Fixes Simplex Desktop displaying the “UI not built” fallback after attaching to an already-running solver.

The desktop shell now serves HTML, JavaScript, styles, icons, and client-side routes from the UI bundled with the installed application. Only `/api/*` and `/health` requests are proxied to the solver.

This prevents stale or locally built solver processes from replacing the installed desktop UI. Existing desktop-versus-solver version compatibility checks remain active.

## Testing

- Desktop unit and integration suite: 148 passed, 1 platform-specific skip
- TypeScript typecheck
- Biome lint
- Electron regression test with a same-version solver missing its UI build
- Unsigned macOS arm64 package build
- Packaged application smoke test


seunlanlege (approved):

seunlanlege (approved):
Splitting the protocol handler so the shell owns the renderer and only `/api/*` and `/health` reach the solver is the right fix. Static serving looks correctly guarded.

seunlanlege (approved):

PR #1300, opened 2026-09-21
[runtime]: Verify super root dispute games on OP Stack

dharjeezy:
OP Sepolia moved its dispute games from game type 8 to type 9, whose root claim is a keccak hash over the super output preimage carried in `extraData` rather than a bare output root, and whose `claimData` sits at slot 1 instead of slot 2. This adds a `SuperFaultDisputeGame` kind that derives the claim from that preimage and requires this chain's output root to be the entry it commits to, and teaches the host to resolve the L2 block from the timestamp the preimage pins. Gargantua has not tracked OP Sepolia since 17 September because every type 9 proof was rejected, and this restores it once type 9 is added to that chain's dispute game config.

Recovery also needs a governance call adding game type 9 for OP Sepolia pointing at `0x19AF533Cc2A2A55786DCB8672aA5717e64213208`.


Wizdave97 (commented):
We need to update the dispute game blacklist task in the fisherman

Wizdave97 (commented):
Reviewed the verifier for soundness and panic safety. **The verification logic is correct**, but the collator fisherman will blacklist every type 9 game, which likely keeps OP Sepolia stalled even after this merges and the governance call goes through.

### Blocking: the fisherman blacklists every honest type 9 game

`fish_opstack` (`tesseract/messaging/fisherman/src/opstack.rs`, not touched by this PR) looks at every `DisputeGameCreated` event without checking the game type. For each game it:

1. reads `l2SequenceNumber()` and treats it as an L2 block number. For a super game this is the preimage timestamp (`0x6ab0ed7c`, about 1.79 billion for game `0x69C0…14DB`), and OP Sepolia has no block at that height;
2. gets "block missing" from every L2 RPC, so the quorum decides `MissingFromQuorum`;
3. calls `blacklist_dispute_game`, which is permanent.

Even if the block existed, it would compare a super root with a plain output root, see a mismatch, and blacklist anyway. The fisherman polls every 30s, while the relayer has to wait for Sepolia finality on Hyperbridge. So on any collator whose config has OP Sepolia with a `dispute_game_factory`, every honest type 9 game is blacklisted before it can be relayed, and the verifier then rejects it with `DisputeGameBlacklisted`.

This fails closed, so it is not a safety hole. It does cancel out what this PR is for, and it leaves super games with no working fraud check. The fix: when the game type is a super kind, the fisherman should parse the preimage with `parse_super_output`, resolve the timestamp to a block (as `block_at_timestamp` does), and compare only this chain's entry against the output root the quorum computes.

### Verified against the live game `0x69C0Af72663bFDaAd091a6D90bE4De658c0c14DB` (Sepolia)

| What the PR assumes | Live result |
|---|---|
| Root claim is `keccak256(extraData)` | `keccak256(extraData)` = `0xcb80…5359` = `rootClaim()` ✅ |
| `claimData` length is at slot 1 | slot 1 = `1` = `claimDataLen()`, slot 2 = `0` ✅ |
| `claimData[0].claim` holds the root claim | claim word at `keccak256(1) + 3` = `0xcb80…5359` ✅ |
| Timestamp comes from the preimage | `l2SequenceNumber()` = `0x6ab0ed7c`, the preimage timestamp ✅ |

The chain from the header to the game holds: the header hash goes into the output root, which must be this chain's entry in the preimage (the chain id comes from the trusted consensus state). `keccak256(preimage)` is the root claim, which goes into the game UUID proven in the factory. Then `gameImpls[9]` and `claimData.length == 1` are checked. A wrong kind set by governance fails closed, because the UUID no longer matches.

### Panic safety

Nothing on the verifier path panics on attacker-controlled input. I checked by reading the code and by fuzzing about 1M inputs in debug mode (overflow checks on). The fuzzing covered random SCALE proofs, structured proofs with a valid factory account proof that reach the super branch, and `claim_data_len` / `decode_address_from_storage_value` against real single-leaf tries holding values of 0–70 bytes. `parse_super_output` checks the length before every index and slice, `claim_data_len` keeps the check that stops values over 32 bytes reaching `from_big_endian`, and a non-EVM state machine returns an error. `block_at_timestamp` in the host can't panic either.

modules/ismp/clients/optimism/src/lib.rs
```diff
@@ -107,6 +121,50 @@ pub struct GameTypeConfig {
...
+
+	let mut timestamp = [0u8; 8];
+	timestamp.copy_from_slice(&encoded[1..HEADER_LEN]);
+
+	Ok(SuperOutput {
+		timestamp: u64::from_be_bytes(timestamp),
+		output_roots: entries
+			.chunks_exact(ENTRY_LEN)
+			.map(|entry| (U256::from_big_endian(&entry[..32]), H256::from_slice(&entry[32..])))
+			.collect(),
```
Wizdave97: Nit (hardening): collecting into a `BTreeMap` means that if a chain id appears twice, the later entry silently wins. Honest super roots never contain duplicates (and are sorted by chain id), so rejecting duplicate or unsorted entries costs nothing and removes any ambiguity about which entry was checked.

Wizdave97:
gargantua runtime version and node version need to be bumped

Wizdave97 (commented):
Follow-up on `0bcc9c07d`, `92f631b52` and `5039cccbe`. Both earlier findings are fixed. There is one gap left in how the fisherman recognises a super game (inline), which I'd close before merge.

**Validation:** `cargo test --release -p op-verifier` passes 7/7, including the three new tests. `cargo check -p tesseract-fisherman` builds. The pinned Rust version, 1.91, supports `is_sorted_by`.

- **`0bcc9c07d`:** correct. A strict ascending check rules out duplicate and out-of-order chain ids in one step. The live one-entry preimage still parses, and a super output with entries out of order is now rejected rather than collapsed, which is the safe direction.
- **`92f631b52`:** fixes the blocking issue. For a super game, each L2 RPC finds the block for the preimage timestamp itself, so a single bad RPC gets outvoted. The quorum compares against this chain's entry in the preimage, which is the same entry the verifier checks.
- **`5039cccbe`:** fine.

The optional check that the header timestamp equals the preimage timestamp wasn't added. It isn't needed for safety.

tesseract/messaging/fisherman/src/opstack.rs
```diff
@@ -161,6 +165,66 @@ async fn scan_target(
...
+	let proxy_addr = Address::from_slice(&proxy.0);
+	let contract = FaultDisputeGame::new(proxy_addr, &*cfg.l1_provider);
+	let extra_data = contract
+		.extraData()
+		.block(BlockId::number(at_block))
+		.call()
+		.await
+		.map_err(|e| anyhow!("fish_opstack: proxy {proxy:?}.extraData() failed: {e:?}"))?;
+
+	let Ok(super_output) = parse_super_output(&extra_data) else {
```
Wizdave97: **Gap: the fisherman and the verifier decide "super game" differently.** Here a game counts as a super game whenever its `extraData` parses as a preimage. The verifier decides by the game type configured by governance. If an output-root game ever had `extraData` that also parsed as a preimage, the fisherman could be fooled:

1. An attacker creates an output-root game with a fake root claim, and `extraData` that parses as a preimage whose entry for this chain is the honest output root.
2. The fisherman takes the super path, compares that honest entry, gets `Verified`, and doesn't blacklist the game.
3. The verifier handles it as an output-root game and accepts the fake header, because it only needs the root claim in the game's ID. It never checks the block number inside `extraData`.

I don't think anyone can do this today. It depends on what the game contracts accept, not on anything in our code: Cannon's `extraData` is fixed at 32 bytes and OP Succinct's at 36, and neither length can parse as a preimage. A preimage also starts with `0x01`, which would mean a block number of at least 2^248.

**Suggested fix:** only take the super path when `keccak256(extraData) == rootClaim`.
- A real super game always passes. I confirmed this on `0x69C0…14DB`.
- An output-root game can't pass: its claim is a hash of 128 bytes of output-root input, and a preimage is never that length, so passing would need a keccak collision.
- `root_claim` is already passed into `evaluate`, so no extra RPC call is needed.

tesseract/messaging/fisherman/src/opstack.rs
```diff
@@ -161,6 +165,66 @@ async fn scan_target(
...
+
+/// Reads the proxy's `extraData` and, when it decodes as a super output, returns the timestamp
+/// and the output root that super output carries for this chain.
+///
+/// Whether a game is a super game is decided by the preimage itself rather than by a list of
+/// game types, so a newly registered super type needs no change here. An output root game's
+/// `extraData` starts with the high byte of a `uint256` block number, which is zero, so it
+/// never passes the version check.
+///
+/// Returns `Ok(None)` for a game that is not a super game, and abstains (`Ok(None)` with a log)
```
Wizdave97: Nit: this comment, and the `abstaining` log below, say the fisherman abstains when the super output has no entry for this chain (and the same path is taken when the chain isn't EVM). In fact the function returns `Ok(None)`, which `evaluate` reads as "not a super game". It then falls through to `read_l2_block_number`, which reads the timestamp as a block number, finds no block at that height, and blacklists the game.

That outcome is harmless, since the verifier rejects such a game with `SuperOutputChainNotFound` anyway. Either change the wording to say it gets blacklisted, or return a separate abstain result so `evaluate` can return `Ok(true)` for it.

Wizdave97 (approved):

Issue #1231, opened 2026-09-09
[simplex]: Allow users to clearly verify their entered private key before proceeding with the setup process


olahfemi:
As it stands, a user can unknowingly input a wrong private key and would only find out close to the end of the setup process. A clear (maybe optional) button to "verify" (which would then show the user's address) might be valuable at this point.

Alternatively, we could ensure the user can only move forward if the private key is valid (however, this would make quick testing with random numbers more difficult).

PR #1307, opened 2026-09-22
[simplex-desktop]: add end-to-end operator push notifications

### LINKED ISSUE #1226: [simplex]: Push notifications
Add push notifications to simplex PWA that can alert operators once their liquidity is below a certain threshold

ddboy19912:
## Summary

Closes #1226 and #1231

Adds end-to-end operator notifications for the Simplex PWA and desktop app.

### Features

- Low-liquidity alerts based on aggregate USD-stable liquidity thresholds.
- Completed-swap notifications.
- Durable notification settings, VAPID keys, and browser subscriptions.
- PWA Web Push delivery through the service worker.
- Native desktop notifications over the private solver socket.
- Device-specific test notifications with delivery failure reporting.
- Subscription reconciliation when solver state or VAPID keys change.
- Duplicate swap-alert prevention.
- Notification click-through to relevant dashboard routes.
- Protection against alerts from partial balance snapshots.

### Verification

- Notification service tests passed.
- UI server tests passed.
- Full desktop test suite passed: 153 passed, 1 skipped.
- PWA TypeScript check and production build passed.
- Desktop TypeScript and Biome checks passed.
- Simplex package build passed.
- `git diff --check` passed.

No commit or push has been performed.


sdk/packages/simplex/ui/src/operator/OperatorBalances.tsx
```diff
@@ -285,13 +284,13 @@ function chainLabel(status: StatusOperator, chainId: number): string {
...
 /**
- * Available USDC and USDT across `assets`, or null when any contributing read failed. Refusing
+ * Available USD stablecoins across `assets`, or null when any contributing read failed. Refusing
  * to estimate is the point: a sum quietly missing a leg understates liquidity without saying so.
  */
 export function availableStablecoins(assets: AssetBalance[], snapshot: SnapshotStatus): number | null {
-	if (snapshot === "loading") return null
-	const stables = assets.filter((asset) => STABLES.has(asset.symbol.trim().toUpperCase()))
-	if (stables.length === 0) return snapshot === "fresh" ? 0 : null
+	if (snapshot !== "fresh") return null
```
seunlanlege: **A partial snapshot now blanks the stables column for every network, not just the failed one.**

This function is called two ways: per network row from `networkOption` (line 131) and for the
whole-fleet aggregate from `availableStablecoinLiquidity` (line 298). The new
`if (snapshot !== "fresh") return null` is right for the aggregate but wrong for the per-row call,
because `snapshot` is the *global* status.

Nine chains configured, the vault read on Base times out, so `BalanceSnapshot.status` is `"partial"`
while every USDC/USDT row on Arbitrum, Ethereum, BSC etc. still has a non-null `available`. All nine
rows in the network switcher now read `Unavailable`. Before this change only `"loading"` blanked the
value, and the existing `stables.some(asset => asset.available === null)` check already handled a
chain whose own reads failed.

Suggest keeping the per-row behaviour and moving the global gate into `availableStablecoinLiquidity`.

sdk/packages/simplex/ui/public/sw.js
```diff
@@ -74,3 +76,42 @@ self.addEventListener("fetch", (event) => {
...
+})
+
+self.addEventListener("notificationclick", (event) => {
+	event.notification.close()
+	const target = new URL(event.notification.data?.url ?? "./", self.registration.scope).href
+	event.waitUntil(
+		self.clients.matchAll({ type: "window", includeUncontrolled: true }).then(async (windows) => {
+			for (const client of windows) {
+				if (new URL(client.url).origin === self.location.origin) {
+					await client.navigate(target)
```
seunlanlege: **`client.navigate()` rejects for an uncontrolled client, so the click can open nothing at all.**

`matchAll({ includeUncontrolled: true })` can return a window this service worker does not control —
a tab loaded before this version took over, e.g. during the v6→v7 update before `clients.claim()`
resolves. Per spec `Client.navigate()` rejects with a `TypeError` for such a client. Nothing catches
it, so the async function rejects, `event.waitUntil` rejects, and neither `client.focus()` nor
`self.clients.openWindow(target)` runs — tapping the notification does nothing.

```js
try {
    const navigated = await client.navigate(target)
    return (navigated ?? client).focus()
} catch {
    return self.clients.openWindow(target)
}
```

Minor: the `new URL(client.url).origin === self.location.origin` test is redundant — `matchAll` is
already same-origin only.

sdk/packages/simplex-desktop/src/main.ts
```diff
@@ -302,8 +320,9 @@ async function stopAndQuit(): Promise<void> {
 	if (await stopSolver()) quitApp()
 }
 
-async function createWindow(): Promise<void> {
+async function createWindow(notificationPath?: string): Promise<void> {
 	if (mainWindow && !mainWindow.isDestroyed()) {
+		if (notificationPath) await mainWindow.loadURL(desktopNotificationUrl(notificationPath))
```
seunlanlege: **A rejected `loadURL` skips showing the window and pops an error dialog instead.**

The navigation is awaited before `restore()` / `show()` / `focus()`, and Electron rejects
`loadURL` with `ERR_ABORTED` when a navigation is superseded. Two alerts on screen and the operator
clicks both in quick succession: the second load aborts the first, `createWindow` throws before it
reaches the show/focus calls, and `safeShowWindow` surfaces
`reportActionError("Simplex could not open its window", error)`. The operator gets an error dialog
and the window stays hidden.

Show and focus first, then navigate, and treat a navigation failure as non-fatal.

Separately: this reloads the whole SPA on every notification click, discarding whatever the operator
had open. Routing through the renderer (the tab route already exists) would be less destructive.

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+function displayAmount(raw: string, decimals: number | null): string {
+	if (decimals === null) return raw
+	const negative = raw.startsWith("-")
+	const digits = negative ? raw.slice(1) : raw
+	const padded = digits.padStart(decimals + 1, "0")
+	const whole = decimals === 0 ? padded : padded.slice(0, -decimals) || "0"
+	const fraction = decimals === 0 ? "" : padded.slice(-decimals).replace(/0+$/, "").slice(0, 4)
```
seunlanlege: **`displayAmount` renders small amounts as all zeros.**

Trailing zeros are stripped from the whole fractional string *before* the truncation to four
characters, so any amount whose significant digits start past the fourth decimal collapses:

- `0.000012 WETH` → raw `"12000000000000"`, 18 decimals
- `padded.slice(-18)` → `"000012000000000000"`
- `.replace(/0+$/, "")` → `"000012"`
- `.slice(0, 4)` → `"0000"` → body reads **"0.0000 WETH → …"**

One wei renders the same way. The slice has to come before the strip — but viem's `formatUnits`
already does this correctly and is a direct dependency used in ten files here, including
`UiServer.ts`; `ui/src/lib/format.ts` has `formatAmount` for the display rounding.

seunlanlege (commented):
Reviewed the full diff plus the surrounding code. 14 comments inline — the first few are the ones I'd want addressed before merge; the tail is cleanup and docs.

**Holds up well.** `desktopNotificationUrl` correctly rejects `https://evil`, `//evil/` and `simplex://remote/` before anything reaches `loadURL`, and `simplex://local/orders` resolves through `handleSimplexProtocol`'s SPA fallback so desktop click-through lands on the right tab. Push payloads stay encrypted end-to-end, the SSE stream and receipt endpoint sit behind the existing host/CSRF/provenance guards, and the acknowledge POST carries `X-Simplex-UI`. Every write goes through `patchRuntimeState`, so notification state will not clobber `paused`.

**Not a comment, just a note:** neither `simplex` nor `simplex-desktop` gets a version bump here (both stay at `0.16.2`). That matches what #1268 and #1291 did, so I left it alone — flagging in case it was meant to move. `workspace-policy.test.ts` asserts the two match, so they would go together.

I did not run the suites locally (no `node_modules` in the review worktree). The new tests read sound, but nothing covers the wrong-error-message path or the partial-fill dedup case.

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,275 @@
...
+	private state!: NotificationRuntimeState
+	private ready: Promise<void>
+	private interval?: NodeJS.Timeout
+	private stopped = false
+	private readonly logger: Logger
+	private readonly notifiedSwapIds = new Set<string>()
+	private readonly onActivity = (event: ActivityEvent) => {
+		if (event.type === "filled" && this.state.settings.swaps) {
+			const id = event.orderId ?? `event-${event.id}`
+			if (this.notifiedSwapIds.has(id)) return
```
seunlanlege, line 95: **Deduping on `orderId` also suppresses later partial fills.**

The duplicate this targets is two rows for the *same transaction* — `ActivityRecorder` emits
`type: "filled"` from both `orderFilled` (recorder.ts:127) and `settle` when `ours`
(recorder.ts:263). But an order filled in parts produces genuinely distinct `filled` rows under the
same `orderId`, and the operator only ever gets an alert for the first one.

Keying on `orderId` + `txHash` (falling back to the event id when `txHash` is null) still collapses
the real duplicate — both paths carry the same fill transaction — while letting distinct fills
through.

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,275 @@
...
+		// known subtotal look lower than the operator's actual liquidity.
+		if (snapshot.status !== "fresh") return
+		const stableAssets = snapshot.chains.flatMap((chain) =>
+			chain.assets.filter((asset) => USD_STABLE_SYMBOLS.has(asset.symbol.trim().toUpperCase())),
+		)
+		if (stableAssets.length === 0 || stableAssets.some((asset) => asset.available === null)) return
+		const liquidity = stableAssets.reduce((total, asset) => total + (asset.available ?? 0), 0)
+		const low = liquidity < threshold
+		if (low === this.state.lowLiquidityActive) return
+		this.state.lowLiquidityActive = low
```
seunlanlege, line 230: **The armed flag is persisted before the alert is published, so a publish failure loses the alert
permanently.**

`lowLiquidityActive = true` is set and persisted, then `await this.publish(...)` runs. If publish
throws, the crossing is durably recorded as already-alerted and is never retried — liquidity has to
recover and drop again before the operator hears anything.

Publish can throw: the synchronous `this.emit("notification", …)` runs `UiServer`'s SSE writer,
which can throw on a response that closed between the `req.on("close")` tick and the write, and the
`persist()` of expired endpoints can reject. Reached from `updateSettings`, the same throw also
surfaces as a `400` on a PUT that did in fact save the settings.

Set the flag after a successful publish, or catch around it.

sdk/packages/simplex/src/services/server/UiServer.ts
```diff
@@ -871,6 +957,123 @@ export class UiServer {
...
+				const receiptId = native ? randomUUID() : undefined
+				const nativeReceipt = receiptId ? this.waitForNativeNotificationReceipt(receiptId) : undefined
+				const delivery = await this.notifications.test({ endpoint, native, push: !native, receiptId })
+				const nativeReceived = nativeReceipt && (await nativeReceipt) ? 1 : 0
+				if (delivery.pushSent + nativeReceived === 0) {
+					return sendJson(res, delivery.pushFailed > 0 ? 502 : 409, {
+						error:
+							delivery.pushFailed > 0
+								? "The notification service could not deliver to this device"
+								: "The desktop app did not confirm displaying the notification",
```
seunlanlege: **A browser push test that matched no subscription tells the operator the desktop app failed.**

For a browser test `native` is false, so `nativeClients` and `nativeReceived` are both 0. If the
endpoint is no longer in `state.subscriptions`, `publish` returns `{ pushSent: 0, pushFailed: 0 }`,
`delivery.pushFailed > 0` is false, and the 409 body reads *"The desktop app did not confirm
displaying the notification"* — in a browser, with no desktop app involved.

Reachable when the endpoint was dropped server-side after a push service returned 410 and the
operator clicks *Send test notification* without remounting the panel (which would re-register it).
The message needs to branch on `native` too — something like "this browser is no longer subscribed".

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,275 @@
...
+		private readonly activity: ActivitySource,
+		logger: Logger = getLogger("notifications"),
+	) {
+		super()
+		this.logger = logger
+		this.ready = this.initialize()
+		void this.ready.catch((error) => this.logger.error({ err: error }, "Notification service could not start"))
+	}
+
+	private async initialize(): Promise<void> {
```
seunlanlege, line 142: **A failed `initialize()` leaves the service permanently dead, and the GET surfaces it as an opaque
500.**

If `store.get()` rejects once at boot (locked SQLite file, transient I/O), `this.state` is never
assigned, the activity listener and the interval are never attached, and there is no retry.
`void this.ready.catch(...)` only silences the unhandled rejection — `this.ready` itself stays
rejected, so every later `status()` / `updateSettings()` / `subscribe()` rejects for the life of the
process.

At `UiServer.ts:959` the GET branch
(`return sendJson(res, 200, await this.notifications.status())`) sits *before* the `try`, so that
rejection escapes to the generic handler catch and the panel shows a bare 500 with no hint that
notifications are permanently unavailable.

sdk/packages/simplex/src/services/server/UiServer.ts
```diff
@@ -871,6 +957,123 @@ export class UiServer {
...
+					return sendJson(res, 400, {
+						error: "Select exactly one browser endpoint or the native desktop client",
+					})
+				}
+				const nativeClients = native ? this.notificationClients.size : 0
+				if (native && nativeClients === 0) {
+					return sendJson(res, 409, { error: "No connected notification device was found", nativeClients })
+				}
+				const receiptId = native ? randomUUID() : undefined
+				const nativeReceipt = receiptId ? this.waitForNativeNotificationReceipt(receiptId) : undefined
```
seunlanlege, line 1088: **The pending receipt and its timer leak when `notifications.test()` throws.**

`waitForNativeNotificationReceipt(receiptId)` registers the entry and a 5 s `setTimeout` *before*
`test()` is awaited. If `test()` rejects, the handler's catch returns 400 and nothing clears either.
The timer is not `unref()`'d, so it can also hold the event loop open past an otherwise-clean
shutdown.

A `finally` that removes the entry and clears the timer closes it.

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,275 @@
...
+		this.interval = undefined
+	}
+
+	private async evaluateLiquidity(): Promise<void> {
+		const threshold = this.state.settings.lowLiquidityThresholdUsd
+		if (threshold === null) return
+		const snapshot = this.balances.getSnapshot()
+		// A partial snapshot can omit an entire failed chain, which would make the
+		// known subtotal look lower than the operator's actual liquidity.
+		if (snapshot.status !== "fresh") return
```
seunlanlege: **This duplicates `availableStablecoinLiquidity`.**

Same fresh gate, same `USD_STABLE_SYMBOLS` filter, same refusal on a null `available`, same reduce —
see `ui/src/operator/OperatorBalances.tsx:290-302`. This PR already had to hand-edit the UI copy to
keep the two in step, which is exactly the maintenance cost worth avoiding: the alert threshold and
the number the operator reads on the Overview can silently diverge on the next change to either.

`OperatorBalances.tsx` already imports `@/config/asset-registry`, so the aggregate can live in
`src/` and be called from both sides.

Also: `asset.symbol.trim().toUpperCase()` duplicates `normalizeSymbol()` from that same module.

sdk/packages/simplex/src/services/server/UiServer.ts
```diff
@@ -871,6 +957,123 @@ export class UiServer {
...
+				if (!this.acknowledgeNativeNotification(body.receiptId)) {
+					return sendJson(res, 404, { error: "Notification receipt was not found" })
+				}
+				return sendJson(res, 204, {})
+			} catch (err) {
+				return sendJson(res, 400, { error: err instanceof Error ? err.message : "Invalid JSON body" })
+			}
+		}
+
+		if (path === "/api/notifications/stream") {
```
seunlanlege, line 1134: **This is a verbatim copy of the `/api/events` handler twelve lines below.**

`UiServer.ts:1066-1073` and `UiServer.ts:1193-1200` are byte-identical apart from the client set.
Any future fix to the SSE preamble — a keep-alive heartbeat, `X-Accel-Buffering`, an `error`
listener on the response so a write-after-end does not become an unhandled `error` event — now has
to be made twice, and will be made once.

A small `openSseStream(req, res, clients)` used by both removes it.

sdk/packages/simplex/src/services/server/UiServer.ts
```diff
@@ -922,10 +1125,24 @@ export class UiServer {
 			}
 			// Share tokens are named after their underlying (stataUSDC, ycNGN): the logo
 			// comes from the underlying's symbol, and the vault badge says it is a share.
-			const legOf = (symbol: string | null | undefined, amount: string | null | undefined, vault: boolean): LedgerLeg | null =>
-				symbol && amount ? { symbol, amount, decimals: null, icon: vault ? underlyingOf(symbol) : symbol, vault } : null
+			const legOf = (
```
seunlanlege, line 1193: **Unrelated prettier reflow makes this diff hard to review.**

Roughly fifteen hunks in this file are pure reformatting of pre-existing lines with nothing to do
with notifications — this `legOf` / `orderLeg` pair, the `assertSocketPathFits` message, the
tunnel-device handlers, `vaultSweepDto`, and others. It forces a reviewer to separate real changes
from reflow across a 308-line diff, and it will conflict with any concurrent PR touching this file.

Meanwhile `simplex-desktop/src/notification-client.ts:41-47` ships indented one level too deep —
`pnpm format` was run in `simplex` but not in `simplex-desktop`, whose `biome lint` does not check
formatting.

Either split the reflow into its own commit, or at least run the formatter in both packages.

sdk/packages/simplex/docs/ai/changelog/2026-09-21-operator-push-notifications.md
```diff
@@ -0,0 +1,17 @@
+# Operator push notifications
```
seunlanlege, line 1: **The balances behaviour change needs the existing decision note updated.**

Repo `CLAUDE.md`: *"Read relevant existing package notes before changing code to preserve documented
assumptions"* and *"Update an existing flow when its behavior changes."*

`docs/ai/decisions/2026-09-10-one-network-at-a-time-with-a-persistent-totals-strip.md` still says
*"a network whose read failed shows `Unavailable` in the stables column on its own row"* and that
*"the section-level 'Some balances are unavailable' notice covers the whole snapshot"* — the
per-network / whole-snapshot split that the new global `snapshot !== "fresh"` gate in
`availableStablecoins` removes.

`docs/ai/changelog/2026-09-10-balances-narrows-to-one-network-with-a-switcher.md` also still
describes the switcher as showing *"available USDC+USDT"*, while the move to `USD_STABLE_SYMBOLS`
adds DAI.

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,275 @@
...
+	NotificationRuntimeState,
+	NotificationSettings,
+	OperatorNotification,
+	StateStore,
+	StoredPushSubscription,
+} from "@/data/types"
+import type { BalanceSnapshot } from "@/services/BalanceProvider"
+import { getLogger, type Logger } from "@/services/Logger"
+
+const BALANCE_CHECK_INTERVAL_MS = 15_000
```
seunlanlege: **A 15 s poll over a snapshot that refreshes every 60 s.**

`BalanceProvider.intervalMs` defaults to `60_000` (BalanceProvider.ts:115), so three of every four
evaluations re-read identical data. The waste is cheap, but the independent timer is the wrong seam:
it can also evaluate a snapshot the provider is midway through replacing — precisely the partial-data
case the `status !== "fresh"` guard exists to avoid.

Evaluating off the balance refresh itself (a callback or an event from `BalanceProvider`) checks
exactly once per new snapshot and removes the second timer.

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,288 @@
...
+}
+
+interface ActivitySource {
+	on(event: "event", listener: (event: ActivityEvent) => void): unknown
+	off(event: "event", listener: (event: ActivityEvent) => void): unknown
+}
+
+interface BalanceSource {
+	getSnapshot(): BalanceSnapshot
+	on?(event: "snapshot", listener: (snapshot: BalanceSnapshot) => void): unknown
```
seunlanlege: **Making `on`/`off` optional here turns a missing emitter into silence rather than a compile error.**

Now that the 15 s interval is gone, snapshot events are the *only* thing that drives
`evaluateLiquidity` after init. Production is fine — `bin/simplex.ts:213` passes the real
`BalanceProvider`, which is an `EventEmitter` as of this commit. But `OperatorContext.balances` is
still typed `Pick<BalanceProvider, "getSnapshot">` (UiServer.ts:193), so nothing at the type level
says a balance source has to emit. An embedder — or a future test helper — that supplies a plain
`{ getSnapshot }` gets one evaluation at init and then no low-liquidity alerts at all, with no error
anywhere. The old poll degraded gracefully for exactly that case.

Widening `OperatorContext.balances` to `Pick<BalanceProvider, "getSnapshot" | "on" | "off">` and
making them required on `BalanceSource` turns this into a compile error at the seam that cares.
(`ui-server.test.ts` passes a plain object today, so it would need the emitter too — which is
arguably worth it, since the liquidity path currently has no UiServer-level coverage.)

sdk/packages/simplex/src/services/server/NotificationService.ts
```diff
@@ -0,0 +1,288 @@
...
+export class NotificationService extends EventEmitter {
+	private state!: NotificationRuntimeState
+	private initialization?: Promise<void>
+	private initialized = false
+	private stopped = false
+	private readonly logger: Logger
+	private readonly notifiedSwapIds = new Set<string>()
+	private readonly onActivity = (event: ActivityEvent) => {
+		if (event.type === "filled" && this.state.settings.swaps) {
+			const fillId = event.txHash ?? `event-${event.id}`
```
seunlanlege: **The `event-${id}` fallback re-opens the duplicate this dedup exists to close.**

Keying on the fill transaction is the right call and does what it should when `txHash` is present.
But when it is null the fallback is the event id, which is unique per row — so two null-hash
`filled` rows for *one* fill produce two alerts. The new test at `notification-service.test.ts:119-122`
encodes that: ids 10 and 11, both `txHash: null`, both alert.

It is reachable because the two recorder paths can both fire for one fill when the order has no
commitment (`recorder.ts:127` records only when `commitment` is unset, `settle` at `recorder.ts:263`
records the observed log), and `orderFillObserved` carries `txHash?: string` — optional all the way
up from `EventMonitor.handleFill`.

Falling back to the order alone keeps the old suppression for that case while still splitting
distinct transactions:

```ts
const id = event.txHash ? `${event.orderId ?? "unknown-order"}:${event.txHash}` : (event.orderId ?? `event-${event.id}`)
```

Low likelihood in practice — a real log always has a hash — but it costs nothing to close.

seunlanlege (commented):
Two follow-ups on the fixes in 246be7c7e — both small, neither blocking. Everything else I raised looks properly addressed, and I've resolved those threads.

Nice touches beyond what was asked: the `error` listener in `openSseStream` closes a latent write-after-end crash path on the SSE responses, and the `ensureReady()` retry cannot double-register listeners since every throw point in `initialize()` precedes the `on(…)` calls.

Unrelated to notifications: 002a52370 ("add real-time private key validation and align status icons", issue #1231) is riding along in this branch — Wizard/Signer/Field/setup-controls plus its own changelog note. Worth splitting out if it wasn't deliberate.

seunlanlege (approved):
LGTM — all 14 review findings addressed across 246be7c7e and 4334193e8, every thread resolved.

Verified locally: `tsc -p` on base vs head introduces no new type errors (the only delta is `web-push`, which just isn't installed in my worktree), and the branch incidentally fixes a pre-existing `bin/simplex.ts` signature error. The `Pick<BalanceProvider, "getSnapshot" | "on" | "off">` contract type-checks against the new emitter-backed test doubles.

Two non-blocking notes for whenever: 002a52370 ("real-time private key validation", #1231) is unrelated to notifications and would be cleaner as its own PR, and neither `simplex` nor `simplex-desktop` gets a version bump here.

seunlanlege:
@ddboy19912 needs rebase

ddboy19912:
> @ddboy19912 needs rebase

There wasn’t an actual Git conflict. GitHub was using stale PR-stack metadata and incorrectly marked the PR as needing a rebase. I removed the stack association and refreshed the unchanged base branch; GitHub now reports the PR as clean and mergeable.

PR #1310, opened 2026-09-22
[simplex, sdk]: bound quorum reads and log scanner retries

seunlanlege:
## What

A quorum read now fails within 12s instead of waiting out its slowest endpoint, and the scanner's retries are logged.

- **`settleUntilQuorum` takes a deadline** (`QUORUM_CALL_DEADLINE_MS`, 12s). Past it, outstanding endpoints count as failures, so the call either decides on the endpoints that answered or throws a `QuorumError` naming the ones it was still waiting on. Early exit on a reachable bar is unchanged, and so is the trust model: a decision still requires the same quorum.
- **Quorum reads use a 5s timeout and one retry**, down from 30s and three. A scan polls every 3s, so a read that has not answered in seconds is already stale. Fill, `eth_call` and gas clients keep their 30s budget.
- **`benchFor` replaces the rate-limit-only suspension check** and names three cases:

  | Failure | Bench | Why |
  |---|---|---|
  | Request-rate limit | `Retry-After`, else 5 min | unchanged |
  | Response that is not JSON | 30s | a plain-text quota notice or HTML error page surfaces as a parse error, so the rate-limit classifier never matched it and the endpoint was re-queried every scan |
  | Head behind the quorum's | 15s | it fails every range at the quorum head until it catches up |

  `QuorumError` labels each failing endpoint with its class: `[rate-limited]`, `[malformed response]`, `[behind the head]`.
- **`RetryConfig.logger` is structural** (`{ trace(message) }`) instead of a `ConsolaInstance`. That type is what kept callers with their own logging stack from passing a logger: `chain-scanner` omitted it, `retryPromise` fell back to the SDK's silent default, and a retrying read logged nothing at any level. `chain-scanner` now passes its module logger and includes the chain id and block range in the message.

## Why

A 24h mainnet run with 9–14 free endpoints per chain had one chain's scanner stop for **366 seconds with nothing in the log**. A slow minority kept the bar unreachable; each attempt waited ~123s for stragglers (30s × 4 transport attempts) and `retryPromise` repeated it three times, holding the scan mutex the whole time. Nothing was logged because the retry lines went to the silent default logger and a later attempt succeeded, so no error surfaced either.

Two endpoints in that set could never have voted on the range being scanned: one answered its quota notice in plain text (a parse error, so never benched), the other was behind the head. With these changes the same episode costs one 12s call that names the endpoints it waited on.

## Validation

- New `src/tests/quorum-failure-handling.test.ts` — 7 tests: the three bench classes and their durations, ordinary failures staying per-call, and the deadline (rejects with `QuorumError` in well under 2s against a hung endpoint, names it, and still decides when the fast responders reach the bar).
- `vitest run src/tests/quorum-public-client.test.ts src/tests/scanner` — 50 passed, 5 skipped (network-gated), no regressions.
- `tsc --noEmit` shows no errors in the changed files.

Behaviour was observed on mainnet across BSC, Ethereum, Base, Arbitrum and Polygon; the fixed cadence held at 0.89–0.99 of expected scans with zero scan errors over 20h, and the tail stalls above are what this change bounds.


Wizdave97 (commented):
Review of the quorum-read bounding and scanner retry logging.

**The bounded call works as described.** The timer is cleared on every exit and `unref`'d, stragglers that fail after the deadline still get benched, rate limits are still classified ahead of the new cases, and the version bumps follow main.

**Two things to fix before merge** (inline):
1. The retry lines are emitted at `trace`, so at the default `info` level they still don't appear.
2. `resolveChainConfigs` now tolerates endpoints that fail the chain-id probe. That also weakens the runtime endpoint-edit check, and the PR description and changelog note don't mention the change.

Other notes (not on diff lines):
- The `QuorumPublicClient` class doc (around L240) still says "The one piece of pausing state is rate limiting … Every other failure mode stays stateless". `suspended()` and `describeResponders` still call every bench "suspended for rate limiting" / "rate-limit-suspended", and `describeResponders`' doc still mentions the "30s×3 retry ladder". All of these are stale now that malformed and behind-the-head responses also bench.
- Changelog note: please add the startup chain-id tolerance and the change in voter-set semantics (inline on `benchFor`).

Reviewed by reading the code; tests were not run.

sdk/packages/simplex/src/scanner/chain-scanner.ts
```diff
@@ -168,10 +168,14 @@ export class ChainScanner {
...
 	private async scan(): Promise<void> {
+		// `logger` is not decoration: a retrying quorum read holds the scan mutex
+		// for the whole of its budget, so without these lines a chain that is
+		// merely retrying looks identical to one that has stopped scanning.
 		const currentBlock = await retryPromise(() => this.quorumClient.getBlockNumber(), {
 			maxRetries: 3,
 			backoffMs: 250,
-			logMessage: "Failed to get current block number",
+			logMessage: `Failed to get current block number on chain ${this.target.chainId}`,
+			logger: this.logger,
```
Wizdave97: **This still logs nothing at the default level.** `retryPromise` only ever calls `logger.trace(...)` (`sdk/packages/sdk/src/utils.ts:348`), and the simplex logger defaults to `info` (`Logger.ts:154`). Passing `this.logger` here doesn't make the retries visible unless the operator runs at `trace`, so the 366s stall from the description would still leave no log lines. The message also leaves out the error.

Suggestion: pass a wrapper that logs at `warn`, e.g. `logger: { trace: (m) => this.logger.warn({ chainId: this.target.chainId }, m) }`. Or have `retryPromise` log at a higher level and include `error`. Same applies to L216.

sdk/packages/simplex/src/scanner/chain-scanner.ts
```diff
@@ -168,10 +168,14 @@ export class ChainScanner {
 	}
 
 	private async scan(): Promise<void> {
+		// `logger` is not decoration: a retrying quorum read holds the scan mutex
+		// for the whole of its budget, so without these lines a chain that is
+		// merely retrying looks identical to one that has stopped scanning.
 		const currentBlock = await retryPromise(() => this.quorumClient.getBlockNumber(), {
 			maxRetries: 3,
```
Wizdave97: This retry loop still wraps the bounded call, so the worst case per scan is 3 × 12s plus backoff, about 37s, rather than the "one 12s call" in the description and changelog. Also, with a 5s timeout and one retry (about 10.5s per endpoint), a single-request read nearly always settles before the 12s deadline anyway. In practice the transport change bounds this; the deadline mostly matters for `getTransactionConfirmations`. Worth wording the claim accurately, or cutting `maxRetries` for the head read.

sdk/packages/simplex/src/services/FillerConfigService.ts
```diff
@@ -78,23 +78,60 @@ export async function fetchChainId(rpcUrl: string): Promise<number> {
...
+): Promise<ResolvedChainConfig[]> {
+	const logger = loggers ? moduleLogger(loggers, "chain-config") : undefined
 	return Promise.all(
 		chains.map(async (chain) => {
 			const rpcUrls = validateRpcUrls(chain.rpcUrls)
-			const chainIds = await Promise.all(rpcUrls.map((url) => fetchChainId(url)))
-			const [first, ...rest] = chainIds
-			for (let i = 0; i < rest.length; i++) {
-				if (rest[i] !== first) {
+			const probes = await Promise.allSettled(rpcUrls.map((url) => fetchChainId(url)))
```
Wizdave97, line 115: **This affects more than boot.** `resolveChainConfigs` is also the probe behind runtime endpoint edits (`simplex.ts:552`), whose comment says: *"Probe before mutating: an endpoint answering for another chain would otherwise silently feed this scanner the wrong chain's logs."* With `allSettled`, a newly added URL that is down during the edit is accepted without its chain ever being checked.

A wrong-chain endpoint usually just fails to join the agreeing group. But on a quiet range it returns `[]`, just like the honest endpoints, so it can count toward a quorum on "no events".

Suggestion: keep the tolerance for boot only (a flag, or a separate function), and still require every newly added URL to answer on edits. Also, this change isn't in the PR description or the changelog note; please add it.

sdk/packages/simplex/src/services/QuorumPublicClient.ts
```diff
@@ -130,6 +171,57 @@ function suspensionMsFor(error: unknown): number {
...
+
+/**
+ * Whether a failure is worth benching the endpoint over, and for how long.
+ *
+ * The three cases share one property: the endpoint will keep failing the same
+ * way for a while, so re-querying it on the next poll only spends the call's
+ * time budget. Everything else — a bad query, a one-off 5xx, a hiccup — stays
+ * per-call, as the class doc promises. Exported for tests.
+ */
+export function benchFor(error: unknown): { durationMs: number; reason: string } | null {
```
Wizdave97, line 238: The description says the trust model is unchanged, but this widens what shrinks the voter set. Before, only rate limits took an endpoint out of the denominator. Now responses that aren't JSON, and endpoints behind the head, do too. Example: 5 endpoints with 2 lagging used to fail the call and retry; now those 2 are benched for 15s and the range is decided by a quorum over 3.

That may be the right trade, but it should be a stated decision in the changelog note, and the class doc should be updated (see the review summary).

sdk/packages/simplex/src/services/QuorumPublicClient.ts
```diff
@@ -34,6 +34,47 @@ const RATE_LIMIT_RPC_CODES = new Set([-32005, -32097, -32016, 429])
...
+/**
+ * Per-endpoint HTTP budget for quorum reads.
+ *
+ * Deliberately far tighter than the wallet/fill clients' 30s × 4 attempts. A
+ * scanner polls every few seconds, so a read that has not answered within a
+ * couple of seconds is already stale — waiting two more minutes for it buys
+ * nothing and costs the scan mutex, which is exactly how a slow minority used
+ * to stall a whole chain. One retry still absorbs a dropped connection.
+ */
+const READ_TIMEOUT_MS = 5_000
```
Wizdave97: The description says "Fill, `eth_call` and gas clients keep their 30s budget", but the fill path's confirmation poller (`filler.ts:806`/`827`) goes through this same class via `ChainClientManager.ts:173`. It now gets the 5s timeout, one retry and the 12s deadline too. `getTransactionConfirmations` makes two requests in sequence per endpoint (head, then receipt), up to about 21s of transport budget, so a slow but honest endpoint can now be cut off by the deadline. The poller retries, so the impact is probably small, but the description should say so. Alternatively, give the confirmation client its own budget.

sdk/packages/simplex/src/tests/quorum-failure-handling.test.ts
```diff
@@ -0,0 +1,106 @@
...
+		const client = new QuorumPublicClient(BASE_CHAIN_ID, URLS, undefined, 60)
+		const clients = client.clients as unknown as Array<{ getBlockNumber: () => Promise<bigint> }>
+		clients[0].getBlockNumber = async () => 100n
+		clients[1].getBlockNumber = async () => 100n
+		clients[2].getBlockNumber = () => new Promise(() => {})
+
+		await expect(client.getBlockNumber()).rejects.toThrow(/c\.example.*quorum deadline/s)
+	})
+
+	it("still decides on the fast responders when they reach the bar", async () => {
```
Wizdave97: This test has no slow endpoint (two URLs, both answering), so it never reaches the deadline code. It passes with or without this PR.

The case where the deadline expires and the call still decides only differs from early exit in `getTransactionConfirmations`, where receipt-not-found votes block early exit. That would be the case to cover: for example, a quorum of endpoints agree on an inclusion, one answers not-found, one hangs → resolves at the deadline.

Wizdave97: Correction to my suggestion above: the example I gave would resolve through early exit, not at the deadline. In all three methods the early-exit check and `finalize` apply the same rule over the same `fulfilled` set (`quorumHead`, `aggregateConfirmations`, group ≥ threshold). So the deadline branch can never resolve where early exit didn't: in practice it always rejects with `QuorumError`. The "decides on the endpoints that answered" wording in the description, changelog and docstring could say that directly, and this test can go or be renamed to cover early exit alone.

sdk/packages/simplex/src/scanner/chain-scanner.ts
seunlanlege: Right, and this was the whole point of the change — thanks. `chain-scanner` now passes an adapter that records at `warn`:

```ts
private get retryLogger() {
  return { trace: (message: string) => this.logger.warn({ chainId: this.target.chainId }, message) }
}
```

and `retryPromise` now appends the error’s first line, so the record says what failed rather than only that something did.

sdk/packages/simplex/src/scanner/chain-scanner.ts
seunlanlege: Correct, and the description was wrong. The head read no longer retries internally (`maxRetries: 1`) — the 3s tick is its retry loop, and a failed head read leaves the cursor untouched, so an inner retry only held the mutex to do what the next tick does anyway. The log read keeps one retry, because a range the cursor is waiting on is worth a second attempt before the tick ends.

Worst case per scan is now ~12s (head) + ~24s (logs), and the changelog states that instead of "one 12s call". Agreed the transport change is what does most of the bounding here.

sdk/packages/simplex/src/services/FillerConfigService.ts
seunlanlege: Good catch — that path would have accepted an unchecked URL, and your `[]`-on-a-quiet-range point is exactly why it matters. Tolerance is now opt-in via `ResolveChainOptions.tolerateUnreachable`: boot and the paymaster keeper set it, `setChainRpcUrls` does not, so every newly added URL must still answer. Added a test for the strict default, and the changelog now covers it.

sdk/packages/simplex/src/services/QuorumPublicClient.ts
seunlanlege: Agreed, and I have stopped claiming otherwise. The class doc now states the widening directly, using your example: five endpoints with two lagging used to fail the call and now decide it over the remaining three — the bar is still a BFT threshold over the endpoints asked, but a call can be decided by fewer than the operator configured. The changelog records it as a deliberate trade rather than a neutral change.

sdk/packages/simplex/src/services/QuorumPublicClient.ts
seunlanlege: You are right that the confirmation poller goes through this class. Rather than only documenting it, budgets are now per call site: `SCAN_BUDGET` (5s / 1 retry / 12s) and `CONFIRMATION_BUDGET` (10s / 1 retry / 30s), with `ChainClientManager.getQuorumClient` taking the latter. That leaves room for the head-then-receipt sequence you describe while still bounding it.

sdk/packages/simplex/src/tests/quorum-failure-handling.test.ts
seunlanlege: True — that test passed with or without the PR. Renamed to say what it actually pins (early exit resolving before the deadline, the path the deadline must not disturb) and added one that does reach the deadline: 3-of-3 with one endpoint hanging, which can only fail.

seunlanlege: This is the sharpest point in the review and you are right: `tryDecide` and `finalize` apply the same rule to the same `fulfilled` set in all three methods, so the deadline branch can never resolve where early exit did not — in practice it always rejects. The docstring, changelog and PR description now say that rather than "decides on the endpoints that answered". I kept the `finalize` call at the deadline because it is what produces the `QuorumError` listing the endpoints that never answered.

Wizdave97 (approved):

PR #1311, opened 2026-09-22
[evm]: one pair per intent order

seunlanlege:
`placeOrder` now requires every leg of an order to name the same input token and the same output
token. An order is a single pair quoted at one or more prices — a ladder — rather than a basket.
Legs on different pairs revert with `InvalidInput`.

Leg-level accounting is unchanged. `_orders`, `_partialFills` and `_protocolFees` stay keyed by
`(commitment, leg index)`. Each leg keeps its own escrow, fill progress and held protocol fee,
different solvers can still fill different legs, and a cancel still refunds each leg's own
remainder.

## Two things follow from the rule

The upper-12-bytes token check now runs once, on leg 0, for the inputs and for the outputs. The
other legs must equal leg 0 byte for byte. That also closes the alias case: `T` and `T | 1 << 255`
can no longer be two legs of one order.

An order carrying both predispatch calldata and predispatch assets must be single-leg. Its escrow
is swept from the `CallDispatcher` and measured per input token, so two of its legs never could
share a token. Now that every leg holds the same token, that is simply a leg count, and a length
check replaces the pairwise comparison that used to express it. Predispatch only runs when both
fields are set; calldata without assets is ignored, as before.

## Not changed

The fill path. `_validateLegs` still checks every leg's tokens against the solver's quote, and still
rejects an output token with its upper 12 bytes set, because the destination chain of a cross-chain
order never sees `placeOrder`. `_execute` still sweeps each output token at its first leg only.

The ABI. The rule reuses `InvalidInput`, so the gateway ABIs carried by the SDK, Simplex and the
indexer stay as they are.

## Validation

`forge build --sizes` is clean. The implementation is 18,220 of 24,576 bytes, two bytes smaller than
before.

307 tests pass across the seven suites that reach `placeOrder`: MultiLeg, SameChain, RateFill,
ProtocolFees, Modules, Reentrancy and V2Test.

The two multi-output reentrancy tests were checked by mutation, not just by running them: with
`nonReentrant` and the `Filled()` check both stripped from `fillOrder`, each one fails. That caught
two ways they had been passing for the wrong reason — a reentrant payload with no ETH to pay a
native leg, and a cross-chain redeem dying on `UnknownInstance()` before any guard was reached. Both
are fixed. `testReentrancy_FeeTheft` and `testCrossChain_ReentrancyBlocked` still pass unguarded,
since their payloads quote zero on every leg; that predates this PR and is left alone.

Three tests placed orders the rule no longer allows and were reworked:

- `testSameChain_RepeatedNativeOutputTracksLegsSeparately` (SRLabs S2-15) used a USDC leg and a DAI
  leg to make per-leg escrow separation observable. It now uses two USDC legs of different size and
  asserts on `_orders(commitment, 1)`.
- `testReentrancy_EscrowTheft_MultiOutput` and `testCrossChain_ReentrancyBlocked_MultiOutput` used
  an ETH output leg and a DAI output leg. Both are now one pair at two prices.

New coverage: a native-input ladder (per-leg `msg.value` deduction and the leftover refund), and an
alias on a later output leg.

One note for reviewers: adding a test function to `IntentGatewayV2MultiLegTest` tipped solc into an
internal compiler error under via-IR (`Tag too large for reserved space`). The placement-revert
tests now share a `_placeReverts` helper and the file no longer carries a mock token, which bought
the space back, but it is sitting near the limit.

## Rollout

The rule takes effect on a chain once governance upgrades that chain's implementation, so for a
while some chains will accept an order shape that others refuse. `VERSION` stays 3, since no proxy
needs re-initialising, which means `version()` does not tell an upgraded gateway from one that still
accepts baskets.

Orders already placed are unaffected; their commitments, escrow and pending messages run through the
same code paths as before. Any integrator posting legs on different pairs starts seeing
`InvalidInput`.


evm/tests/foundry/IntrinsicIntentsReentrancyTest.sol
```diff
@@ -333,12 +332,9 @@ contract IntrinsicIntentsReentrancyTest is MainnetForkBaseTest {
 
         // ── 2. Arm the malicious beneficiary ─────────────────────────────────
 
-        deal(address(dai), address(maliciousBeneficiary), outputDai);
-        maliciousBeneficiary.approveGateway(address(dai), outputDai);
-
         TokenInfo[] memory reentrantOutputs = new TokenInfo[](2);
         reentrantOutputs[0] = TokenInfo({token: bytes32(0), amount: 0});
-        reentrantOutputs[1] = TokenInfo({token: daiToken, amount: outputDai});
+        reentrantOutputs[1] = TokenInfo({token: bytes32(0), amount: outputEth2});
```
Wizdave97, line 339: **This test would still pass if the escrow-theft bug were reintroduced.**

Leg 1 is now native, but `ReentrantBeneficiary.receive()` re-enters with `gateway.fillOrder(storedOrder, storedOptions)` and **no value**. `_payLeg` (`IntentsBase.sol:560`) checks `nativeRemaining < paid` against `msg.value == 0`, so the reentrant fill always reverts with `InsufficientNativeToken` — guard or no guard. That fails the beneficiary's `receive`, and the outer call reverts with the same `ERR_INSUFFICIENT_NATIVE` the test expects.

Previously the attacker genuinely paid the DAI leg and could claim leg 1's USDC escrow; only the `Filled()` guard stopped it.

Suggested fix: add a stored `reentrantValue` to `ReentrantBeneficiary` and re-enter with `gateway.fillOrder{value: reentrantValue}(...)` — it has just received 1 ETH from leg 0, so it can fund `outputEth2`. Arm it with `outputEth2`.

(Pre-existing: `fillOrder` is also `nonReentrant`, so none of these tests would catch reverting only the CEI ordering.)

evm/src/apps/IntentGatewayV2.sol
```diff
@@ -214,18 +219,8 @@ contract IntentGatewayV2 is
...
-                    unchecked {
-                        ++j;
-                    }
-                }
-                unchecked {
-                    ++i;
-                }
-            }
+            // The sweep below measures one balance per input token, so it needs a leg to itself.
+            if (inputsLen != 1) revert InvalidInput();
```
Wizdave97, line 223: Correct and load-bearing. Minor: this only runs when `predispatch.call` **and** `predispatch.assets` are both non-empty. A multi-leg order with calldata but no assets is accepted and its calldata is silently never executed (pre-existing, no funds risk). The new NatSpec and `overview.mdx` say "An order with predispatch calldata must be single-leg", which overstates this — either reword to "calldata and assets" or reject calldata without assets.

sdk/packages/core/contracts/apps/IntentGatewayV2.sol
```diff
@@ -562,6 +562,9 @@ interface IIntentGatewayV2 {
 
     /**
      * @notice Places an order for cross-chain intent fulfillment.
+     * @dev Leg `i` sells `order.inputs[i]` for `order.output.assets[i]`. Every leg trades the same
+     *      pair: all inputs name one token and all outputs name one token, so an order is one pair
+     *      quoted at one or more prices. An order with predispatch calldata must be single-leg.
```
Wizdave97: Enforced only when both `predispatch.call` and `predispatch.assets` are non-empty (see comment on `IntentGatewayV2.sol:223`). Suggest "An order with predispatch calldata and assets must be single-leg."

docs/content/developers/evm/intent-gateway/overview.mdx
```diff
@@ -31,7 +33,7 @@ An order is a list of legs: leg `i` sells `inputs[i]` for `output.assets[i]`. `i
 
 The gateway tracks each leg's escrow, fill progress and protocol fee separately, keyed by the order commitment and the leg index (`_orders`, `_partialFills` and `_protocolFees`). Different solvers can fill different legs, and filling or cancelling one leg never releases another leg's escrow.
 
-An order with predispatch calldata is the one exception: its escrow is swept from the dispatcher and measured per token, so its legs may not repeat an input token.
+An order with predispatch calldata must be single-leg: its escrow is swept from the dispatcher and measured per token, and every leg holds the same token.
```
Wizdave97: Same as the NatSpec: the single-leg rule applies when predispatch has both calldata and assets.

evm/tests/foundry/IntentGatewayV2MultiLegTest.sol
```diff
@@ -557,12 +562,25 @@ contract IntentGatewayV2MultiLegTest is MainnetForkBaseTest {
...
+        Order memory mixedOutputs = _ladder("", host.host());
+        mixedOutputs.output.assets[1].token = usdcToken;
+        _placeReverts(mixedOutputs);
+    }
+
+    /// @notice A token is the address in the low 20 bytes, so `T` and `T | 1 << 255` are one token. Legs must
+    /// name it in one form: otherwise, with a token whose `transfer` returns false instead of reverting, the
+    /// two copies would each credit the one deposit that landed, and a cancel would pay the second copy out of
+    /// other orders' escrow in that token.
+    function testPlaceOrder_RejectsAliasedInputToken() public {
```
Wizdave97: Nit: with the equality check, the aliased leg now reverts before any transfer, so the `FalseReturningToken` setup and the "other order's escrow" assertion no longer exercise anything. Harmless, but the docstring's scenario is no longer what this tests — could be simplified to a plain `_placeReverts`.

evm/docs/ai/changelog/2026-09-22-one-pair-per-intent-order.md
```diff
@@ -0,0 +1,39 @@
...
+`order.inputs` must name one token and all of `order.output.assets` must name one token, or
+placement reverts with `InvalidInput`. An order is therefore a single pair quoted at one or more
+prices — a ladder — rather than a basket.
+
+Leg-level accounting is unchanged. `_orders`, `_partialFills` and `_protocolFees` stay keyed by
+`(commitment, leg index)`, so each leg still carries its own escrow, fill progress and held
+protocol fee, different solvers can fill different legs, and a cancel refunds each leg's own
+remainder.
+
+## What the rule replaces
```
Wizdave97: Per CLAUDE.md's docs rules, the "replaces / now runs once / used to express it" framing here and the `## Files` section are implementation history. Suggest keeping just the lasting rules (no `T` / `T | 1<<255` aliasing within an order; predispatch orders are single-leg) and the "no ABI change" paragraph, plus a line that the rule takes effect per chain on implementation upgrade.

Wizdave97 (commented):
Adversarial review (4 parallel reviewers: placement bypass, fill/settle/cancel, tests, integrators/docs).

**Contract change looks sound.** Leg-0 upper-bytes check + byte-equality covers every leg; `inputsLen != 1` is load-bearing (the predispatch sweep measures one balance per token, so two same-token legs would double-count); native `msg.value` is deducted per leg correctly; fill/settle/cancel are all keyed by leg index so the same-pair case is handled; Simplex already rejects multi-leg orders and nothing in SDK/indexer builds baskets. No storage layout change behind the proxy.

**One thing to fix before merge:** the rewritten `testReentrancy_EscrowTheft_MultiOutput` no longer detects the attack (inline).

Other notes (not on diff lines):
- `IntrinsicIntentsReentrancyTest.sol` `testCrossChain_ReentrancyBlocked_MultiOutput` (~L450) still builds a USDC→ETH / USDC→DAI order. It passes because the destination never sees `placeOrder`, but the shape is now unplaceable — convert it like its same-chain twin or update the docstring.
- `sdk/packages/core/docs/ai/changelog/2026-09-17-intent-gateway-escrow-keyed-by-leg.md:15` references `testPlaceOrder_PredispatchRejectsAliasedInputToken`, renamed here to `testPlaceOrder_RejectsAliasedInputToken`.
- Rollout: the rule only takes effect once governance upgrades the implementation on each chain, and `VERSION` stays 3 so `version()` can't distinguish upgraded gateways. Worth stating in the PR's Rollout section.
- Coverage gaps (low): no test places a 2-leg native-input ladder (per-leg `msgValue` deduction + leftover refund); no test for a `T | 1<<255` alias on a later *output* leg.

Forge not run locally (fork tests need `MAINNET_FORK_URL`); findings are from code reading, and the main one was verified by hand.

evm/tests/foundry/IntrinsicIntentsReentrancyTest.sol
seunlanlege: Confirmed, and it was worse than reading suggested. I checked by mutation: stripped both `nonReentrant` and the `Filled()` check from `fillOrder`, and the test still passed. It now fails.

Fixed the way you suggested — `ReentrantBeneficiary` stores a `reentrantValue` and re-enters with `{value: ...}`, funded by the ETH leg 0 has just paid it.

Running the cross-chain twin under the same mutation turned up a second one. It passed too, but because the reentrant fill paid leg 1 and then reverted `UnknownInstance()` dispatching the redeem to `EVM-2`, which `setUp` never registered as a peer — so it was never reaching a guard at all, before or after this PR. Peer registered, and both multi-output tests now fail when the guards come out.

`testReentrancy_FeeTheft` and `testCrossChain_ReentrancyBlocked` still pass unguarded: their payloads quote zero on every leg, so there is nothing to steal either way. Pre-existing, left alone.

evm/src/apps/IntentGatewayV2.sol
seunlanlege: Right. Reworded rather than changed — the NatSpec, the core interface and `overview.mdx` now say "carrying both predispatch calldata and predispatch assets", and the docs add that calldata without assets does not run.

Rejecting that shape is a behaviour change, so I've kept it out of this PR.

sdk/packages/core/contracts/apps/IntentGatewayV2.sol
seunlanlege: Done — "An order carrying both predispatch calldata and predispatch assets must be single-leg."

docs/content/developers/evm/intent-gateway/overview.mdx
seunlanlege: Done, same wording, plus a line that predispatch only runs when both fields are set.

evm/tests/foundry/IntentGatewayV2MultiLegTest.sol
seunlanlege: Agreed. Folded the alias cases into `testPlaceOrder_RejectsLegsOnAnotherPair` — a later leg naming `T | 1 << 255`, on the input side and the output side — and deleted `FalseReturningToken`, which nothing used any more.

That also bought back headroom: adding a test function to this file had tipped solc into a via-IR ICE (`Tag too large for reserved space`), which is why the placement-revert tests share a `_placeReverts` helper now.

evm/docs/ai/changelog/2026-09-22-one-pair-per-intent-order.md
seunlanlege: Rewritten to the rules only — no "what it replaces" framing, no `## Files`.

Added the rollout paragraph you asked for: the rule lands per chain as governance upgrades each implementation, so some chains will accept a shape others refuse, and `VERSION` stays 3 so `version()` cannot tell them apart.

I don't think `VERSION` should go to 4. It is the `Initializable` generation, and bumping it would force a `migrate` on every proxy for a change that needs no re-initialisation.

Wizdave97 (approved):

PR #1306, opened 2026-09-21
[sdk, simplex]: fix bidding on testnet: SolverAccount version check, cNGN matching, outbid retraction, fee rounding

Wizdave97:
## Summary

These are four fixes found by running three testnet simplex solvers against a live HyperFX orderbook.

### SolverAccount is no longer version-checked (sdk)
#1292 removed `version()` from `SolverAccount`. #1287 made the SDK require the solver's live delegation and the configured `SolverAccount` to report release 3 before a bid is signed. Accounts built from `main`, including the testnet SolverAccount `0x153DB990FE3b761B54ad71D0f1A0987dF11740DB` from #1301, revert on the getter. So `prepareSubmitBid` refused every bid, and simplex could neither bid nor post limit orders:

```
Fills are not supported by the destination gateway, live delegation, and configured SolverAccount
```

The release check now reads only the gateway:
- `supportsRateFills(client, gateway)` drops its `solverAccount` argument.
- `RateFillCapabilityReader` and `readRateFillCapability` take `(evmRpcUrl, gatewayAddress)`.
- `prepareSubmitBid` throws `Fills are not supported by the destination gateway`.
- `GasEstimator` relies on `assertGatewayRelease`.

Phantom aggregation still requires the sender to be delegated to a configured `SolverAccount`.

### Cross-chain order fees round their 5% buffer up (sdk)
On testnets `convertGasToFeeToken` returns `1n`, so a cross-chain quote was `(2 × 105) / 100 = 2`, exactly the solver's requirement. Simplex requires a strictly positive fee profit, so it refused every SDK-placed cross-chain testnet order (`feeProfit = 0`, `profitable: false`). `quoteOrderFees` now computes `(sum × 105 + 99) / 100`.

### Swaps paying cNGN in match limit orders (simplex)
`serves()` compared the book's leg (`cNGN`, as the orderbook spells it) with the registry's upper-cased symbol (`CNGN`). Every swap paying cNGN in was skipped with `No limit order matches this order`. `serves()` and the same-asset check in `FXFiller` now compare through `normalizeSymbol`.

### Outbid bids are retracted when a rival completes the order (simplex)
`EventMonitor.handleFill` returned early on a rival's fill. A losing solver kept its bid, its limit-order hold and its 0.1 BRIDGE deposit until the stale-bid sweep, which only takes bids older than an hour.
- `ScannedFill` gains an optional `complete` field: `true` for `OrderFilled`, `false` for `PartialFill`.
- `orderFillObserved` passes it through.
[...]

PR #1312, opened 2026-09-22
[simplex, sdk]: size later legs from a shared limit order, skip bids on completed legs, limit order UI

Wizdave97:
Follow-up to #1306. With #1311, every leg of an order is on the same pair, so it is now common for two legs to match the same limit order.

## Simplex: a later leg is sized from what an earlier leg left
Each leg's bid used to be sized from the limit order's full `available`. When two legs matched the same limit order, the bids together promised more than the order holds. The later leg was then dropped at reservation, so it got no bid (seen live: `multi-leg-same-input` stranded leg 1). `calculateProfitability` now tracks what earlier legs of the same swap order have planned against each limit order (`plannedOn`). A later leg bids on what is left, as a partial fill, and is skipped only when nothing is left.

## SDK: bids on a completed leg are not offered
`OrderExecutor` kept offering bids whose only non-zero outputs were on legs the destination had already credited in full. These reverted in simulation with `RateFillTooSmall` (0xc20ebedd). Each round now drops a bid unless it pays into at least one open leg (`servesOpenLeg`). A bid whose outputs don't line up with the order's legs is still offered.

## Operator UI
- The live and closed limit order lists are paginated, 10 orders per page.
- `New limit order` is a primary pill button.
- The detail panel shows fills as a table (when, amount paid out, transaction link), with the total paid out below it.

## Validation
- SDK: `bidSelection`, `fillOrderCodec`, `phantomAggregation` and `orderFees` all pass (122 tests). The new completed-leg test fails with the filter disabled.
- Simplex: 873 tests pass, 25 skipped (live/testnet/rebalancer suites excluded). The new test is `sizes a later leg from what an earlier leg left on a shared limit order`.
- Biome lint is clean. The UI type-check is clean for the changed files.


seunlanlege (approved):
Approving. I checked the three things that could have gone wrong in the leg-sizing change and they're all right:

- `plannedOn` units line up. It accumulates `toScaled(finalOutputAmount)` at 1e18, the same scale as `candidate.available`, and since `toScaled(toRaw(x)) <= x` the running total can never over-debit an order.
- `plannedOn.set` sits after every `continue` gate, so a skipped candidate doesn't consume capacity.
- `servesOpenLeg` indexes `remainingAssets` the same way `syncWithDestination` builds it (from `order.output.assets`, by index), and it's refreshed at the top of every round.

Locally: 49 tests pass across the changed SDK files, 103 across simplex `fx.payout` + `cli`. Biome is at baseline.

Four nits, none blocking:

**Worth doing —** the page clamp already has a cleaner form one file over. `Wallet.tsx:158` computes `const current = Math.min(page, lastPage)` during render; `LimitOrders.tsx` uses a `useEffect` instead. The effect version renders one empty frame when a page disappears under a refresh: the slice is empty, but `live.length !== 0` so the "no orders" hint doesn't show either. Matching Wallet.tsx drops both the effect and the flash.

- `.limit-order-fills-head` needs `!important` only because `.limit-order-fills li` (0,1,1) outranks it (0,1,0). `.limit-order-fills li.limit-order-fills-head` removes both.
- The fills list is described as a table but built from `<ol>`/`<li>` with an `aria-hidden` header row. A real `<table>` gets the header semantics for free.
- Stray trailing blank line at the end of `bidSelection.test.ts`.

Needs a version bump before release — sdk and simplex together, since sdk changed.

Wizdave97:
Thanks. All four nits and the version bump are in 60e4cdfff:

- **Page clamp:** `LimitOrders.tsx` now computes `liveCurrent` / `closedCurrent` during render (`Math.min(page, lastPage)`), as `Wallet.tsx` does. The `useEffect` is gone, and with it the empty frame when a page disappears under a refresh.
- **Fills markup:** the list is now a real `<table>`, with a `<thead>` of `scope="col"` headers and one `<tr>` per fill, so it has header semantics without the `aria-hidden` row.
- **`!important`:** the header styles are now on `.limit-order-fills th`, so nothing competes and both `!important`s are gone. On narrow screens the header is hidden and each row stacks as before.
- **Trailing blank line:** removed from `bidSelection.test.ts`.
- **Versions:** `@hyperbridge/sdk` 2.8.16 and `@hyperbridge/simplex` 0.16.5, plus `simplex-desktop` 0.16.5, matching how #1301 moved them together.

UI type-check, Biome check and `vite build` are clean for the changed files.

PR #1313, opened 2026-09-22
[sdk, docs]: quote intents, liquidity and rates from the HyperFX orderbook

Wizdave97:
## Summary

`IntentGateway.quoteIntent`, `queryAvailableLiquidity` and `queryBuyAndSellRates` now read the HyperFX orderbook's GraphQL API exclusively. Every other intent quoting path is removed.

### SDK
- **`HyperFxOrderbook`** is a new read-only client for the orderbook (`topOfBook`, `routeLiquidity`, `quote`). Amounts and rates are 1e18 `bigint`s, and rates are quote per 1 base.
- **`OrderbookMarket`** resolves configured token addresses and symbols to orderbook symbols and chain ids, and converts raw amounts to and from 1e18.
- **`quoteIntent`**
  - Exact input uses the orderbook's clearing price, which can combine several solvers' orders.
  - Exact output starts from the best rate and re-quotes at each clearing price until the output is delivered. The input is rounded up to a whole raw unit.
  - Throws `InsufficientOrderbookLiquidityError` (carrying `maxFillableIn`) when the route cannot fill.
  - No protocol-fee adjustment: orderbook prices already carry the gateway fee haircut.
- **`queryAvailableLiquidity`** returns the route kind, side, best rate, true liquidity, virtual depth, `maxFillableIn`, and order and solver counts.
- **`queryBuyAndSellRates`** returns `bid`, `ask`, `mid`, `spread` and `spreadBps` for orders filled on the destination that accept the source chain.
- **Orderbook deployments:** `IntentGateway` reads the deployment for its destination chain's network (`ORDERBOOK_URLS`, `orderbookUrlFor`): `https://orderbook.hyperfx.finance/testnet/graphql` for testnet chains, `…/mainnet/graphql` otherwise. **`withOrderbook(url | client)`** overrides it. These three methods no longer need `withQueryClient`.

### Config
- BSC Chapel and Polygon Amoy now configure cNGN: the hCNGN testnet stand-in listed in the indexer's `config-testnet.json`, with 6 decimals (`0x2bbbd701…778a` and `0xe4ff5d2a…a758`).
- Simplex defaults to the same per-network URLs (`DEFAULT_ORDERBOOK_URLS`). Its setup wizard writes the one for the selected network, and legacy-config migration no longer pins mainnet.

### Removed
- The `indexed_rates`, `phantom_snapshot` and `uniswap_v4` quote strategies, and the `strategy` and `uniswapV4` params.
- `LiquidityEngine`, the `AVAILABLE_LIQUIDITY`, `BUY_AND_SELL_RATES` and phantom-snapshot queries, and the Uniswap V4 pool list in chain config.
- Phantom-order quoting:
  - `aggregatePhantomBids`, with its balance and V4-position weighting and its haircuts;
  - the coprocessor's phantom-order polling, block scan, and phantom-bid submission.
- **Kept:**
  - The phantom bid declaration codec, now in `phantom-bid.ts`. Simplex uses it to post orders.
  - `fetchBidsForOrder` and `setAggregationFetch`, which the indexer uses.
  - `quoteUniswap`, `UniswapQuoteEngine` and `Swap`, which convert gas to fee tokens and value bids. They are not intent quotes.

### Breaking changes
- `QuoteIntentResult` has no `strategy` field; its metadata is now `route`, `side`, base and quote symbols, `rate`, `maxFillableIn` and `orderCount`.
- `AvailableLiquidity` and `BuyAndSellRates` have new shapes.
- The removed exports are listed above.

### Docs
- New page: *Querying the Orderbook* (`developers/evm/intent-gateway/orderbook`). It covers conventions, rates, top-of-book and per-route spreads, liquidity, depth ladders, quotes, and `rateBars`/`bookHistory`.
- The SDK API reference and the placing-orders guide are updated.
- SDK AI notes describing deleted code are removed, and a changelog note is added.

## Validation
- **Unit tests:** `orderbookMarket.test.ts` (12 tests, including deployment selection, stubbed transport, on BSC Chapel and Amoy tokens) covers scaling, exact-output convergence and rounding, unfillable routes, GraphQL errors, orientation and spreads.
- **Full SDK unit suite:** 208 passed. The only failures are 5 cases in `stateCalls.test.ts`, which could not reach a live RPC.
- **Type checks:** `tsc` is clean for changed files. The 4 remaining errors are in test files this PR does not touch. Simplex and the indexer compile against the new exports.
- **Live suite:** `IntentGateway orderbook reads` passed 4/4 through `IntentGateway` against the deployed testnet orderbook on BSC Chapel and Polygon Amoy, selected by default. It covers rates, same- and cross-chain liquidity, and exact-in and exact-out quotes. `HYPERFX_ORDERBOOK_URL` overrides the URL; `BSC_CHAPEL`/`POLYGON_AMOY` override the rate-limited public RPCs.
- **Docs queries:** every GraphQL query on the new page ran against a local orderbook and the deployed testnet orderbook.

## Follow-ups
- The mainnet deployment (`/mainnet/graphql`) is not live yet, so mainnet quotes fail until it is.


seunlanlege (approved):
Approving, but there is one regression to fix before this merges.

## `migrateLegacyConfig` no longer writes `[orderbook]`

`migrate-legacy.ts` stopped setting `config.orderbook`, but `filler-toml.ts:364` still requires it. Confirmed both ways: **5 of the 6 tests in `src/tests/cli/migrate-legacy.test.ts` fail on this branch; all 6 pass on main.**

The user-facing effect is worse than the test names suggest. In `cli/init/index.ts:84` the migration runs, then `validateConfig(config)` throws `an [orderbook] section is required` into the *outer* catch. The wizard then reports the config "doesn't pass validation" and offers only start-fresh, so every legacy prefill — chains, signer, pairs, vault — is discarded.

`assembleConfig` filling it in later doesn't help, because validation happens first. Either keep setting `config.orderbook` in the migration, or fill it in before validating. The note it now pushes ("Added [orderbook] for the selected network") also claims something it no longer does at that point.

## Verified good

- SDK unit suite: 209 passed, only `stateCalls.test.ts` fails (5 cases, live RPC) — matches the PR description exactly.
- `tsc --noEmit` clean on sdk; simplex and indexer sit at their pre-existing baselines.
- The cNGN config is right. I queried both chains: `0x2bbbd701…778a` on BSC Chapel and `0xe4ff5d2a…a758` on Amoy each return `decimals() = 6` and `symbol() = hCNGN`, and both match `config-testnet.json`.
- All three new GraphQL documents validate against the pinned `orderbook-schema.graphql`.
- Nothing is stranded by the phantom removal: the indexer dropped phantom indexing on 2026-09-14, its only remaining use is `fetchBidsForOrder`/`setAggregationFetch` in `utils/bid-data.ts` (both kept), and simplex only uses `encodePhantomBidDeclaration`.
- No dangling references to the 14 deleted markdown files, and the new docs page is registered in `meta.json`.

## Worth fixing

**Nothing checks the new queries.** `simplex/src/tests/orderbook/schema.test.ts` already validates simplex's documents against the pinned schema, and its docstring describes this exact failure: "a field renamed on the server ... would pass all of them and fail on the first real request." The SDK's three documents are now in that position. `ORDERBOOK_QUERIES`' own comment says "so they can be checked against the orderbook's published schema" — but nothing does. Adding them to that test, plus `sdk/packages/sdk/src/protocols/intents/orderbook/**` to `check-orderbook-schema.yml`'s paths, closes it.

**`quoteExactOutput` reports non-convergence as insufficient liquidity.** Both the `break` (genuine insufficiency) and running out of `MAX_EXACT_OUTPUT_ROUNDS` fall through to the same `throw this.insufficient(route, liquidity.maxFillableIn, ...)`. In the second case the caller gets "cannot fill" carrying a `maxFillableIn` that contradicts the message. A distinct error, or just different wording, saves someone a confusing debugging session. Each round is also a sequential round-trip, so the worst case is 17 requests.

**`buildQuote` casts `quote.rate as bigint`.** If a `fillable` quote ever comes back with `rate: null`, that's a `formatUnits(null, 18)` crash rather than a clear error.

## Notes

- There are unrelated prettier rewraps in otherwise-untouched files (`emit-toml.ts`, `migrate-legacy.ts`, `write.ts`, `ChainConfigService.ts`). They make the diff harder to read than it needs to be.
- `assembleConfig` keeps `base.orderbook` verbatim on an update run, so an operator still pointing at `orderbook.hyperbridge.network/graphql` silently keeps a dead URL.
- Mainnet `quoteIntent` fails until the mainnet deployment is live. Worth deciding explicitly whether this lands before or after that.
- Needs a version bump. This is breaking for the SDK — removed exports, reshaped `QuoteIntentResult`, `AvailableLiquidity` and `BuyAndSellRates` — so more than a patch.

Wizdave97:
Thanks for the review. Addressed as follows.

**`migrateLegacyConfig` / `[orderbook]`** — fixed in d415a72d2: migration writes `[orderbook]` again, so the migrated config validates and legacy prefills survive. The wizard's `assembleConfig` then swaps a built-in default for the selected network's URL. The migration note now says exactly that (6cba7d9d0).

**Nothing checks the new queries** — `simplex/src/tests/orderbook/schema.test.ts` isn't in any CI step (`test:unit` covers `cli` + `logger` only), so the check went into the SDK instead: `orderbookSchema.test.ts` validates the three `ORDERBOOK_QUERIES` documents and every GraphQL example on the new docs page against the same pinned schema. It runs in the SDK concurrent CI step. `check-orderbook-schema.yml` now also watches `sdk/packages/sdk/src/protocols/intents/orderbook/**` and the docs page.
Doing this showed the pin had drifted from `hyperfx-orderbook` main: `RateEntry` has gained `base`/`quote` (the docs use them), and `MISSING_VALID_UNTIL` is gone from `RejectionCode`. The pin is refreshed, and the code is dropped from simplex's `REJECTION_CODES` so its schema test still holds. Against the old pin, the new test fails on the docs example using `base`/`quote`, as intended.

**`quoteExactOutput` non-convergence** — running out of rounds now throws `OrderbookQuoteNotConvergedError` (carrying `rounds` and the last input tried); `InsufficientOrderbookLiquidityError` is only thrown once the input passes `maxFillableIn`. Unit-tested, including that it stops at 16 quotes.

**`quote.rate as bigint`** — the cast is gone; a fillable quote served with a null rate is an `OrderbookRequestError` naming the problem. Unit-tested.

**Prettier rewraps** — reverted in `emit-toml.ts`, `migrate-legacy.ts`, `write.ts` and `ChainConfigService.ts`; only the substantive lines differ from main now.

**Dead URL kept on update** — `RETIRED_ORDERBOOK_URLS` lists `https://orderbook.hyperbridge.network/graphql`; the wizard replaces it with the selected network's URL, like the built-in defaults, and still keeps an operator's own URL. Tests cover both.

**Version bump** — leaving versions as they are: 2.8.15 hasn't been published yet, so this ships in it.

**Mainnet** — open: `quoteIntent` on mainnet chains needs `https://orderbook.hyperfx.finance/mainnet/graphql` to be live; only testnet is deployed today.

CI on d717a4b13 is green (the one `fx.testnet` timeout was a Hyperbridge tx-pool `1014 Priority is too low` collision with concurrent `simplex/testnet-swap-e2e` runs sharing the testnet key; the rerun passed).

PR #1314, opened 2026-09-22
[indexer]: fix missing handleGetRequestTimeoutHandledEvent on EVM chains

Wizdave97:
## Summary
The generated EVM chain manifests route `GetRequestTimeoutHandled(bytes32,string)` logs to `handleGetRequestTimeoutHandledEvent`, but the mapping exported the handler as `handleGetRequestTimeoutHandled`. SubQuery only resolves a handler when it first sees a matching log. So each EVM indexer crashed with `Handler function handleGetRequestTimeoutHandledEvent is not found` on its first GET request timeout. This is currently happening on the Ethereum mainnet indexer.

This PR renames the export to `handleGetRequestTimeoutHandledEvent`, which matches the manifest and the `*Event` naming of the other EVM host handlers. No reindex is needed: the failing block was never written, so indexing resumes from it.

## Validation
- Checked every handler name in `scripts/generate-chain-yamls.ts` against `src/mappings/mappingHandlers.ts`. This was the only mismatch.
- Not built or tested locally.


seunlanlege (approved):
Verified independently: every `handler:` name in `scripts/generate-chain-yamls.ts` now resolves to an export in `src/mappings/mappingHandlers.ts`, and this was the only mismatch. Nothing else in the repo — manifest, yaml or script — still names the old export.

Good catch on a failure that only shows up on the first matching log.

PR #1316, opened 2026-09-22
[simplex, sdk]: testnet swap harness in CI; sort bids by rate

Wizdave97:
Adds the testnet swap harness to CI as the `simplex testnet swaps` workflow. It runs three simplex solvers against the live HyperFX orderbook on BSC Chapel and Polygon Amoy. Users place orders, and the SDK executes the solvers' bids until each order is filled on the destination chain. Settlement through Hyperbridge is not awaited.

## What runs
- `e2e/run.mjs` does the following:
  - writes each solver's config from env variables and starts the solvers;
  - posts a fresh book before every scenario;
  - runs each scenario in its own process (`e2e/swap.mjs`);
  - withdraws the book and stops the solvers at the end;
  - writes a results table to the job summary.
- A scenario passes when its order reaches `FILLED` in at least `minFills` fill transactions.
- Scenarios: `same-chain`, `cross-chain`, `partial`, `same-solver-levels` and `multi-leg-same-input`. `multi-leg` and `multi-leg-levels` put two pairs in one order, which #1311 forbids, so they run only when named. `workflow_dispatch` takes a comma-separated subset.
- Wallets are rebalanced in both directions, so they only need gas. Before the run, a short user is topped up from a solver. Before each book is posted, a solver that can't back its limit orders is topped up from a user. The run fails early if no wallet can spare a top-up.

## Secrets and endpoints
All keys and URLs come from repository secrets (`E2E_*`). Only public contract addresses are committed. Error text and printed solver logs are redacted against every variable, both as given and as completed, and solver logs are not uploaded. The variables are listed in `sdk/packages/simplex/docs/ai/changelog/2026-09-22-testnet-swap-e2e.md`.

## Triggers
The workflow runs on pushes to main, on PRs that touch the simplex strategy, core or orderbook code, the SDK intents code or the harness, and on demand. Runs share wallets, so one runs at a time and a started run is never cancelled; GitHub keeps one pending run per group, so a newer run replaces a pending one. Draft PRs and PRs from forks are skipped. Only the protobuf submodule is fetched, and the pnpm store is cached.

## SDK: bids sort by rate
`BidManager.sortBids` is now one function that sorts bids by rate, best to worst. A leg's rate is the bid's output over its input on that leg. Every leg trades the same pair (#1311), so a bid's rate is its total output over its total input across the legs it quotes. Rates are compared exactly; ties keep arrival order.

A bid is dropped when it doesn't quote the order's legs one for one, names another token on a leg, quotes one side of a leg only, quotes no leg, or quotes a leg below the order's rate. Sorting no longer calls a DEX; `Bid.outputUsdValue` still does. The four previous strategies are removed: exact rate for one output, $1 per stable, DEX-quoted USD, and raw-amount fallback.

## Notes
- The CI wallets must not be used by any other simplex instance at the same time.

## Validation
- CI passed all 7 scenarios against `https://orderbook.hyperfx.finance/testnet`. Fill counts were 1, 1, 2, 2, 3, 5 and 4. `/graphql` is appended to the orderbook URL when missing.
- SDK tests pass: `mandatoryQuoteRanking`, `bidSelection`, `fillOrderCodec`, `phantomAggregation` and `orderFees` (126 tests).


seunlanlege (approved):
Approving. The redaction is genuinely thorough — every path that prints goes through `redact`: scenario stdout, solver log tails, the summary table, `summary.json`, and the `main().catch` handler. Solver configs are written `0o600` and never uploaded. `dist/bin/simplex.js` is what `scripts/build.sh` produces, and protoc installs before both builds with no install-time hooks to trip on.

Two things worth changing:

**No draft guard.** With `types: [opened, synchronize, reopened, ready_for_review]` and no `github.event.pull_request.draft == false`, every push to a draft PR touching those paths spends testnet funds and takes the serialized slot.

**The concurrency comment overstates what it buys.** It says runs "never cancel one another", but with `cancel-in-progress: false` GitHub still cancels the *previously pending* run when a third joins the group — so with three PRs in flight one shows as cancelled rather than queued. This bites right away: #1312 and #1313 both touch `sdk/packages/sdk/src/protocols/intents/**`.

Smaller:

- `E2E_SCENARIOS` isn't trimmed, so `"same-chain, cross-chain"` throws `Unknown scenario  cross-chain`.
- `readEnv` appends `/graphql` before `SECRETS` is computed, so only the suffixed URL is redacted. Anything printing the bare base URL leaks it — worth adding the raw value to `SECRETS` too.
- `submodules: recursive` also clones `evm/lib/sp1-contracts` and `solidity-stringutils`, neither of which this job needs.
- No pnpm store cache on `setup-node`, so every run does a cold install.
- `multi-leg` and `multi-leg-levels` are on a timer. #1311 is already on main, so the moment the testnet gateway is upgraded those two go red and will read like a regression. Better to drop or gate them now than to explain the red later.

Wizdave97:
Thanks. All seven points are addressed in a28590f57:

- **Draft guard:** the job skips draft PRs (`!github.event.pull_request.draft`).
- **Concurrency comment:** it now says what the group does. One run at a time, and a started run is never cancelled. A newer run replaces a pending one, which shows as cancelled.
- **`E2E_SCENARIOS`:** names are trimmed and empty entries dropped, so `"same-chain, cross-chain"` works.
- **Redaction:** every `E2E_*` value is redacted as given as well as as completed. That covers the base orderbook URL alongside the `/graphql` one. Values are replaced longest first, so one never half-redacts another.
- **Submodules:** only `sdk/packages/simplex/proto/mpcvaultapis` is fetched.
- **pnpm cache:** `setup-node` caches the pnpm store (`cache: pnpm`, keyed on `sdk/pnpm-lock.yaml`). pnpm is set up first.
- **Mixed-pair scenarios:** `multi-leg` and `multi-leg-levels` are marked `mixedPairs`. They run only when named, so the default run no longer depends on the gateway upgrade.

One more fix since the review, in 163b79cc1: the partial scenario pays USDC out of the solvers, and a solver could run too low to post its next book. Before each book is posted, a solver is now topped up from the users, so the wallets rebalance in both directions.

